PR #91, opened 2021-07-18
Initial mesh module

### LINKED ISSUE #89: Implementing Meshing module in bluemira
Bluemira is missing a meshing module that can interface the geometry module with future FEM modules.

The mirapy `meshing` module interfaces mirapy `Shape1D` (wires) with the python GMSH API to generate meshes that can be read by meshio and imported into common FEM solvers as Fenics and FreeFEM. It can be adapted for bluemira creating the interface for `BluemiraWire` objects. However we should rethink the use of `lcar` property (as in mirapy `Shape`) since it could create confusion due to the "container" nature of `BluemiraWire` objects.

**Describe alternatives you've considered**
- FreeCAD implements some meshing methods that use GMSH or NETGEN to mesh FreeCAD topological data. However such methods are well integrated into the GUI but difficult to use in a python script (they are strongly interfaced with some GUI options as the selection one). That would require a strong effort.

**Additional context**
It would be worth considering the use of `pygmsh` library instead of the native GMSH python API. Under evaluation.

FYI @DanShort12 @CoronelBuendia

ivanmaione:
This PR sets up an initial mesh module for bluemira. It supports: 

- the definition of an interface between GMSH and BluemiraGeo instances.
- Only a characteristic length is considered as meshing option (mesh field size not implemented here). 
- Physical groups for interface with FEM modules.

The present implementation supports only 1D and 2D mesh.

Closes #89


CoronelBuendia:
@DanShort12 this ready for review once rebased?

DanShort12:
I don't think this is quite ready yet - I don't think we've quite addressed the point about geometries needing to be in the XY plane to be meshed.

bluemira/mesh/meshing.py
```diff
@@ -0,0 +1,721 @@
...
+# import mesher lib (gmsh)
+import gmsh
+
+# import bluemira modules
+import bluemira.geometry as geo
+
+from .error import MeshOptionsError
+
+if TYPE_CHECKING:
+    from bluemira.base.components import Component
```
ivanmaione, line 41: @DanShort12, @je-cook, I am trying to solve a cyclic import problem using type checking (in the same way one of you did in another module with BluemiraGeo). However, when I run the example/mesh/mesh_tutorial.py, I get `NameError: name 'Component' is not defined`. Any idea on how to solve this problem?

je-cook: `TYPE_CHECKING` only protects against circular imports on annotated types (eg `myvar: Union[MyFancyClass, None] = None`) in function/class definitions. You would also need to `from __future__ import annotations` which converts all type annotations to strings therefore they do not need to be imported unless you are running something like `mypy` which at the moment we don't (we are just slowly heading that way)

It doesn't protect against direct type checking which it looks like you are doing

ivanmaione:
@CoronelBuendia, @DanShort12 could you please give a check to this module? It is not "perfect" but it does what is necessary for the development of the fix boundary module. Thanks.

bluemira/mesh/meshing.py
```diff
@@ -0,0 +1,769 @@
...
+                    f"Mesh procedure not implemented for {obj.__class__.__name__} type."
+                )
+            # Each object is recreated into gmsh. Here there is a trick: in order to
+            # allow the correct mesh in case of intersection, the procedure
+            # is made meshing the objects with increasing dimension.
+            for d in range(1, dim + 1, 1):
+                self.__convert_item_to_gmsh(buffer, d)
+            obj.ismeshed = True
+        else:
+            print("Obj already meshed")
```
CoronelBuendia, line 237:
```suggestion
            bluemira_print("Object already meshed")
```
? :)

examples/mesh/mesh_tutorial.py
```diff
@@ -0,0 +1,127 @@
...
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with bluemira; if not, see <https://www.gnu.org/licenses/>.
+
+"""
+Some examples of using bluemira mesh module.
+"""
+
+# %%
```
CoronelBuendia, line 26: this example fails for me in IPython, maybe we can just make it a normal "script" example for now? You'd need to convert it to ipynb technically too, but I think this could wait. Perhaps I'm the only one with this issue (VS Code):

```python
RuntimeError                              Traceback (most recent call last)
Input In [2], in <module>
      1 import os
----> 3 import dolfin
      4 import matplotlib.pyplot as plt
      6 import bluemira.geometry.tools as tools

File ~/miniforge3/envs/bluemira/lib/python3.8/site-packages/dolfin/__init__.py:142, in <module>
    139 from .common.timer import Timer, timed
    140 from .common.plotting import plot
--> 142 from .fem.assembling import (assemble, assemble_system, assemble_multimesh,
    143                              SystemAssembler, assemble_local)
    144 from .fem.form import Form
    145 from .fem.norms import norm, errornorm

File ~/miniforge3/envs/bluemira/lib/python3.8/site-packages/dolfin/fem/assembling.py:34, in <module>
     32 import ufl
     33 import dolfin.cpp as cpp
---> 34 from dolfin.fem.form import Form
     35 from dolfin import MPI
     36 from dolfin.function.multimeshfunction import MultiMeshFunction

File ~/miniforge3/envs/bluemira/lib/python3.8/site-packages/dolfin/fem/form.py:12, in <module>
     10 import ufl
     11 import dolfin.cpp as cpp
---> 12 from dolfin.jit.jit import dolfin_pc, ffc_jit
     15 class Form(cpp.fem.Form):
     16     def __init__(self, form, **kwargs):
     17 
     18         # Check form argument

File ~/miniforge3/envs/bluemira/lib/python3.8/site-packages/dolfin/jit/jit.py:18, in <module>
     16     dolfin_pc = pkgconfig.parse("dolfin")
     17 else:
---> 18     raise RuntimeError("Could not find DOLFIN pkg-config file. Please make sure appropriate paths are set.")
     21 # Copied over from site-packages
     22 def mpi_jit_decorator(local_jit, *args, **kwargs):

RuntimeError: Could not find DOLFIN pkg-config file. Please make sure appropriate paths are set.
```

tests/bluemira/mesh/test_meshing.py
```diff
@@ -0,0 +1,91 @@
...
+try:
+    from bluemira.utilities.tools import get_module
+
+    msh2xdmf = get_module(
+        os.path.join(get_bluemira_root(), "..", "msh2xdmf", "msh2xdmf.py")
+    )
+
+    HAS_MSH2XDMF = True
+except ImportError as err:
+    print(f"Unable to import msh2xdmf, dolfin examples will not run: {err}")
```
CoronelBuendia, line 37:
```suggestion
    bluemira_warn(f"Unable to import msh2xdmf, dolfin examples will not run: {err}")
```

CoronelBuendia (changes requested):
A few minor things from my side, but as this is a prerequisite for the fixed boundary equilibrium coupling I think we should move it in and fix things later on (merge and repent later!)

bluemira/mesh/meshing.py
```diff
@@ -0,0 +1,769 @@
...
+                start_point_tag = gmsh.model.occ.addPoint(
+                    start_point[0], start_point[1], start_point[2]
+                )
+                points_tag.append(start_point_tag)
+                end_point = buffer[type_]["EndPoint"]
+                end_point_tag = gmsh.model.occ.addPoint(
+                    end_point[0], end_point[1], end_point[2]
+                )
+                points_tag.append(end_point_tag)
+                print("Ellipse")
```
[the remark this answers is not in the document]
ivanmaione: removed debug prints

examples/mesh/mesh_tutorial.py
ivanmaione: Removed IPython markdown

CoronelBuendia (approved):

DanShort12 (approved):
Good for an initial implementation and happy with the additions.

Issue #457, opened 2021-11-23
Divertor shape adaptation for SPR-022


DarioV86:
The SPR-022 divertor shape has some issues due to the available space in the divertor region.
In the previous firstwall.py version, that I used to provide a first wall shape to other groups, I fixed the problem.
The new version, does not include this change.
Hopefully, it would be enough to have a SPR-022 first wall profile, but there might be additional problems due to the space.

je-cook:
After our rearchitecture this is part of a reactor design not bluemira core

PR #461, opened 2021-11-24
Spr 022 divertor adaptation

DarioV86:
Issue #457 

The DivertorBuilder, for the Single Null Divertor configuration, to draw the outer leg used to follow the inner leg curvature, giving an inconsistent shape.

Also, some variable names were not that clear. In the divertor region, whatever was labeled with "external" referred to the scrape off layer side, and whatever was labeled with "internal" referred to the private flux region side.


je-cook:
STEP designers should take care of this now

PR #491, opened 2021-12-02
Define glancing angle

DarioV86:
Method to calculate the poloidal projection of the glancing.
Input: glancing angle.

### DIFF
--- a/bluemira/equilibria/flux_surfaces.py
+++ b/bluemira/equilibria/flux_surfaces.py
@@ -150,15 +150,15 @@ class FluxSurface:
         # Poloidal and toroidal magnetic field at the strike point
         Bp_strike = eq.Bp(x_strike, z_strike)
         Bt_strike = eq.Bt(x_strike)
+        # Total magnetic field
+        Btot = np.sqrt(Bp_strike**2 + Bt_strike**2)
 
-        # Numerator and denominator of next operation
-        num = Bp_strike**2 - (Bt_strike**2 * (np.tan(gamma_rad)) ** 2)
-        den = Bp_strike**2 + (Bp_strike**2 * (np.tan(gamma_rad)) ** 2)
-
+        # Numerator of next operation
+        num = Btot * np.sin(gamma_rad)
         # Poloidal projection of the glancing angle
-        cos_theta = np.sqrt(num / den)
+        sin_theta = num / Bp_strike
 
-        return np.rad2deg(np.arccos(cos_theta))
+        return np.rad2deg(np.arcsin(sin_theta))
 
     def plot(self, ax=None, **kwargs):
         """
--- a/bluemira/equilibria/flux_surfaces.py
+++ b/bluemira/equilibria/flux_surfaces.py
@@ -123,6 +123,51 @@ class FluxSurface:
         """
         return np.sum(self._dl(eq))
 
+    def calculate_poloidal_angle(self, eq, x_strike, z_strike, gamma):
+        """
+        From glancing angle to poloidal angle.
+
+        Parameters
+        ----------
+        eq: Equilibrium
+            Equilibrium from which to take magnetic fields
+            at the strike pointcalculate the flux expansion
+        x_strike: float
+            x coordinate of strike point
+        z_strike: float
+            z coordinate of strike point
+        gamma: float
+            Glancing angle at the strike point [deg]
+
+        Returns
+        -------
+        theta: float
+            Poloidal angle at the strike point [deg]
+        """
+        # Technological limit
+        if gamma < 3:
+            raise ValueError(
+                "Glancing angle below the technological limit of 3 degrees."
+            )
+
+        # From deg to rad
+        gamma_rad = np.radians(gamma)
+
+        # Poloidal and toroidal magnetic field at the strike point
+        Bp_strike = eq.Bp(x_strike, z_strike)
+        Bt_strike = eq.Bt(x_strike)
+
+        # Numerator and denominator of next operation
+        num = Bp_strike**2 - (Bt_strike**2 * (np.tan(gamma_rad)) ** 2)
+        den = Bp_strike**2 + (Bp_strike**2 * (np.tan(gamma_rad)) ** 2)
+
+        # Poloidal projection of the glancing angle
+        cos_theta = np.sqrt(num / den)
+
+        # Rad to deg
+        theta = np.rad2deg(np.arccos(cos_theta))
+        return theta
+
     def plot(self, ax=None, **kwargs):
         """
         Plot the FluxSurface.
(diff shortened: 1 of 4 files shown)


tests/equilibria/test_flux_surfaces.py
```diff
@@ -40,6 +41,44 @@
 TEST_PATH = get_bluemira_path("equilibria/test_data", subfolder="tests")
 
 
+class TestFluxSurface:
+    def setup_class(self):
+        eq_name = "DN-DEMO_eqref.json"
+        filename = os.sep.join([TEST_PATH, eq_name])
+        self.eq = Equilibrium.from_eqdsk(filename)
```
je-cook, line 48: If you want this read in once for all methods in this test class it should be the below alos use os.path over os.sep : 
```suggestion
    @classmethod
    def setup_class(cls):
        eq_name = "DN-DEMO_eqref.json"
        filename = os.path.join([TEST_PATH, eq_name])
        cls.eq = Equilibrium.from_eqdsk(filename)
```

bluemira/equilibria/flux_surfaces.py
```diff
@@ -123,6 +123,51 @@ def connection_length(self, eq):
...
+        # Numerator and denominator of next operation
+        num = Bp_strike**2 - (Bt_strike**2 * (np.tan(gamma_rad)) ** 2)
+        den = Bp_strike**2 + (Bp_strike**2 * (np.tan(gamma_rad)) ** 2)
+
+        # Poloidal projection of the glancing angle
+        cos_theta = np.sqrt(num / den)
+
+        # Rad to deg
+        theta = np.rad2deg(np.arccos(cos_theta))
+        return theta
```
je-cook, line 169:
```suggestion
        return np.rad2deg(np.arccos(cos_theta))
```

je-cook (changes requested):
Codewise a couple of minors

bluemira/equilibria/flux_surfaces.py
```diff
@@ -123,6 +123,51 @@ def connection_length(self, eq):
         """
         return np.sum(self._dl(eq))
 
+    def calculate_poloidal_angle(self, eq, x_strike, z_strike, gamma):
```
sebkahn, line 126: `calculate` is not necessary in the function name.     

Let's remove it, it will be also have a better consistency with `connection_length` name.

```suggestion
    def poloidal_angle(self, eq, x_strike, z_strike, gamma):
```

tests/equilibria/test_flux_surfaces.py
```diff
@@ -40,6 +41,44 @@
...
+
+    def test_calculate_poloidal_angle(self):
```
sebkahn, line 50: Follow up on the function name change

```suggestion
    def test_poloidal_angle(self):
```

bluemira/equilibria/flux_surfaces.py
```diff
@@ -123,6 +123,51 @@ def connection_length(self, eq):
...
+        Returns
+        -------
+        theta: float
+            Poloidal angle at the strike point [deg]
+        """
+        # Technological limit
+        if gamma < 3:
+            raise ValueError(
+                "Glancing angle below the technological limit of 3 degrees."
+            )
```
sebkahn, line 151: I am reluctant on raising a true error that halts the code on this angle. The reason is that that technological developpement can actually allow lower angle in the future and we want to remain open to that. The user can also purpusefully use unrealistic angle for scoping studies and see if it is worth finding novel ways on improving on the the angle.

A Bluemira error that does not halts the code but notify the user is more appropritate.

tests/equilibria/test_flux_surfaces.py
```diff
@@ -40,6 +41,44 @@
...
+        # Flux surface in the SoL
+        x_start, z_start = 14, 0
```
sebkahn, line 52: Are `x_start` and `z_start` integers ?

If thet are floats please use floats to define them so we avoid int to float convertion and other mess related to integer (more of a good practice than an issue for this specific code)

```suggestion
        x_start, z_start = 14.0, 0.0
```

tests/equilibria/test_flux_surfaces.py
```diff
@@ -40,6 +41,44 @@
...
+            self.eq.x,
+            self.eq.z,
+            self.eq.psi(),
+            x_start,
+            z_start,
+            self.eq.psi(x_start, z_start),
+        )
+        fs = FluxSurface(Loop(x=x_loop, z=z_loop))
+        # Glancing angle
+        gamma = 5
```
sebkahn, line 63:
```suggestion
        gamma = 5.0

```

tests/equilibria/test_flux_surfaces.py
```diff
@@ -40,6 +41,44 @@
...
+        # Poloidal angle
+        theta = fs.calculate_poloidal_angle(
+            self.eq,
+            10,
+            -7.5,
+            gamma,
+        )
+        assert theta > gamma
```
sebkahn, line 71:
```suggestion
        assert theta > gamma

```

tests/equilibria/test_flux_surfaces.py
```diff
@@ -40,6 +41,44 @@
...
+        # Testing the raise error
+        with pytest.raises(ValueError):
+            fs.calculate_poloidal_angle(
```
sebkahn, line 74:
```suggestion
            fs.poloidal_angle(
```

sebkahn (changes requested):
This is nice contrubution

I have not checked the angle calculation so I am trusting you on it !
We should probably check it with the FIESTA calcualtion though (or the one AH uses at least)

The rest of the comment are mostly code practice and cosmetics

DarioV86:
> This is nice contrubution
> 
> I have not checked the angle calculation so I am trusting you on it ! We should probably check it with the FIESTA calcualtion though (or the one AH uses at least)
> 
> The rest of the comment are mostly code practice and cosmetics

The actual calculation is probably the safest thing you can rely on. For all the coding stuff, never trust me

tests/equilibria/test_flux_surfaces.py
DarioV86: @sebkahn I don't understand this comment

sebkahn (approved):
Ok for me !

sebkahn:
After verification on the derivation, the grazing angle calculation seems valid.

An alternative but equivalent calculation has been found:

`theta = arcsin( sin(gamma) * |Btot| / Bp )`

with `theta` the poloidal angle we are looking for, `|Btot|` the norm of the B field on the target, `gamma` the 3D grazing angle we use as input and `Bp` the norm of the projection of the B field on the poloidal plane.

It can be usefull to get `|Btot|` as:   

`|Btot| = sqrt( Bp**2 + Bt**2 )` 

If this is not directly calculated by the equilibrium object

bluemira/equilibria/flux_surfaces.py
```diff
@@ -123,6 +123,43 @@ def connection_length(self, eq):
         """
         return np.sum(self._dl(eq))
 
+    def poloidal_angle(self, eq, x_strike, z_strike, gamma):
+        """
+        From glancing angle to poloidal angle.
+
```
sebkahn, line 129: 1. Simpler formula with sin to be used
2. Please write the formula in LaTeX here

sebkahn (changes requested):

sebkahn (approved):

bluemira/equilibria/flux_surfaces.py
```diff
@@ -123,6 +123,43 @@ def connection_length(self, eq):
         """
         return np.sum(self._dl(eq))
 
+    def poloidal_angle(self, eq, x_strike, z_strike, gamma):
```
CoronelBuendia, line 126: It strikes me that this calculation only makes sense for an `OpenFluxSurface`, therefore I think it should not live in the `FluxSurface` base class

bluemira/equilibria/flux_surfaces.py
```diff
@@ -123,6 +123,43 @@ def connection_length(self, eq):
...
+
+        Parameters
+        ----------
+        eq: Equilibrium
+            Equilibrium from which to take magnetic fields
+            at the strike pointcalculate the flux expansion
+        x_strike: float
+            x coordinate of strike point
+        z_strike: float
+            z coordinate of strike point
```
CoronelBuendia, line 138: This point should be on the `OpenFluxSurface` but there is no check to that effect. I could ask for the poloidal angle of an `OpenFluxSurface` at a point nowhere near the flux surface and get a result from this.

In fact, this can be seen by the fact that this method is effectively a totally static method (no self used anywhere).

The question is then: is this simply a function that is independent from a flux surface? I'm leaning towards yes. In which case, this might better live in `equilibria/find.py` or indeed here, but as a stand-alone function.

CoronelBuendia (changes requested):

bluemira/equilibria/flux_surfaces.py
sebkahn: > This point should be on the `OpenFluxSurface` but there is no check to that effect. I could ask for the poloidal angle of an `OpenFluxSurface` at a point nowhere near the flux surface and get a result from this.
> 
> In fact, this can be seen by the fact that this method is effectively a totally static method (no self used anywhere).
> 
> The question is then: is this simply a function that is independent from a flux surface? I'm leaning towards yes. In which case, this might better live in `equilibria/find.py` or indeed here, but as a stand-alone function.

I vote for stadalone version and keeping the function here.

CoronelBuendia: Yep, sounds good

sebkahn:
@CoronelBuendia, is there anything we need to do to get this Pull request ?

Ref: @DarioV86, @je-cook

CoronelBuendia (approved):

PR #507, opened 2021-12-06
PF coil builder (with `read` and `mock`)

CoronelBuendia:
## Description
Adds a simple builder for coilsets to enable EU-DEMO stuff to go ahead in future.

Runmodes `mock` and `read` implemented, allowing the user to simply load or specify a `CoilSet` and generate some geometry

Also cleans up plasma colouring

## Interface Changes

None

## Checklist

Requires #521


examples/design/EU-DEMO/EUDEMO_reactor.py
```diff
@@ -237,6 +237,9 @@
             "nx": 2,
             "ny": 2,
         },
+        "PF Coils": {
+            "runmode": "read",
+        },
```
DanShort12, line 242: Can we add plots for the PF coils to the example too? Or at least add them to the xz and xy plots.

DanShort12 (changes requested):
Just a couple of bits. The resulting PF positioning doesn't seem quite right too, as they end up intersecting the outer leg of the TF coils (presumably something isn't being updated after the TF coil optimisation has run?).

image

examples/design/EU-DEMO/EUDEMO_reactor.py
CoronelBuendia: > Just a couple of bits. The resulting PF positioning doesn't seem quite right too, as they end up intersecting the outer leg of the TF coils (presumably something isn't being updated after the TF coil optimisation has run?).
> 
> image

So here we're just reading the PF coils from the reference eqdsk, which was made without any knowledge of what the TF coils should look like. The correct PF coil positions should be delivered with the "run" mode, which I will work on in a separate PR. It's unfortunate, but this state will just leave us with a non-sensical design.

DanShort12 (approved):

DanShort12:
All good from my end after the last changes - @je-cook over to you

je-cook:
Theres a few fixes needed to the notebook but they havent come from here. Additions look really good, starting to look like a reactor properly now! Merging

Issue #309, opened 2021-10-05
Put together an initial EU-DEMO build procedure


CoronelBuendia:
## Description of issue / requirement to address

Using the new `Component`, `Parameter` and `geometry` infrastructure we have put into place over the past few months, we need to establish an as-complete-as-possible build procedure for an indicative EU-DEMO device. Much like was done in BLUEPRINT and MIRA.

This involves writing implementations of many design activities (e.g. running PROCESS, shaping TF coils, design optimisation of PF system and equilibria, etc.).

![image](https://user-images.githubusercontent.com/26097289/150839007-8cc0fe53-e1b4-4600-91da-5820c32961d6.png)

## Proposed solution

Implement the proposed calculation logic in a once-through approach, following the black lines in the image above.

## Alternative solutions

This is intrinsic to the nature of the code.

## Additional context
This will form the starting point of the EU-DEMO design pipeline.

The code to generate the above image is here: https://gist.github.com/CoronelBuendia/2a196584334a87b57e042c4625327045

you must first `pip install pyxdsm` before running `python full_reactor_xdsm.py`

CoronelBuendia:
@DanShort12 @ivanmaione @jmorris-uk @je-cook @hsaunders1904 , as discussed earlier today

je-cook:
As https://github.com/Fusion-Power-Plant-Framework/bluemira/issues/1073#issuecomment-1298308864 says there is a branch currently being worked on for this: [hsaunders1904/eudemo_reactor](https://github.com/Fusion-Power-Plant-Framework/bluemira/tree/hsaunders1904/eudemo_reactor)

je-cook:
This has now been completed, at least in the initial state

PR #521, opened 2021-12-10
Add TF coils to EU-DEMO build

CoronelBuendia:
## Linked Issues
Partially address #309 

## Description

Adds a fairly complete TF coil build to the EUDEMO design
Still some outstanding issues related to bounding box and some sweep inaccuracies, but good enough for now.

## Interface Changes

Quite a few.. 

## Checklist


bluemira/builders/tf_coils.py
```diff
@@ -0,0 +1,337 @@
...
+        Toroidal field ripple constraint function
+        """
+        constraint[:] = self.calculate_ripple(x)
+
+        if grad.size > 0:
+            # Only called if a gradient-based optimiser is used
+            grad[:] = self.optimiser.approx_derivative(
+                self.calculate_ripple, x, constraint
+            )
+        bluemira_debug(f"Max ripple: {max(constraint+self.params.TF_ripple_limit)}")
```
je-cook, line 244: I think we should have a debug flush logger otherwise the terminal gets quite messy (it will all be printed in the log file anyway)

examples/design/EU-DEMO/EUDEMO_reactor.py
```diff
@@ -28,17 +28,22 @@
 
 from bluemira.base.config import Configuration
```
je-cook, line 29: Can we make this file notebook convertible please

bluemira/builders/EUDEMO/reactor.py
```diff
@@ -77,14 +92,58 @@ def build_plasma(self, **kwargs):
...
 
-        builder = PlasmaBuilder(self._params.to_dict(), plasma_config)
+        config = self._process_design_stage_config(name, default_config)
+
+        builder = PlasmaBuilder(self._params.to_dict(), config)
         self.register_builder(builder, name)
 
         return super()._build_stage(name)
+
+    def build_TF_coils(self, component_tree: Component, **kwargs):
```
je-cook, line 104: kwargs are unused here. Is it needed for future dev?

Same with the above two functions. I think they should only be added if they are used right now (we can always add them if needed)

bluemira/builders/EUDEMO/plasma.py
```diff
@@ -116,7 +117,7 @@ class PlasmaBuilder(Builder):
     _plot_flag: bool
     _segment_angle: float
     _eqdsk_path: Optional[str] = None
-    _default_run_mode: str = "run"
+    _default_runmode: str = "run"
```
je-cook, line 120: None of the kwargs in this file are needed at the moment I dont think, tell me if I'm wrong

je-cook (changes requested):
I think I understand the logic of everything here. The example is reasonably straight forward. I went and had a look for unneeded kwargs. But mostly this looks good.

I think after this and #427 go in we should do the factor out of RunMode as you suggested @DanShort12

bluemira/builders/tf_coils.py
DanShort12: Will do, not sure how many flush logs we'll end up with (I guess it makes less sense for warnings and above), so I'll probs just put a new version of the existing function in that supports debug styling rather than generalising any further.

examples/design/EU-DEMO/EUDEMO_reactor.py
DanShort12: Will be in next push (with some actual docs about what the design is doing 😮) - an interesting observation though is that since the bluemira custom printing writes to stderr, the notebook gets a red background for all of our printing, which perhaps isn't so pretty.

DanShort12:
Nearly there, just want to check the other designs still work. Will do that on Monday.

examples/design/EU-DEMO/EUDEMO_reactor.py
je-cook: Oo yeah, so it does. I think the only way to fix that is to go to proper class based logging. 
At the moment we have sort of functionalised the logging but maybe a class based approach would give us some more flexibility to for instance pipe info and debug to stdout. 
Also it might help with the case when after a flush is done there is no newline so the 'box' starts on the wring line for the next message

je-cook:
Looks good so far, really like the example now!

examples/design/EU-DEMO/EUDEMO_reactor.py
DanShort12: Yea, I was in a quandary about how to handle writing the new line after a set of flushing has finished - having a class that has a 'flushing' state that can be switched off on the next print would solve that, and I suspect would help with a lot of other logging flexibility.

DanShort12:
The other design examples still run fine, so I think this is good to go. I'll raise a new issue to convert the plasma and shapes examples into notebooks (and add some better descriptive text) too.

je-cook (approved):

PR #522, opened 2021-12-10
Add volume name metadata to STP output

helen-brooks:
## Description

Propagate CAD model name data to OCC interface so that it is output in the STP file. This is so it may be utilitised in CAD workflows that require this data (targeting especially neutronics).

Strictly speaking this PR is not *necessary*, #523 provides the necessary information for neutronics automation. However if Cubit resolves this  [bug](https://forum.coreform.com/t/missing-volume-metadata-after-stp-import-in-solid-with-void/1324) the additional JSON might be redundant. Also it could be a useful addition for other CAD provider / workflows.

Test manually compares lines between a reference STP file; this was a somewhat hacky solution to exclude lines which change (timestamp, and colour information which even specified directly seems to change its ordering arbitrarily).


helen-brooks:
Ready for review.

sebkahn (approved):

je-cook:
As we are about to update to occt 7.5.1 (because of the freecad dependency in #556) pythonocc needs to be upgraded. Locally testing this the test fails. It looks like there is a bug in pythonocc:
```python
shp_label.AddAttribute(TDataStd_Name(), True)
*** TypeError: in method 'TDF_Label_AddAttribute', argument 2 of type 'opencascade::handle< TDF_Attribute > const &'
```
This is odd because `TDataStd_Name` inherits from `TDF_Attribute` see the [docs](https://dev.opencascade.org/doc/occt-7.5.0/refman/html/class_t_data_std___name.html). Also on pythonocc 7.4.1:

```python
ipdb> isinstance(occ_name, TDF_Attribute)
True
```
Whereas on pythonocc 7.5.1
```python
ipdb> isinstance(occ_name, TDF_Attribute)
False
```

I have rebased and pushed in the hope that its me (also added a little to pass the line ending whitespace)

helen-brooks:
This PR is not essential to facilitating an automated neutronics workflow. Python OCC API changes mean it is not is not worth the headache to fix. Please proceed to review #523 @sebkahn  @je-cook

PR #523, opened 2021-12-10
Write JSON of systems to volume names and IDs.

helen-brooks:
## Description

The purpose of this PR is to provide sufficient information to automate the creation of a DAGMC file from the STP assembly file. To be specific, DAGMC requires the addition of material groups in metadata, so it is necessary to make explicit which volumes in the CAD file correspond to a single material. 

Had it not been for a bug in Cubit, it may have been possible to extract such information from the CAD name data (c.f. #522 ). The specific bug, which is detailed [here]( https://forum.coreform.com/t/missing-volume-metadata-after-stp-import-in-solid-with-void/1324), has the consequence that name data is not imported when the solid is hollow. This is the case for e.g. the PF coil casing. Since we cannot depend on this data being present, I decided to write these grouping in a JSON file. Although less elegant, this approach potentially allows for more future flexibility if it transpires we should not be grouping by system, but by another metric (e.g. a material tag?).

## Key assumptions
- The Volume ID corresponds to the ID in the STP file; this is internally generated by OCC and corresponds to the ordering in which the volumes are added to STP file.
- The list of volume names and IDs are grouped against the system they come from. This is based on the assumption that such systems may correspond to homogenised materials. If this is not the case then the mappings will need to be adjusted accordingly.

### DIFF
--- a/BLUEPRINT/cad/coilCAD.py
+++ b/BLUEPRINT/cad/coilCAD.py
@@ -106,7 +106,7 @@ class PFSystemCAD(ComponentCAD):
 
     def __init__(self, pf_system, **kwargs):
         self.pf_system = pf_system
-        self.name = "Poloidal_field_system"
+        self.name = "Poloidal field system"
         self.component = {
             "shapes": [],
             "names": [],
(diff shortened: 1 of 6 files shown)


je-cook (approved):
Looks good to me over to you @sebkahn

PR #520, opened 2021-12-09
WIP: ConnectionLengthOptimiser for divertor optimisation

alexanderianblair:
## Linked Issues

Optimiser to partially address #204, by introducing an optimiser for connection length calculated starting from the outboard side of the SOL on the midplane to an exterior bounding first wall loop.

## Description

Provides a new `ConnectionLengthOptimiser`, used in the new `divertor_opt.py` example. Currently connection lengths calculated at each step of the optimisation are found to fluctuate significantly for similar coil currents in the example; WIP label will be removed when stabilised.
Uses constraint formulation from #530 to constrain residuals from provided isoflux targets during the connection length optimisation.

Issue #204, opened 2021-09-15
Free boundary equilibrium reformulation for divertor optimization


DanShort12:
In GitLab by @sebkahn on Jun 21, 2021, 12:02

## Description

The free boundary optimisation problem does not include divertor considerations at present.This should be made possible somehow.

## Proposal

The proposal is to make a configurable `OptimisationProblem` in which:
 - The objective function for an equilibrium [current &or position] optimisation can be set by the user
     - Minimisation of error to magnetic constraints
     - Minimisation of the absolute sum of coil currents
     - Minimisation of the total volume of coils 
     - Combined and weighted versions of the above and others, including e.g. connection length &or flux expansion
 - The constraint function(s) for an equilibrium [current &or position] optimisation can be set by the user
     - Error to magnetic constraints
     - Force, field, current constraints

## Additional context

As discussed with @alexanderianblair and @CoronelBuendia on the 3/11/2021

CoronelBuendia:
@sebkahn @markj997 @alexanderianblair I believe that with #1027 all of the functionality required above will be possible, but not implemented "ready-made" as this will likely be quite a specific problem. Could we close this issue when that branch gets merged?

PR #530, opened 2021-12-15
Enable application of user-specified optimisation constraints for coilset optimisation

alexanderianblair:
## Linked Issues

Partially addresses #204, by allowing user specification of constraints to be applied during the coilset optimisation.

## Description

Adds `ConstraintLibrary` and `ObjectiveLibrary` to store NLOpt compatible objective constraint and objective functions for use in Coilset optimisers.

Optimisation constraints are provided by the user as an iterable of `OptimiserConstraint` objects, that store:
- the constraint function
- additional arguments to be passed to the constraint function
- constraint tolerance
- constraint type (inequality or equality)

Split off from #520

### DIFF
--- a/bluemira/builders/EUDEMO/equilibria.py
+++ b/bluemira/builders/EUDEMO/equilibria.py
@@ -25,7 +25,7 @@ Equilibrium objects for EU-DEMO design
 
 import numpy as np
 
-from bluemira.equilibria.constraints import (
+from bluemira.equilibria.eq_constraints import (
     DivertorLegCalculator,
     FieldNullConstraint,
     MagneticConstraintSet,


bluemira/equilibria/opt_problems.py
```diff
@@ -0,0 +1,813 @@
...
+        gamma=1e-8,
+        optimiser=Optimiser(
+            algorithm_name="SBPLX",
+            opt_conditions={
+                "stop_val": 1.0,
+                "max_eval": 100,
+            },
+            opt_parameters={},
+        ),
+        opt_constraints=[],
```
je-cook, line 484: I'm a bit wary of setting mutable objects as defaults to function arguments, both here and opt_parameters just above. Can we do a `None` and internally convert to `[]` etc jump about

https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument

je-cook (changes requested):
Looks generally good. A few minors that I notices.

Future comment for new PR. Some of the detail here probably needs to go into our docs as a standalone file too.

bluemira/equilibria/opt_problems.py
alexanderianblair: Thanks - that could have created some real headaches - particularly for constraints!

Should be noted that the NLOptOptimiser API also suffers from this... unlikely to be too much of an issue, unless it's init'd multiple times, though.

examples/equilibria/double_null_ST.py
```diff
@@ -32,16 +32,23 @@
 from bluemira.equilibria.constraints import IsofluxConstraint, MagneticConstraintSet
 from bluemira.equilibria.equilibrium import Equilibrium
 from bluemira.equilibria.grid import Grid
-from bluemira.equilibria.optimiser import (
```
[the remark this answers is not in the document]
alexanderianblair: Added

examples/equilibria/coilset_opt_problem_tutorial.py
```diff
@@ -0,0 +1,359 @@
...
+# %%
+opt_constraints = [
+    OptimisationConstraint(
+        f_constraint=opt_constraints.current_midplane_constraint,
+        f_constraint_args={"eq": eq, "radius": 1.0},
+        tolerance=np.array([1e-4]),
+        constraint_type="inequality",
+    )
+]
+# %%[markdown]
```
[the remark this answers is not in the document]
alexanderianblair: Should now be fixed

bluemira/equilibria/opt_problems.py
je-cook: Probably should split that out into an issue

je-cook (approved):
I'm happy. Looks good, a great usability improvement!

examples/equilibria/double_null_ST.py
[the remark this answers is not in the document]
je-cook: thanks :)

bluemira/equilibria/opt_objectives.py
```diff
@@ -0,0 +1,156 @@
...
+        ||(Ax - b)||²/ len(b)] + ||Γx||²
+    residual: np.array(n)
+        Residual vector (Ax - b)
+    """
+    residual = np.dot(a_mat, x) - b_vec
+    number_of_targets = np.float(len(residual))
+    fom = residual.T @ residual / number_of_targets + gamma * gamma * x.T @ x
+
+    if not fom > 0:
+        raise bluemira_warn("Least-squares objective function less than zero or nan.")
```
CoronelBuendia, line 155: raise EquilibriaError

bluemira/equilibria/opt_problems.py
```diff
@@ -0,0 +1,811 @@
...
+        initial_currents: np.array
+            Array containing initial (scaled) coil currents prior to passing
+            to sub_opt
+        sub_opt: CoilsetOP
+            Coilset OptimisationProblem used to optimise the array of coil
+            currents at each trial position.
+
+        Returns
+        -------
+        fom: Value of objective function (figure of merit).
```
CoronelBuendia, line 796: float\n

bluemira/utilities/opt_problems.py
```diff
@@ -0,0 +1,249 @@
...
+            Optimiser on which to apply the constraints. Updated in place.
+        opt_constraints: iterable
+            Iterable of OptimisationConstraint objects containing optimisation
+            constraints to be applied to the Optimiser.
+
+        Notes
+        -----
+        Lambda functions are used here to ensure the CoilsetPositionCOP is passed
+        to the function, to allow any properties that are stored in the
+        CoilsetPositionCOP to be accessed at runtime.
```
CoronelBuendia, line 212: Make less specific to equilibrium opt problems

examples/equilibria/coilset_opt_problem_tutorial.py
```diff
@@ -0,0 +1,361 @@
...
+
+# # Introduction
+
+# In this example, we will outline how to specify a CoilsetOptimisationProblem
+# that specifies how an 'optimised' coilset state is found during the Free Boundary
+# Equilibrium solve step.
+
+# # Imports
+
+# Import necessary Equilbrium module definitions.
```
CoronelBuendia, line 37: equilibria

bluemira/equilibria/opt_problems.py
```diff
@@ -0,0 +1,811 @@
...
+        state = self.opt.optimise(initial_mapped_state)
+
+        # Call objective function final time on optimised state
+        # to set coilset.
+        # Necessary as optimised state may not always be the final
+        # one evaluated by optimiser.
+        self.get_state_figure_of_merit(state)
+        return self.coilset
+
+    @staticmethod
```
CoronelBuendia, line 576: I take it there is a good reason for using a static method and dodging a super call? What is it?

examples/equilibria/coilset_opt_problem_tutorial.ipynb
```diff
@@ -0,0 +1,472 @@
+{
```
CoronelBuendia, line 1: As discussed, this doesnt work as an ipynb. As agreed; leave this is as a script for now and drop the ipynb. Might need to add it to an ignore list.

CoronelBuendia (changes requested):
Apologies in advance for the rush review...

This is looking good, and hopefully this OptimisationProblem abstraction will serve more widely.

There are a couple of moderately large pieces of outstanding work that we will have to tackle in future PRs. I'm thinking in particular:
* Dragging OptimisationProblem into GeometryOptimisationProblem #792 
* Adding all of the other constraints into opt_constraints.py #793 
* Tearing down the old optimisers and making sure the old problems still work with this new structure #793 

That said, I'm happy for us to move forward like this (subject to minor changes). Certainly looks like a more solid foundation upon which to expand.

examples/equilibria/coilset_opt_problem_tutorial.py
alexanderianblair: Updated

bluemira/utilities/opt_problems.py
alexanderianblair: Updated

bluemira/equilibria/opt_problems.py
alexanderianblair: Updated

bluemira/utilities/opt_problems.py
```diff
@@ -0,0 +1,249 @@
...
+        # of state vector
+        self.opt.build_optimiser(n_variables=dimension)
+
+        # Apply objective function to optimiser
+        self._objective.apply_objective(self)
+
+        # Apply constraints
+        self.apply_constraints(self.opt, self._constraints)
+
+        # Set state vector bounds (current limits)
```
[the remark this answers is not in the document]
alexanderianblair: Updated

bluemira/utilities/opt_problems.py
```diff
@@ -0,0 +1,249 @@
...
+            self._args["opt_problem"] = opt_problem
+
+        # Apply objective to optimiser
+        opt_problem.opt.set_objective_function(self)
+
+
+class OptimisationProblem(ABC):
+    """
+    Abstract base class to be subclassed for defining optimisation
+    routines in Bluemira.
```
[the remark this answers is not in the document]
alexanderianblair: Updated

bluemira/equilibria/opt_problems.py
alexanderianblair: Primarily to make re-use of objectives easier, from other OptimisationProblems - if the methods are not static, NLOpt will complain if the function is used elsewhere as the first argument will no longer be the state vector.

This way, moving objective functions to the opt_objectives library, or calling them in another optimisation problem, doesn't require any changes to the function.

bluemira/equilibria/opt_objectives.py
alexanderianblair: Updated

bluemira/equilibria/constraints.py
```diff
@@ -366,6 +366,16 @@ def __len__(self):
         """
         return sum([len(c) for c in self.constraints])
```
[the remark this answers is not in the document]
alexanderianblair: Renamed for now. I think changing the name to 'targets' is good for a future PR, as it can be confusing to explain to new users that the MagneticConstraintSet in constraints.py is actually not used as constraints for the optimisation in the usual sense, but the set of targets for the objective function....

I'd be happier if 'constraints' were strictly used only in the OptimisationProblem sense, and that the field targets are renamed, so we can be clear in our documentation about what we're referring to as 'constraints' without the name clash muddying the waters. Nothing to stop the field targets being used as constraints of course, as you say, but I think they remain conceptually different. Should be a separate PR.

Issue #482, opened 2021-12-01
Wire discretisation is sometimes CW, sometimes CCW


CoronelBuendia:
## Description of issue / requirement to address

When we discretise wires to numpy arrays, we do not do any treatment on those wires. Apparently, discretisation does not result in consistent orientations, which unfortunately leads to some really difficult to diagnose problems.

For example:

```python
from copy import deepcopy
from bluemira.geometry.parameterisations import PrincetonD
from bluemira.geometry.tools import offset_wire
from bluemira.geometry._deprecated_tools import check_ccw

x_tf_wp_center = 3.2
parameterisation = PrincetonD(
    {
        "x1": {"value": x_tf_wp_center, "fixed": True},
        "x2": {"lower_bound": 10, "value": 14, "upper_bound": 18},
        "dz": {"lower_bound": -0.5, "value": 0, "upper_bound": 0.5, "fixed": True},
    }
)

wire = parameterisation.create_shape()
array = wire.discretize(byedges=True)

print(check_ccw(array[:, 0], array[:, 2]))  # True

dx_values = [-0.25, 0, 0.25]
dy_values = [-0.4, 0, 0.4]

new_wires = []

for dx in dx_values:
    new_wire = offset_wire(wire, dx)

    for dy in dy_values:
        temp = deepcopy(new_wire)
        temp.translate((0, dy, 0))
        new_wires.append(temp)

for w in new_wires:
    array = w.discretize(byedges=True)
    print(w._orientation)  # They're all _Orientation.FORWARD
    print(check_ccw(array[:, 0], array[:, 2]))  # First 6 True, last 3 False
```
The last three are in the positive quadrant, and somehow go clock-wise, which could either be a result of the offsetting or a result of the translation, or both, don't know.

If you put those arrays into e.g. a Biot-Savart solver, which for better or for worse assumes there is some consistency in terms of CW/CCW, then you get no errors (atm) but you certainly do get weird results. The direction of the sets of the coordinates feeds into the direction of the currents.

## Proposed solution

At the very, very least, we need to have some consistency (even choice) in terms of the CCW-nature of numpy arrays that describe ordered sets of coordinates. 

## Alternative solutions
In my mind, we need a class that behaves a hella lot like a numpy array, but specifically for 3-D sets of coordinates, in which you can choose the CW/CCW orientation of the points, and access them easily via `.x, .y, .z, .xyz`, etc.

This is issue is just another reason to consider this, along with #438

[...]

PR #543, opened 2021-12-16
Implement Coordinates interface

### LINKED ISSUE #438: Standardise the shape of point arrays

In several places in geometry, and elsewhere in some solvers, we are working with numpy `ndarray`s representing clouds of sequentially connected points.

At present, there is an unfortunate mixture of `xyz.shape = (3, N)` and `xyz.shape = (N, 3)`. Accessing a point or a coordinate vector is therefore different depending on how the array was made and is problematic in use and development. 

The fact that these sequentially connected points do not have a clockwise or counter-clockwise convention also causes issues in some places.

```python
  separatrix = JohnerLCFS(
      {
          "r_0": {"value": 9},
          "z_0": {"value": 0},
          "a": {"value": 9 / 3.1},
          "kappa_u": {"value": 1.65},
          "kappa_l": {"value": 1.8},
      }
  )

print(separatrix.create_array().shape)               # --> (3, 203)
print(separatrix.create_shape().discretize().shape)  # --> (100, 3)
```

## Proposed solution

At the very least, we need to standardise the shape of the coordinate arrays across the code-base. I personally prefer (3, N), as it unpacks with `x, y, z = *xyz` enables functions to be called with (*xyz).

Making a polygon and other primitives should also be standardised, so we'd have to choose between:

```python
make_polygon(points=[[0, 0, 0], [1, 0, 0], [2, 0, 0]])
```
and 
```python
make_polygon(x=[0, 1, 2], y=[0, 0, 0], z=[0, 0, 0])
```

## Alternative solutions

This isn't going to be popular, but the `Loop` class in BLUEPRINT was originally made to address these issues, and provide easy access to point and coordinate vectors by indexes and `.x`, `.y`, `.z`, `.xyz` attributes. I'm not saying that we use the `_deprecated_loop.Loop` class, but this is a very real problem, and there is already a solution in the code-base. We could have a very reduced subset of `Loop` functionality for coordinate arrays that treats dimensioning and orientation self-consistenly. Of course, many of the other features of `Loop` are now superseded by other functionality, so we wouldn't need much. For me, and this might sound tragic, but there is also a "readability" gain from `xyz_array[:, 1]` to `points.y` and `xyz_array[0, :]` to `points[0]`

CoronelBuendia:
## Linked Issues

Closes #438, #482 

## Description

Threads in `Coordinates` as the result of a wire discretisation. Effectively give you a (3, N) numpy array, with a couple of extra features.

## Interface Changes

`make_polygon` now takes (3, N) and (N,3) arrays as well as lists of 3-D points, or a dict of `{"x": [...], "y": [...], "z": [...]}` coordinates.

The only change (which was significant due to our many tests with 3-point polygons) is that a (3, 3) array will now be assumed to be a (3, N) array, i.e. the reverse of what it was previously.

The result of `BluemiraWire.discretize` now returns `Coordinates` which is a (3, N) array and not a (N, 3) array as previously.

`Coordinates` can now be accessed via `.x`, `.y`, `.z`, `.xy`, `.xz`, `.yz`, and we have `.points` or `.T` or for (N,3) lists and arrays, respectively.

`Coordinates` are not set as CCW by default, but one can set them to CCW w.r.t. an axis (defaults to the best-fit planar-axis).

`GeometryParameterisation.create_array` removed

## Checklist


DanShort12:
Hi @CoronelBuendia - I can see you've marked this as WIP but not made it a draft PR. I've put this into draft state, so just click "Ready for review" and add someone (or even more than one person now if you want). We get both draft PRs and the ability to add multiple reviewers / assignees now that we are public.

bluemira/geometry/coordinates.py
```diff
@@ -0,0 +1,943 @@
...
+
+    if area != 0:
+        # Zero division protection
+        cx /= 6 * area
+        cz /= 6 * area
+
+    return [cx, cz]
+
+
+def get_centroid_3d(x, y, z):
```
DanShort12, line 417: SonarCloud identified this as a duplicate of [`get_centroid_3d` ](https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/9e94a3f4e58d73bab1521a0b5c6468b0333de139/BLUEPRINT/geometry/geomtools.py#L516). Can we remove that BLUEPRINT version and replace it with this one?

bluemira/geometry/coordinates.py
```diff
@@ -0,0 +1,943 @@
...
+                # Occasionally the two c values are not the same, and one is 0
+                # Take non-trivial value (this works in the case of 2 zeros)
+                return array[np.argmax(np.abs(array))]
+            else:
+                return array[0]
+
+    return [get_rational(i, c) for i, c in enumerate([cx, cy, cz])]
+
+
+def rotation_matrix_v1v2(v1, v2):
```
DanShort12, line 470: SonarCloud also sees this function duplicated in https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/9e94a3f4e58d73bab1521a0b5c6468b0333de139/BLUEPRINT/geometry/geombase.py#L136 and https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/9e94a3f4e58d73bab1521a0b5c6468b0333de139/BLUEPRINT/utilities/plottools.py#L263. Can we use this function in the implementation of those two static functions?

je-cook (approved):
Looks good to me

bluemira/geometry/parameterisations.py
```diff
@@ -195,17 +195,15 @@ def create_array(self, n_points=200, by_edges=True, d_l=None):
 
         Returns
         -------
-        xyz: np.ndarray
+        xyz: Coordinates
```
DanShort12, line 198: Just wondering if this method might end up being confusing since it's called `create_array` but actually returns a `Coordinates` object, which looks like an array but has some different functionality from an array. More a point for discussion really, but perhaps what should be returned here is the underlying array rather than the coordinates?

DanShort12 (commented):
Sorry, one other comment that I should have probably noted before. Happy for it to go in without a change, but figured it at least merits a brief discussion.

bluemira/geometry/parameterisations.py
CoronelBuendia: Hmm.. It is a little misleading. Thing is, I think we normally want coordinates, rather than arrays, seeing as the totality (as far as I have seen so far) of operations involve indexing rather than e.g. a dot product or cross product or whatever. `Coordinates` were intended to behave completely like `np.ndarray` but I found that it wasn't required, and would introduce a lot more work.

Would you be happy with `create_coordinates` as a method name?

DanShort12: Perhaps we could call the method `discretize` since we're essentially wrapping the discretisation of the shape corresponding to that parameterisation?

CoronelBuendia: I guess it is also debatable that this method should exist at all. I know Ivan mentioned this in the past. The reason I left it in was to allow the possibility for somebody to override it by creating the coordinates directly without making the primitive shape. For instance this could be done for the `PrincetonD` by returning the original points, rather than making the original points into a primitive and then discretising the primitive. That said, we're not really using this at the moment. It does look like some premature optimisation... the idea was lets not make a primitive if we don't need to.

PR #546, opened 2021-12-17
🐛 further protect subprocess

je-cook:
## Description

Realised its possible to craft a dict like object that ignores pop. This will now error with two arguments specified if someone tries this. I think we're protected by kwargs anyway but best to be safe.

```python

In [9]: d
Out[9]: {'a': 1}

In [10]: class CustomDict(dict):
    ...:    def __init__(self,*arg,**kw):
    ...:       super().__init__(*arg, **kw)
    ...:    def pop(self, *args, **kw):
    ...:        pass
    ...: 

In [11]: d = CustomDict({'a':1})

In [12]: d.pop('a')

In [13]: d
Out[13]: {'a': 1}
```

## Checklist

### DIFF
--- a/bluemira/codes/interface.py
+++ b/bluemira/codes/interface.py
@@ -98,7 +98,7 @@ class Task:
         kwargs.pop("shell", None)  # Protect against user input
 
         with subprocess.Popen(  # noqa (S603)
-            command, stdout=stdout, stderr=stderr, **kwargs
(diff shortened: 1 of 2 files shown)


DanShort12 (approved):
Good spot, explicitly setting `shell=False` looks like a good way to resolve this and add a backstop behind popping out the shell parameter.

PR #549, opened 2021-12-20
Circular imports

### LINKED ISSUE #533: Strategy for avoiding circular imports

We have quite a lot of cases now where it is easy to introduce circular imports and have collectively spent quite a bit of time resolving cases where they do get introduced. I'm putting this in as a place to collect some thoughts on how we might improve that. Some points of reference:

- We have some cases where inter-dependent functionality are quite far apart, so these should probably be brought together.
- In some cases we have imports that are only used for type hinting. There are some strategies to resolve this (using `TYPE_CHECKING` to only run the imports if a type checker is running), and we should get in the habit of applying these.
- In my experience most of the circular imports boil down to loading something in a `__init__.py`. I personally like being able to see the module for an import when it's being used, but perhaps removing these imports and importing from the module may help in the short term.

## Proposed solution

To be discussed.

## Alternative solutions

To be discussed.

je-cook:
## Linked Issues

Hopefully Closes #533 

## Description

Attempt to deal with circular imports

flake8 fails because it cant find some annotations as imports

## Checklist

### DIFF
--- a/bluemira/base/components.py
+++ b/bluemira/base/components.py
@@ -25,14 +25,14 @@ Module containing the base Component class.
 
 from __future__ import annotations
 
-import anytree
-from anytree import NodeMixin, RenderTree
 from typing import Any, List, Optional, Union
 
-from bluemira.display.plotter import Plottable
-from bluemira.display.displayer import DisplayableCAD
+import anytree
+from anytree import NodeMixin, RenderTree
 
 from bluemira.base.error import ComponentError
+from bluemira.display.displayer import DisplayableCAD
+from bluemira.display.plotter import Plottable
 
 
 class Component(NodeMixin, Plottable, DisplayableCAD):
--- a/bluemira/base/components.py
+++ b/bluemira/base/components.py
@@ -25,7 +25,7 @@ Module containing the base Component class.
 
 from __future__ import annotations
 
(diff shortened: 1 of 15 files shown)


bluemira/base/components.py
```diff
@@ -25,16 +25,14 @@
...
-from typing import Any, List, Optional, Union
 
-from bluemira.display.plotter import Plottable
+from bluemira.base.error import ComponentError
 from bluemira.display.displayer import DisplayableCAD
-
-import bluemira.geometry as geo
-
-from .error import ComponentError
+from bluemira.display.plotter import Plottable
```
DanShort12, line 37: Can we include an `if TYPE_CHECKING:` block so that IDEs can still pick up what the actual class being used is? We'll also need to import `TYPE_CHECKING` from `typing`.

```suggestion
if TYPE_CHECKING:
    from bluemira.geometry.base import BluemiraGeo

```

The `from __future__ import annotations` will treat the annotations as strings so that things will run, but the neat benefits of IDEs being able to work out what a class can do would be lost if we remove the import altogether.

DanShort12 (approved):
Great, thanks a lot for cleaning this up. When things settle down a bit we should definitely document the recommended way of keeping the imports clean.

PR #552, opened 2021-12-20
Use secure links where possible

### LINKED ISSUE #545: Use https when linking to references
## Describe the bug

In a few contexts we use http links to references, we should use https so that we encourage users to securely access those references (where https connections are available).

## Steps to reproduce

1. Check the codebase for http://

## Expected behaviour

Links should use secure connections.

## Additional context

Flagged by SonarCloud as a low priority security hotspot.

DanShort12:
## Linked Issues

Closes #545 

## Description

Some of our links were using http, this updates to go direct to https (flagged by SonarCloud). Also fixes a couple of broken / dead links.

## Interface Changes

N/A

## Checklist


DanShort12:
SonarCloud complained that some of the URLs are duplicate strings and we should be using constants - I resolved that as won't fix because this is metadata and that seemed like overkill to what is already a rather belt and braces change.

je-cook (approved):

PR #554, opened 2021-12-21
🎨 default mapping

### LINKED ISSUE #532: Default parameter mapping should be empty dict not None

Currently we set undefined mappings to None. This is somewhat less user friendly than setting the default to be an empty dict, so we may want to change that default. This would mean one can do:

```python
param.get_param(param_name).mapping["PLASMOD"] = ParameterMapping(plasmod_name, send=True, recv=False)
```

Rather than:

```python
param.get_param(param_name).mapping = {"PLASMOD": ParameterMapping(plasmod_name, send=True, recv=False)}
```

In particular the former case allows the mapping to be created in the same way whether other mappings exist or not, while the latter could lead to unexpectedly removing mappings for other codes.

## Proposed solution

Change the default value of mapping on parameters to be {} rather than None.

je-cook:
## Linked Issues

Closes #532 

## Description

Default mapping is now a dictionary converted inside parameter

## Interface Changes

No longer need to protect against None for parameter.mapping

## Checklist

### DIFF
--- a/tests/BLUEPRINT/cli/test_indir/SMOKE-TEST_template.json
+++ b/tests/BLUEPRINT/cli/test_indir/SMOKE-TEST_template.json
@@ -5,7 +5,7 @@
         "unit": "N/A",
         "description": null,
         "source": "Input",
-        "mapping": null
+        "mapping": {}
     },
     "Av": {
         "name": "Reactor availability",
@@ -13,7 +13,7 @@
         "unit": "N/A",
         "description": null,
         "source": "Input",
-        "mapping": null
+        "mapping": {}
     },
     "B_0": {
         "name": "Toroidal field at R_0",
@@ -21,7 +21,7 @@
         "unit": "T",
         "description": null,
         "source": "Input",
-        "mapping": null
+        "mapping": {}
     },
     "B_tf_peak": {
         "name": "Peak field inside the TF coil winding pack",
@@ -29,7 +29,7 @@
         "unit": "T",
         "description": null,
         "source": "PROCESS",
-        "mapping": null
+        "mapping": {}
     },
     "CS_material": {
         "name": "Conducting material to use for the CS modules",
@@ -37,7 +37,7 @@
         "unit": "N/A",
         "description": null,
         "source": "Input",
-        "mapping": null
+        "mapping": {}
     },
     "C_Ejima": {
         "name": "Ejima constant",
@@ -45,7 +45,7 @@
         "unit": "N/A",
         "description": null,
         "source": "Ejima, et al., Volt-second analysis and consumption in Doublet III plasmas, Nuclear Fusion 22, 1313 (1982)",
-        "mapping": null
+        "mapping": {}
     },
     "F_cs_sepmax": {
         "name": "Maximum separation force between CS modules",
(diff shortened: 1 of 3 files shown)


DanShort12 (changes requested):
Good times. I think this will help a lot with usability for the mapping values. Only thing to check - I don't see any of the template.json files in this... I suspect they will want to be updated from the cli tests and from running examples/design/EU-DEMO/EUDEMO_reactor.py.

je-cook:
We have an occasional failure unrelated to this I think (I've seen it elsewhere): https://github.com/Fusion-Power-Plant-Framework/bluemira/runs/4598203141?check_suite_focus=true

CoronelBuendia:
Naïvely wondering if this could clean up some of the hacking Dan had put into `examples/design/EUDEMO/EUDEMO_reactor.py`?

je-cook:
I don't think so because dictionaries are created rather than using a ParameterFrame directly. It looks to me like the motivation is just showing the mechanics behind the Configuration and therefore ParameterFrame but I could be wrong.

tests/BLUEPRINT/cli/test_indir/SMOKE-TEST_template.json
```diff
@@ -5,2338 +5,1862 @@
...
         "description": null,
         "source": "Input",
-        "mapping": {
-            "PROCESS": {
-                "name": "aspect",
-                "recv": true,
-                "send": true
-            }
-        }
+        "mapping": {}
```
DanShort12, line 8: I get the mappings here have been blanked by the latest changes in the codes that store some default mappings separately from the parameter definitions. However, I think it would still be good to be able to demonstrate how to apply / override those defaults via a script that generates the input json. As long as we have that somewhere then blanking these is fine, I think - would that mean that these cases would be treated as having no mapping, or would they take the defaults?

DanShort12 (commented):
Looks good - a blank dict default should make things a fair bit easier to use than None. My only thought is on the mappings that get saved as a blank dict and how that interacts with the defaults that are now stored separately in codes.

CoronelBuendia:
@je-cook needs a rebase, @DanShort12 happy with this?

tests/BLUEPRINT/cli/test_indir/SMOKE-TEST_template.json
je-cook: They would take the defaults at the moment. It is empty here because no external code is actually run before the template is written and the default mappings are only added then. 

I'm in two minds about whether it should still but null here or an empty dict. I suppose they are essentially saying the same thing.

DanShort12: Still not quite sure about this. If I have some bluemira inputs that set up a run of another code in a particular way (certain mappings and read/write values) and a default is changed in the code then I may silently end up with a different result. My gut says being explicit here will avoid some confusion down the line, but happy to pick up a bit more of a discussion around this.

DanShort12: I've raised #620 to look at a unified way to get the default mappings for an external code onto the ParameterFrame that's being used for a design.

DanShort12 (approved):

PR #556, opened 2021-12-21
⬆️ Upgrades

je-cook:
## Description

Upgrades FreeCAD among other things

## Checklist

- [X] Tests run locally and pass `pytest tests --reactor --longrun --private`


je-cook:
Sonarcloud doesnt like versioneer much

je-cook:
also weird this worked locally...

je-cook:
I dont get these failures locally, any ideas?

je-cook:
Ok I now get it locally. More investigation needed

je-cook:
Definitely updating scipy optimiser changes something. The optimiser finishes at different points giving diffferent solids leading to the failure. Ive pushed a commit here to improve the error catching so that its easier to find the root cause

conda/environment.yml
```diff
@@ -20,22 +20,18 @@ variables:
...
   - python=3.8
-  - pythonocc-core=7.4.1
   - freecad=0.19.2
-  - freecad.python=0.0.3
-  - gmsh=4.6.0
-  - python-gmsh=4.6.0
-  - meshio=4.4.5
-  - mshr=2019.1.0
+  - python-gmsh=4.8.4
+  - gmsh=4.8.4
```
CoronelBuendia, line 25: Might want to check with @ivanmaione if this upgrade is OK, as we're still developing the meshing and I don't think we have a test suite for it yet.

CoronelBuendia (changes requested):
As we've noted, moving to scipy 1.5.3 will cause subtle differences in equilibrium results, which might pass regression tests but might not satisfy e.g. STEP. (especially considering the ongoing benchmark). Tagging @alexanderianblair for future reference

bluemira/geometry/wire.py
```diff
@@ -103,18 +103,30 @@ def _wires(self) -> List[apiWire]:
...
-                            edges.append(edge)
-                        wire = apiWire(edges)
+                        wire = self._reorinate(wireedge)
                     wires += [wire]
             else:
                 wires += o._wires
         return wires
 
+    @staticmethod
+    def _reorinate(wireedge):
```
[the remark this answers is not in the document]
je-cook: yeah :smile: my bad spelling strikes again!

je-cook:
Same tests still failing....very annoying....This one I definitely cant reproduce, one for the new year.

Still not sure whether this is freecad or scipy. I'm now thinking its a freecad problem as locally old freecad and all scipy versions pass fine

conda/environment.yml
je-cook: The changes do not seem to be that major according to the gmsh changelog. It looks like they are all optional additional arguments

je-cook:
scipy nlopt test can be seen here:  https://github.com/Fusion-Power-Plant-Framework/bluemira/discussions/600

conda/environment.yml
DanShort12: There aren't meshing tests yet, but there is at least a tutorial - I can update my environment and give that a run at least... if there are differences then given that the meshing isn't used yet then I'd suggest it's best to upgrade and fix things up after the upgrade.

je-cook: In that case gmsh was updated over the weekend so I've pushed the most recent version

CoronelBuendia (approved):
Nice

DanShort12:
I've applied these updates locally, but had to run this to get `llvmlite` to upgrade:

```bash
pip install -r requirements.txt --ignore-installed llvmlite
conda env update -f conda/environment.yml --prune
```

Just running through the full test suite locally after that in case it introduces any issues.

DanShort12:
Fine with me after a local run of the tests. Probably worth asking users to try upgrading normally with `conda env update -f conda/environment.yml --prune` and then to do `pip install -r requirements.txt --ignore-installed llvmlite` if needed.

PR #564, opened 2021-12-23
Remove `copy` method on objects in `equilibria`

### LINKED ISSUE #430: Remove copy() aliases for deepcopy()

In many modules, many classes (such as `Coil,` `CoilGroup,` `MagneticConstraintSet,` `Equilibrium,` `FluxSurface` among others) have a custom `copy()` method that is an alias for `copy.deepcopy()`. This is undesirable, as this confuses the python shallow and deep copy methods, and makes it more difficult to track when new copies of objects are being created. 

In the worst case, this confusion could lead to unexpected behaviour, where pointers may be pointing to different objects due to a `deepcopy` being performed when a shallow `copy` was expected.

## Proposed solution

Remove these methods, and just use `copy` and `deepcopy` from the `copy` module directly.

CoronelBuendia:
## Linked Issues

Closes #430 

## Description

Removes copy method and use deepcopy where it was being used previously 

## Interface Changes

Objects in the equilibria module no longer have a copy method.
## Checklist


je-cook (approved):
Looks good to me

DanShort12 (approved):

Issue #565, opened 2021-12-24
Update and improve powerlearn


je-cook:
## Description of issue / requirement to address

Update, improve and make powerlearn more generic and create a few test models. 

## Proposed solution

Have a look into what needs updating, probably with an example problem

## Alternative solutions

N/A

## Additional context

powerlearn.py has been removed from the main tree as of #{ID} it is in the `power_learn` branch
Powerlearn is nice functionality to have but it isnt integrated at the moment. A nice to have

PR #566, opened 2021-12-24
🔥 Remove stuff

je-cook:
## Description

Remove stuff, along with moving powerlearn and its surrounding mechanics out of the main branch. It now lives in the branch `power_learn` as stated in #565 

## Checklist


BLUEPRINT/utilities/powerlearn.py
```diff
@@ -1,499 +0,0 @@
...
-        eq += "{0:.9f}*".format(np.exp(self.model.intercept_))
-        for i in range(len(varss)):
-            eq += "{0}^{1:.2f}".format(varss[i], powers[i])
-            if i != len(varss) - 1:
-                eq += "*"
-        eq += "\nR\u00b2 = {0:.2f}".format(self.r_2)
-        bluemira_print(eq)
-
-
-def surface_fit(x, y, z, order: int = 2, n_grid: int = 30):
```
CoronelBuendia, line 391: I actually used this only the other day to fit some CS fatigue stress / n_cycle data... I will probably need it again in the very near future. Could you move this to somewhere in utilities? I'd say in `tools.py` but this feels specific enough to warrant a new file, like maybe `fit_tools.py` or something. When you rebuild the ML stuff and the power and linear law fitting stuff, they would probably live in the same place. It's a bit of a grey zone between fitting and optimisation, but it is (for me at least) conceptually different and definitely a relatively common activity.

CoronelBuendia (changes requested):
LGTM, just want to save one function from the bonfire :)

BLUEPRINT/utilities/powerlearn.py
je-cook: I'll do you one better. In #571 I've generalised the function and moved it to a new fit_tools.py for now

CoronelBuendia (approved):

PR #567, opened 2021-12-24
✨ Create a citation file

### LINKED ISSUE #107: Add CITATION.cff
It should be clear how we expect people to cite bluemira in their work.

GitHub has implemented a summary of the citation file on the front page of a project. We should add a citation file so that it gets picked up by this new mechanism.

**Describe alternatives you've considered**
We could add a citation string on the README, but the GitHub in-built mechanism is cleaner.

je-cook:
## Linked Issues

Closes #107 

## Description

Created from https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md.
May want to change authors from bluemira team to actual names

## Checklist

### DIFF
--- a/ACKNOWLEDGEMENTS.md
+++ b/ACKNOWLEDGEMENTS.md
@@ -1,26 +1,3 @@
-
-# Contributors
-
-## Copyright
-
-M. Coleman<sup>1, 2 </sup>, J. Cook<sup> 1 </sup>, F. Franza<sup> 3 </sup>, I. Maione<sup> 3 </sup>, S. McIntosh<sup>1 ,4</sup>, J. Morris<sup> 1 </sup>, D. Short<sup> 1 </sup>
-
-## Contributors
-
-A. I. Blair<sup>1</sup>, M. Bluteau<sup>1</sup>, M. Coleman,  J. Cook, S. Desai<sup>1</sup>, M. Foord<sup>1, 5</sup>, F. Franza, M.Johnson<sup>1</sup>, S. Kahn<sup>1</sup>, L. Humphrey<sup>1</sup>, C. MacMackin<sup>1</sup>, I. Maione, S. McIntosh, J. Morris, D. Short, D. Vaccaro<sup>1</sup>
-
-### Affiliations:
-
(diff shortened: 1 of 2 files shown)


je-cook:
to view how this works go to view file and then click the root directory . You will see a cite this repo in the right panel. I think we may need to add a preferred citation but there are currently a few issues

- If we used preferred citation only 1 is alowed
- the repo is no longer cited in the link if we use preferred citation

The other option seems to be creating a citation.bib file that the widget will link to. If we do that we could copy the bib file to the working directory on running

CoronelBuendia:
> to view how this works go to view file and then click the root directory . You will see a cite this repo in the right panel. I think we may need to add a preferred citation but there are currently a few issues
> 
>     * If we used preferred citation only 1 is alowed
> 
>     * the repo is no longer cited in the link if we use preferred citation
> 
> 
> The other option seems to be creating a citation.bib file that the widget will link to. If we do that we could copy the bib file to the working directory on running

Perhaps a good example to follow: https://github.com/boutproject/BOUT-dev/blob/master/CITATION.cff

The decision there seems to be to list all the contributors and point to a single paper. Seeing as we will have publications on BLUEMIRA in future, we can add that paper later. For now, if we have to pick one, I would pick none, and just point the users to the README, where the relevant publications are already listed.

Listing all the actual contributors (i.e. not just members on the repo) in the CITATION.cff seems like the best way forward to me, as it acknowledges the work of those who weren't around when those publications were written. If the users want an actual publication to cite (and they usually do) they are in the README, and in future the one BLUEMIRA paper which we can add if and when it gets published.

DanShort12:
In principle I agree with @CoronelBuendia that listing the contributors is a fair way to do this - I suspect a lot of what is in the existing MIRA / BLUEPRINT papers wouldn't align greatly with the current bluemira functionality. However, it does mean that the citation file would need to be maintained as new contributors are added, and we would need to create a doi on zenodo to make the releases citable (when we have a first release).

je-cook:
oops no that didnt work

DanShort12 (approved):
Looks good to me.

CoronelBuendia (approved):

PR #569, opened 2021-12-26
`find_OX_points` improvements

CoronelBuendia:
## Linked Issues

Requires #574 

Started off wanting to remove O and X points that sit inside coils. Turns out they are kind of useful when no O-points are around. In the end I gave up on masking these as they seem to be pretty useful in recovering from bad iterations. It's not perfect..

## Description

* Treat cases where no O-points are found (default to grid centre)
* Treat cases where no X-points are found (default to grid boundary)
* Removes scipy minimisation and focuses on the CG approach
* I tried NLOpt (BGFS, SLSQP and others) and although it works too often there are nlopt failures (of unspecified nature).

## Interface Changes

These changes may break some regression tests, because they affect the iteration path slightly, and therefore the results.

## Checklist


bluemira/equilibria/find.py
```diff
@@ -334,80 +294,82 @@ def find_OX_points(x, z, psi, limiter=None, x_min=None):  # noqa :N802
...
         if point:
             points.append(point)
 
     points = drop_space_duplicates(points)
 
-    o_points, x_points = triage_OX_points(f, points)
+    o_points, x_points = triage_OX_points(f_psi, points)
 
     if len(o_points) == 0:
         print("")  # stdout flusher
```
je-cook, line 334: All our logging goes to `stderr` at the moment so these are a bit redundant

je-cook: in fact it probably leads in to needing a classified logging system so we can track things like this. Ignore for now

je-cook (approved):
Yep :+1:  just going to have a look at some step stuff

In a next PR as we said earlier might be good to look at a better monotonic decay check

Issue #504, opened 2021-12-06
Environment encoding sometimes set implicitly to ASCII instead of utf8


CoronelBuendia:
## Describe the bug

UnicodeEncodeError when running EUDEMO_reactor.py

## Steps to reproduce

1. Run script EUDEMO_reactor.py

4. See error:

```python
---------------------------------------------------------------------------
UnicodeEncodeError                        Traceback (most recent call last)
~/code/bluemira/examples/design/EU-DEMO/EUDEMO_reactor.py in <module>
     53 
     54 with open(f"{get_bluemira_root()}/examples/design/EU-DEMO/template.json", "w") as fh:
---> 55     json.dump(params, fh, indent=2, ensure_ascii=False)
     56 
     57 config = {

~/miniforge3/envs/bluemira/lib/python3.8/json/__init__.py in dump(obj, fp, skipkeys, ensure_ascii, check_circular, allow_nan, cls, indent, separators, default, sort_keys, **kw)
    178     # a debuggability cost
    179     for chunk in iterable:
--> 180         fp.write(chunk)
    181 
    182 

UnicodeEncodeError: 'ascii' codec can't encode character '\xb0' in position 1: ordinal not in range(128)
```
## Expected behaviour

The script should run or mock or read PROCESS and make an initial reactor.

## Environment (please complete the following information)

- OS: Ubuntu 18.04
- Branch: develop

je-cook:
Narrowed it down to 1 line [here](https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/00e082e94514d2cea0dbb68c92ed2cbdb6632044/bluemira/display/auto_config.py#L57)

Its something to do with PySide with app line used to be a context manager I thought too...

je-cook:
Ok its a pyside bug which is fixed by the most recent pyside version (5.15.2 on pip) which is not available at the moment on conda-forge. I see no way around the bug unless we update pyside.
Because we're on conda it restricts us quite forcefully to versioning so maybe the time to try and go pure pip (or maybe poetry?)

CoronelBuendia:
Hmm.. OK, and we can't just pip pyside, taking it out of conda-forge because of FreeCAD, I guess?

je-cook:
It didnt seem to work when I tried that last month, but I will double check at some point this week

je-cook:
Not seen in a long time, will reopen if it reappears

PR #570, opened 2021-12-26
Grad-Shafranov and boundary recognition on grid centroids

CoronelBuendia:
Requires #574 

## Description

Updates grad-shafranov operator so that it is fair bit faster, thanks to @je-cook . I had changed it originally on the centroids, but I'm pretty sure that is plain wrong.
Updates the plasma current distribution such that grid centroid are checked for being inside the LCFS.

## Interface Changes
N/A, but results will change slightly

## Checklist


CoronelBuendia:
Need to update the reference eqdsk in tests but I can't do this atm because of #504

bluemira/equilibria/grad_shafranov.py
```diff
@@ -79,8 +79,8 @@ def __call__(self, nx, nz):
         """
         d_x = (self.x_max - self.x_min) / (nx - 1)
         d_z = (self.z_max - self.z_min) / (nz - 1)
-        x = linspace(self.x_min, self.x_max, nx)
-        d_x2, d_z2 = d_x ** 2, d_z ** 2
+        x = linspace(self.x_min, self.x_max, nx) + 0.5 * d_x  # centroids
```
CoronelBuendia, line 82: This feels stranger and stranger the more I think about it... I think it might actually just be wrong to do this.

je-cook:
brought in some changes from #641 here is the resultant profiling graph, I only changed v3 (even weirder...)

GS_op_prof_2

bluemira/equilibria/grad_shafranov.py
```diff
@@ -93,33 +92,36 @@ def __call__(self, nx, nz):
...
-                A[ind, ind - nz] = (x[i] / d_x2) / rm  # j, l+1
+        i, j = np.meshgrid(np.arange(1, nx - 1), np.arange(1, nz - 1), indexing="ij")
+        i = i.ravel()
+        j = j.ravel()
+        ind = i * nz + j
+        _dx_2_m_halfx = _dx_2 - half_xdx[i]
+        _dx_2_p_halfx = _dx_2 + half_xdx[i]
+        A[ind, ind] = -2 * (_dx_2 + _dz_2)  # j, l
+        A[ind, ind + nz] = _dx_2_m_halfx  # j, l-1
+        A[ind, ind - nz] = _dx_2_p_halfx  # j, l+1
```
alexanderianblair, line 105: Doesn't seem consistent with the values used here in the solver previously? To be consistent, it would be something like
```suggestion
        x_dx_2_xp = 2 * _dx_2 * x[i] / (2 *x[i] + (1.0/_dx) )
        x_dx_2_xm = 2 * _dx_2 * x[i] / (2 *x[i] - (1.0/_dx) )
        A[ind, ind] = - ((x_dx_2_xm + x_dx_2_xp) + 2 *_dz_2)  # j, l
        A[ind, ind + nz] = x_dx_2_xp  # j, l-1
        A[ind, ind - nz] = x_dx_2_xm  # j, l+1
```

(Possible I've made an algebra mistake, of course)

bluemira/equilibria/grad_shafranov.py
```diff
@@ -93,33 +92,36 @@ def __call__(self, nx, nz):
...
-                ghost_factor = 1 + nz % 2
-
-                A[ind, ind] = -(x[i] / d_x2) * (1 / rp + 1 / rm) - ghost_factor / d_z2
-                A[ind, ind - 1] = ghost_factor / d_z2
-                A[ind, ind + nz] = (x[i] / d_x2) / rp  # j, l-1
-                A[ind, ind - nz] = (x[i] / d_x2) / rm  # j, l+1
+        i, j = np.meshgrid(np.arange(1, nx - 1), np.arange(1, nz - 1), indexing="ij")
+        i = i.ravel()
+        j = j.ravel()
+        ind = i * nz + j
```
alexanderianblair, line 100: Much neater

bluemira/equilibria/grad_shafranov.py
```diff
@@ -93,33 +92,36 @@ def __call__(self, nx, nz):
...
+            # d(psi)/dz = 0 across midplane,
+            # else if symmetry boundary is centred on cells, d(psi)/dz
+            # is equal either side of midplane but reverses sign.
+            ind = i * nz + nz - 1
+            ghost_factor = 1 + nz % 2
+
+            A[ind, ind] = -(2 + ghost_factor) * _dz_2
+            A[ind, ind - 1] = ghost_factor * _dz_2
+            A[ind, ind + nz] = _dx_2_m_halfx  # j, l-1
+            A[ind, ind - nz] = _dx_2_p_halfx  # j, l+1
```
alexanderianblair, line 123: See above - no longer consistent with array used previously.

bluemira/equilibria/find.py
```diff
@@ -874,8 +874,10 @@ def _in_plasma(x, z, mask, sep):
         The masking array where 1 denotes inside the zone, and 0 outside
     """
     n, m = x.shape
+    dx = 0.5 * abs(x[0, 1] - x[0, 0])
+    dz = 0.5 * abs(z[1, 0] - z[0, 0])
     for i in range(n):
         for j in range(m):
-            if in_polygon(x[i, j], z[i, j], sep):
+            if in_polygon(x[i, j] + dx, z[i, j] + dz, sep):
```
alexanderianblair, line 881: What's the reason for this shift?

Currents and psi should be stored at the same points in the grid, I believe (since current is related to the second derivative of psi). Magnetic fields are offset.

alexanderianblair (changes requested):
Not quite sure the new solver solves the same problem as previously - although it could be an algebra slip on my end - needs checking.

bluemira/equilibria/grad_shafranov.py
je-cook: I've inverted a few of the divisions at the beginning so this could well have been me

CoronelBuendia: Yes, the divisions switch did break stuff a little (but amazingly not any tests...)
@alexanderianblair is right though, this is a change in scheme. Honestly I was hoping the maths would pan out the same, but there are subtle differences. we discussed it and decided to stick with this one, because it's at least a more faithful implementation of the Jeon paper.

bluemira/equilibria/find.py
CoronelBuendia: honestly I was just trying to get to the bottom of this "discrepancy" we were seeing with FIESTA and was trying this out.

Now that I'm more confident that the discrepancy is due to how the equilibrium constraints / control are being set up differently, I'm less confident this is necessary (it seems to have little effect anyway).

The idea was that when we check for where the plasma current is, that any cell whose centroid is inside the LCFS is registered as carry current.

alexanderianblair: I think I understand the logic - but the cells don't carry current, the nodes do (in this discretisation), and so to lowest order, it should just check if a given node is inside the LCFS. I *think* this is actually adding an error where there wasn't one before...

CoronelBuendia: yeah fair

bluemira/equilibria/grad_shafranov.py
alexanderianblair: Yup - I'm happy with the new scheme - it just appears to be a different second order discretisation, but still solves the same problem.

bluemira/equilibria/grad_shafranov.py
alexanderianblair: As above - resolved as it's a subtle switch to a different - but still valid - second order discretisation of the problem.

alexanderianblair:
Happy with this now `find.py` has been updated and we've got to the bottom of the discretisation scheme differences!

alexanderianblair (approved):

alexanderianblair (approved):

je-cook (approved):

PR #573, opened 2022-01-05
Re-jig documentation structure and add an introduction section

CoronelBuendia:
## Description

This PR proposes an alternative organisation of our documentation, adds an introduction section, and hopefully makes things  little easier to browse through.

I think that the high level divions between: started / modules / developer docs is probably something that will scale better as we move forward too.

## Interface Changes

N/A

## Checklist


documentation/source/index.rst
```diff
@@ -1,15 +1,42 @@
 Welcome to bluemira's documentation!
 =====================================
 
+The documentation is divided into the following sections:
+
+*  :ref:`getting-started`, where you can learn more about the code and how to install it
+*  :ref:`bluemira-modules`, where you can learn more about the individual modules
+*  :ref:`dev-docs`, where you can the API reference, and more developer-focussed documentation
```
je-cook, line 8:
```suggestion
*  :ref:`dev-docs`, where you can see the API reference, and more developer-focused documentation
```

documentation/source/installation.rst
```diff
@@ -0,0 +1,116 @@
+Installation
+============
+
+Configuring SSH keys
```
je-cook, line 4: As we talked about yesterday. I agree that the ssh stuff doesnt need to live here but maybe we should link to it if it is put somewhere else because I can see questions coming up about it

je-cook (commented):
Just a few things

je-cook (approved):
Looks good to me, Ive merged #575 so that test should now 'pass'

documentation/source/parameterisations.rst
```diff
@@ -1,5 +1,5 @@
-Parametrisations
-================
+Parameterisations
```
DanShort12, line 1: Fair, we are rather inconsistent in how we spell parameterisation, it turns out. I think either is valid but we can pick and stick on this one.

DanShort12 (approved):

documentation/source/parameterisations.rst
DanShort12: Raised #579

DanShort12:
Much nicer layout!

Issue #516, opened 2021-12-09
Add support for start and end tangencies when using `make_bspline`


CoronelBuendia:
## Description of issue / requirement to address

Sometime when making a BSplineCurve, it can be very useful to set the start and end tangencies.

## Proposed solution

Add arguments and associated handling for them in `make_bspline` in `bluemira.geometry.tools.py`

## Alternative solutions

I think I had something for this in BLUEPRINT, but it delved quite deep into PythonOCC/OCC... this is in BLUEPRINT/geometry/cadtools.py

```python
def points_to_bspline(
    pnts,
    deg=3,
    periodic=False,
    tangents=None,
    scale=False,
    continuity=GeomAbs_C2,
    tol=TOLERANCE,
):
    """
    Makes a BSplineCurve from an x, y, z array of points

    Parameters
    ----------
    pnts : list or numpy array
        array of x, y, z points
    deg : integer
        degree of the fitted bspline
    periodic : Bool (default=False)
        If true, OCC.GeomAPI_Interpolate will be used instead of the
        GeomAPI_PointsToBspline. Curve tangent vectors can then be enforced at
        the interpolation pnts
    tangents : array (default=None)
        list of [x, y, z] tangent vectors to be specificied at points:
        if only 2 tangents are specified, these will be enforced at the
        start and end points, otherwise tangents should have the same length
        as pnts and will be enforced at each point.
    scale : bool (default=False)
        Will scale the tangents (gives a smoother Periodic curve if False)
    continuity : OCC.GeomAbs.GeomAbs_XX type (default C2)
        The order of continuity (C^0, C^1, C^2, G^0, ....)

    Returns
    -------
    crv : OCC.Geom.BSplineCurve
    """
    if not periodic and (tangents is None):
        crv = _points_to_bspline(pnts, deg=deg, continuity=continuity)
    else:
        typer = TColgp_HArray1OfPnt
        try:
            pnts = point_array_to_TColgp_PntArrayType(pnts, typer).GetHandle()
        except AttributeError:
            pnts = point_array_to_TColgp_PntArrayType(pnts, typer)
        interp = GeomAPI_Interpolate(pnts, periodic, tol)
        if tangents is not None:
            n_tangents = tangents.shape[0]
            if n_tangents == 2:
                interp.Load(gp_Vec(*tangents[0, :]), gp_Vec(*tangents[1, :]), scale)
            else:
[...]

PR #577, opened 2022-01-10
Add EUDEMO TF coils x-z plotting and improve CAD

CoronelBuendia:
## Linked Issues

Our EUDEMO TF CAD was waiting on `slice_wire` (added in #477). We can now make better CAD (because we can use the bounding box of a wire which is apparently much better than that of a solid), and have a proper (x-z plot).

## Description

Using `slice_wire` we can now make some better CAD of the TF coil casing and have an x-z plot of this from a slice of the 3-D solid. Sadly, I still can't see a way to get the 2-D cross-section of casing using primitives (other than resorting to some custom offsetting algorithm, followed by interpolation of a set of coordinates to a BSpline).

image

## Interface Changes
N/A.

## Checklist

### DIFF
--- a/bluemira/builders/EUDEMO/tf_coils.py
+++ b/bluemira/builders/EUDEMO/tf_coils.py
@@ -30,12 +30,14 @@ import numpy as np
 import bluemira.utilities.plot_tools as bm_plot_tools
 from bluemira.base.components import Component, PhysicalComponent
 from bluemira.base.config import Configuration
+from bluemira.base.error import BuilderError
 from bluemira.builders.EUDEMO.tools import circular_pattern_component
 from bluemira.builders.shapes import OptimisedShapeBuilder
 from bluemira.display.palettes import BLUE_PALETTE
 from bluemira.geometry.face import BluemiraFace
 from bluemira.geometry.optimisation import GeometryOptimisationProblem
 from bluemira.geometry.parameterisations import GeometryParameterisation
+from bluemira.geometry.plane import BluemiraPlane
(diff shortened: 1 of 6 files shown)


DanShort12 (approved):
Much nicer implementation for 3D now, and good to see that the xz casing is now available. I can see we're starting to get a few functions that e.g. populate xz from the 3D design, which is probably ok for now, but we might eventually want to think about isolating those - at the moment if we wanted to 2D design then we'd need to make a 3D design first... perhaps that's a logical constraint on the design, but it could start to make things feel a bit spaghetti-like.

CoronelBuendia:
> Much nicer implementation for 3D now, and good to see that the xz casing is now available. I can see we're starting to get a few functions that e.g. populate xz from the 3D design, which is probably ok for now, but we might eventually want to think about isolating those - at the moment if we wanted to 2D design then we'd need to make a 3D design first... perhaps that's a logical constraint on the design, but it could start to make things feel a bit spaghetti-like.

Agreed, I'm not very happy with having to build 3-D in order to get 2-D. A very real immediate problem will be constraining the inner casing shape w.r.t the keep-out zone with the optimisation of the TF current centreline. So for every iteration we'd need to do a whole bunch of 3-D Boolean operations etc. just to get back to a 2-D x-z.

Problem is that to do a variable thickness offset we can only resort to some operations on coordinates (not primitives) and then fit a spline to those. In order to do this robustly we'd need to be able to fit a spline whilst setting the start and end tangencies (see #516). Otherwise we're going to see some pretty weird splines for sure.

Issue #478, opened 2021-11-30
Add coupling to PLASMOD and fixed boundary equilibrium


CoronelBuendia:
## Description of issue / requirement to address

At present, our free boundary equilibrium are solved with input custom profiles, or profiles constrained for `I_p`, `beta_p` and `l_i`.

There is no information relating to actual plasma profiles: the flux functions `p'` and `FF'` are used as mathematical tools, but do not harbor any relation to plasma transport.

We need a way to pass from the PLASMOD 1.5-D transport and fixed boundary equilibrium solver to a asymmetric 2-D fixed boundary equilibrium, using the `p'` and `FF'` from PLASMOD, which treats these more thoroughly (i.e. with a physics basis).

## Proposed solution
Add a 2-D finite element fixed boundary equilibrium solver.
Add coupling to PLASMOD with an asymmetric separatrix, constraining `kappa_95`, `delta_95`, `R_0`, `A`, such that we can reach an asymmetric fixed boundary equilibrium with PLASMOD profiles, which can later be used in free boundary equilibrium calculations.

Iterations between PLASMOD and the 2-D fixed boundary equilibrium will be required.

## Alternative solutions

Continue to use custom profiles, or constraining profiles with plasma integrals. The proposed approach will be superior from a consistency perspective.

ivanmaione:
@CoronelBuendia, I would propose the following steps to close this issue:

1. create a `transport_solver` class that interfaces with PLASMOD (or, in future, other transport codes) and provides `p'` and `FF'`
2. create a 2D fem-fixed-boundary GS solver that calculates `J_plasma` for a given `transport_solver` object
3. create an equilibrium class that iterates between PLASMOD and GS ensuring convergence on specific parameters

As support of step2, a 2D-fem magnetostatic solver is necessary.

A single issue can be created for each of these steps (as already discussed).

CoronelBuendia:
linked to #784 #785 #786 #787

PR #578, opened 2022-01-10
Plasmod bluemira interface update

je-cook:
## Linked Issues

The plasmod side of #478 

## Description

This brings the bluemira interface inline with plasmod on the plasmod repository. This is mostly @fabrizio-franza's work but I have updated the example a bit for notebooks.

I have also changed about some printing formats so they're a bit prettier.

## Interface Changes

N/A

## Checklist


CoronelBuendia (approved):
Example runs nicely. There might be some tweaks to be made when we do the coupling with the fixed boundary equilibrium, but this looks like a solid starting point. We will need to map more parameters into the template and modes into the overal build_config, but we can do that on an as-needed basis once the coupling works gets threaded in.

The changes to the printing do not seem to affect the appearance of anything (aside from parameters).

DanShort12 (approved):

PR #587, opened 2022-01-11
Fix EU-DEMO example

DanShort12:
## Linked Issues

N/A

## Description

Currently the EU-DEMO ipynb doesn't work because it is missing an import that has been sorted outside of the notebook cell. This is resolved by manually correcting the order of the imports in the cell to align with what isort expects. I've also removed the json files, since they are more of an output of the example (for now) than an input, so will change and cause awkward and likely to miss updates as the example progresses.

## Interface Changes

Remove of currently unused json files.

## Checklist


je-cook (approved):
:+1: getting your remove numbers up :smile:

PR #594, opened 2022-01-12
Fix geometry example

DanShort12:
## Linked Issues

N/A

## Description

The geometry example was failing because a 3x3 array was being transposed too many times. There were also other transposes that were unnecessary now that we have the Coordinates translation, and one spelling mistake (at least inconsistency).

## Interface Changes

N/A

## Checklist


CoronelBuendia (approved):
Thanks for catching those!

PR #598, opened 2022-01-13
Re-implement neater spatial interpolators and position mapper

CoronelBuendia:
## Description
Effectively a re-factor of some equilibria positioner.py tools ahead of the re-implementation of coil position optimisation for EU-DEMO.

## Interface Changes

None for now, we agreed to leave the old ones in place until a fully working version is set up.
Still some work to be done, but the idea would be to set up a `PositionMapper` based on a pf_coil track and some exclusion zones, which was mashed together in equilibria/positioner.py

Changed to working as directly as cheaply possible with primitive geometry, i.e. eradicating Loop.

## Checklist


je-cook:
Nice! I like how you can have a combination of region and line optimisers

bluemira/utilities/positioning.py
```diff
@@ -0,0 +1,369 @@
...
+        """
+        if not self.geometry.is_closed:
+            raise PositionerError("RegionInterpolator can only handle closed wires.")
+
+        xz_coordinates = self._get_xz_coordinates()
+        hull = ConvexHull(xz_coordinates.T)
+        # Yes, the "area" of a 2-D scipy ConvexHull is its perimeter...
+        if not np.allclose(hull.area, geometry.length, atol=EPS):
+            raise PositionerError(
+                f"RegionInterpolator can only handle convex geometries. Perimeter difference between convex hull and geometry: {hull.volume - geometry.area}"
```
je-cook, line 181: Dunno how this line got passed the black check, its quite long!

bluemira/utilities/positioning.py
```diff
@@ -0,0 +1,369 @@
...
+                        List[Union[float,
+                        Tuple[float]]]]
+
+            The set of parametric-space values
+        """
+        self._check_length(x)
+        self._check_length(z)
+        return [
+            self.interpolators[i].to_L(x[i], z[i])
+            for i in range(len(self.interpolators))
```
je-cook, line 368:
```suggestion
            tool.to_L(x[i], z[i])
            for i, tool in enumerate(self.interpolators) 
```

bluemira/utilities/positioning.py
```diff
@@ -0,0 +1,369 @@
...
+        Returns
+        -------
+        x: np.ndarray
+            Array of x coordinates
+        z: np.ndarray
+            Array of z coordinates
+        """
+        self._check_length(l_values)
+        xz = [tool.to_xz(l_value) for tool, l_value in zip(self.interpolators, l_values)]
+        return np.array(xz)[:, 0], np.array(xz)[:, 1]
```
je-cook, line 342:
```suggestion
        return np.array([tool.to_xz(l_value) for tool, l_value in zip(self.interpolators, l_values)]).T
```

je-cook: x, z = func() would then work automatically I think

je-cook (commented):
A couple of things I noticed on an initial look

bluemira/utilities/positioning.py
CoronelBuendia: I know, right!?

CoronelBuendia:
OK, so this is in a reasonable state, hesitating over whether to do the following in this PR, or split them out:
* Re-factor `inscribed_rect` to use primitives +/ coordinates rather than `Loop`
* Incorporate the use of inscribed rect somehow in a region positioner or other tool
* Implement a tool in builders/pf_coils.py to break down the PF track into several interpolator regions based on exclusion zones and the like.

je-cook:
I think it may be better for simplicity to add all of those things in a new PR.

I think a refactor of `inscribed_rect` is needed, and probably both interpolators should have access to it. That would make the flow the same. For a path that would then give you a strip. It wouldn't matter that the strip could be concave because the old `inscribed_rect` already deals with that (its only the optimiser that has that issue).

The tools to breakdown the PF track into interpolator paths feels the same as creating a region for the RegionInterpolator. It would be good if the function could be easily generic (even if only the track is implemented for now).

From that I see CS coils as a special case of PathInterpolator with one coordinate. You could even just call PathInterpolator with straight tracks. The tracks will already be split by the time they get to the PositionMapper anyway.

je-cook (approved):

PR #601, opened 2022-01-14
1-D radiation model

DarioV86:
#602 

This 1-D radiation model follows the existing model in PROCESS for core radiation.

The PROCESS model is in the function `plot_radprofile` in `plot_proc.py` .

As in PROCESS, the model for now gives in output a power density [W m^-3] as function of r/a.
Once validated the model, I would add the missing dV, in order to output a plain power [W].

The final goal is to have a 2-D radiation, as function of normalised psi (?). 

Talked so far with Alexander Pearce.
To talk with Francis Casson, and Stuart Henderson.

Note about the radiation in PROCESS: For STEP it currently assumes 80% of Psep is radiated in the SOL (arbitrary number). 
PROCESS has a more complex 1D SOL Model which can produce a consistent 1D radiation profiles. But so far no STEP runs were done including it, and it is not benchmarked for ST's.

Steps:
- adding dV
- implementing a SOL radiation model
- 2-D radiation

Depends on #592


bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,232 @@
...
+        Returns
+        -------
+        species_frac*z_val**2: np.array
+            species ion charge
+        """
+        species_data = Path(Path(imp_data.__file__).parent, species_file)
+        species_frac = self.process_solver.get_process_parameters(fimp_no)
+
+        t_ref = np.genfromtxt(species_data).T[0]
+        z_ref = np.genfromtxt(species_data).T[2]
```
je-cook, line 138: Probably dont want to read in the same file twice (don't forget the last comma otherwise you wont get what you want
```suggestion
        t_ref, z_ref = np.genfromtxt(species_data).T[(0,2),]
```

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,232 @@
...
+            electron temperature
+
+        Returns
+        -------
+        l_val: np.array [W m^3]
+            local values of the radiative power loss function
+        """
+        species_data = Path(Path(imp_data.__file__).parent, species_file)
+        t_ref = np.genfromtxt(species_data).T[0]
+        l_ref = np.genfromtxt(species_data).T[1]
```
je-cook, line 167:
```suggestion
        t_ref, l_ref = np.genfromtxt(species_data).T[(0, 1),]
```

je-cook (commented):
At first glance looks good, just a few minors. Will review properly next week

je-cook:
Ok there are a few things that need to change

1) We cant import directly from process it should be in the codes.process.api and imported from codes.process
2) Probably the plotting shoud be done with the object orientated approach and maybe you shouldnt have to pass the data into the function and it is stored and you pass in a list of elements you want to plot

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,454 @@
...
+        c_b: float
+
+        Returns
+        -------
+        p_den: np.array
+            Power density [W m^-3]
+        """
+        p_den = c_b * z_species * ne ** 2 * np.sqrt(te)
+
+        return p_den
```
je-cook, line 385:
```suggestion
        return c_b * z_species * ne ** 2 * np.sqrt(te)
```

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,454 @@
...
+
+    def collect_flux_tubes(self, psi_n):
+
+        return [self.transport_solver.eq.get_flux_surface(psi) for psi in psi_n]
+
+    def plot_1d_profile(self, rho_conf, rho_sol, te_conf, te_sol):
+
+        rho = np.append(rho_conf, rho_sol)
+        te = np.append(te_conf, te_sol)
+        plt.plot(rho, te)
```
je-cook, line 426:
```suggestion
    def plot_1d_profile(self, rho_conf, rho_sol, te_conf, te_sol, ax=None):
        if ax is None:
            fig, ax = plt.subplots()
        rho = np.append(rho_conf, rho_sol)
        te = np.append(te_conf, te_sol)
        ax.plot(rho, te)
```

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,454 @@
...
+
+    def plot_2d_map(self, flux_tubes, te_flux_tubes):
```
je-cook, line 428:
```suggestion
    def plot_2d_map(self, flux_tubes, te_flux_tubes, ax=None):
        if ax is None:
            fig, ax = plt.subplots()
```

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,454 @@
...
+
+        if len(flux_tubes) == 2:
+            flux_tubes = flux_tubes[0] + flux_tubes[1]
+            te_flux_tubes = te_flux_tubes[0] + te_flux_tubes[1]
+
+        te_min = min([np.amin(t) for t in te_flux_tubes])
+        te_max = max([np.amax(t) for t in te_flux_tubes])
+
+        ax = plt.gca()
```
je-cook, line 437:
```suggestion

```

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,454 @@
...
+                c=t,
+                s=10,
+                cmap="plasma",
+                vmin=te_min,
+                vmax=te_max,
+                zorder=40,
+            )
+
+        f = plt.gcf()
+        f.colorbar(cm, label="KeV")
```
je-cook, line 454:
```suggestion
        fig.colorbar(cm, label="KeV")
```

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,454 @@
...
+            power entering the SoL. P_sol = P_tot - P_rad_main. Unit [W]
+        lambda_q: float
+            near SoL decay lenght. Unit [m]
+
+        Returns
+        -------
+        t_div: np.array
+        """
+        # t_sep from process is in KeV. I need eV.
+        t_sep_mp = self.params["tesep"] * 1000
```
sebkahn, line 237: Make sure it is being tested as @je-cook is working on the units

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,454 @@
...
+        )
+
+        a_par_sol = 4 * np.pi * self.r_sep_mp * lambda_q_near * (Bp_sep_mp / B_sep_mp)
+        p_drop = p_sol / a_par_sol
+
+        t_x_point = ((t_sep_mp ** (7 / 2)) - 7 / 2 * (p_drop / k_0) * (L - L_sep)) ** (
+            2 / 7
+        )
+
+        te_div = self.calculate_sol_decay(t_x_point) / 1000
```
sebkahn, line 263:
```suggestion
        te_div = self.calculate_sol_decay(t_x_point) * 1.0e-3
```

sebkahn (changes requested):

BLUEPRINT/reactor.py
```diff
@@ -280,12 +280,13 @@ def run_systems_code(self):
         """
         Runs, reads, or mocks the systems code according to the build config dictionary.
         """
-        PROCESS_output: ParameterFrame = run_systems_code(
+        self._process_solver = run_systems_code(
```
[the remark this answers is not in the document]
je-cook: This was me, so you can access all the information in the solver instead of just the subset that get carried over to params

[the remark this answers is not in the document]
je-cook: see https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/636

bluemira/radiation_transport/radiation_profile.py
je-cook: At the moment my plan is to have user facing units and between codes units consistent. In the first run through at least I will not be dealing with consistency across the codebase.

If params is a ParameterFrame its units should end up being consistent. I can't guarantee anything else

je-cook (changes requested):
Some changes that I found last week, some easier than others, we can discuss as needed

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,1805 @@
...
+        if lfs:
+            p_fraction = self.params.lfs_p_fraction
+            d = self.sep_corrector
+        else:
+            p_fraction = 1 - self.params.lfs_p_fraction
+            d = -self.sep_corrector
+
+        q_u = p_fraction * q_u * self.params.div_p_sharing
+
+        # Distance between the chosen point and the the target
```
je-cook, line 507: Collapse into a if else for defining forward

je-cook:
```suggestion
if (lfs and z_p < self.points["o_point"]["z"]) or (not lfs and z_p > self.points["o_point"]["z"]):
    forward = True
else: 


forward = forward
```

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,1842 @@
...
+            * constants.E_CHARGE
+            * (self.params.gamma**2)
+            * (constants.E_CHARGE**2)
+            * (n_u**2)
+            * (t_u**2)
+        )
+        # Upstream forcing function
+        f_j = num_f / den_f
+        # To address all the conversion from J to eV
+        f_ev = constants.raw_uc(f_j, "J", "eV")
```
je-cook, line 453:
```suggestion
        # To address all the conversion from J to eV
        f_ev = constants.raw_uc(num_f / den_f, "J", "eV")
```

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -1456,13 +1480,19 @@ def build_sector_profiles(
         return t_pol, n_pol
 
 
-class StepCore(Core, Mathematics):
+class StepCore(Core, TwoPointModelTools):
```
je-cook, line 1483:
```suggestion
class STCore(Core, TwoPointModelTools):
```

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -1554,15 +1584,22 @@ def build_core_radiation_map(self):
         self.plot_2d_map(flux_tubes, total_rad)
 
 
-class StepScrapeOffLayer(ScrapeOffLayerSector, ScrapeOffLayer, Mathematics):
+class StepScrapeOffLayer(ScrapeOffLayerSector, ScrapeOffLayer, TwoPointModelTools):
```
je-cook, line 1587:
```suggestion
class STScrapeOffLayer(ScrapeOffLayerSector, ScrapeOffLayer, TwoPointModelTools):
```

je-cook (changes requested):

je-cook (commented):
Codewise looks ok from me just one minor comment

tests/radiation_transport/test_radiation_profile.py
```diff
@@ -0,0 +1,154 @@
...
+        ])
+        # fmt: on
+
+        impurity_content = {
+            "H": [8.6794e-01],
+            "He": [5.1674e-02],
+        }
+
+        t_ref_name = "T_ref.npy"
+        l_ref_name = "L_ref_He.npy"
```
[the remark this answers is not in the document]
DarioV86: 1x100.
I will leave the two files.

bluemira/base/config.py
```diff
@@ -135,6 +147,17 @@ class Configuration(ConfigurationSchema, ParameterFrame):
         ["f_lfs_upper_target", "Fraction of SOL power deposited on the LFS upper target (DN only)", 0.5, "dimensionless", None, "Input"],
         ["f_hfs_upper_target", "Fraction of SOL power deposited on the HFS upper target (DN only)", 0.5, "dimensionless", None, "Input"],
 
+        # Variables for radiation
+        ["p_sol", "power entering the SoL", 300e6, "W", None, "Input"],
```
CoronelBuendia, line 151: There is for sure some duplication of parameters here... and something like `gamma` is certainly to generic a name for something so specific

bluemira/base/config.py
```diff
@@ -135,6 +147,17 @@ class Configuration(ConfigurationSchema, ParameterFrame):
...
+        ["lambda_q_n", "near SoL decay length", 0.01, "m", None, "Input"],
+        ["lambda_q_f", "far SoL decay length", 0.1, "m", None, "Input"],
+        ["k_0", "material's conductivity", 2000, "dimensionless", None, "Input"],
+        ["gamma", "sheat heat transmission coefficient", 7, "dimensionless", None, "Input"],
```
CoronelBuendia, line 155: sheath I presume

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,1752 @@
...
+        self.separatrix = self.transport_solver.eq.get_separatrix()
+        # The two halves
+        self.sep_lfs = self.separatrix[0]
+        self.sep_hfs = self.separatrix[1]
+        # The mid-plane radii
+        self.x_sep_omp = self.transport_solver.x_sep_omp
+        self.x_sep_imp = self.transport_solver.x_sep_imp
+        # To move away from the mathematical separatrix which would
+        # give infinite connection length
+        self.r_sep_omp = self.x_sep_omp + SEP_CORRECTOR
```
CoronelBuendia, line 115: There is surely a better way to do this than via a CONSTANT...

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,1752 @@
...
+        core=False,
+        n_rad_in=None,
+        n_rad_out=None,
+        rad_i=None,
+        rec_i=None,
+        n_tar=None,
+        x_point_rad=False,
+    ):
+        """
+        Along a single flux tube, it assignes different density values.
```
CoronelBuendia, line 216: assigns

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,1752 @@
...
+            ax.plot(rho, rad_power)
+        else:
+            [ax.plot(rho, rad_part) for rad_part in rad_power]
+            rad_tot = np.sum(np.array(rad_power, dtype=object), axis=0).tolist()
+            ax.plot(rho, rad_tot)
+
+        return ax
+
+
+class TwoPointModelTools(Radiation):
```
CoronelBuendia, line 292: We don't use classes as function namespaces. All of these methods look like functions and shouldn't need to initialise a namespace with Parameters in order to get this behaviour.

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,1752 @@
...
+        flux_tubes: array
+            set of flux tubes
+        x_strike: float
+            x coordinate of the first open flux surface strike point
+        z_strike: float
+            z coordinate of the first open flux surface strike point
+        main_ext: float [m]
+            extention of the radiation region from the x-point
+            towards the main plasma
+        firstwall_geom: grid
```
CoronelBuendia, line 1293: grid or geometry?

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,1752 @@
...
+                n_tar_prof,
+            )
+        ]
+
+        return t_pol, n_pol
+
+
+class STCore(Core, TwoPointModelTools):
+    """
+    Specific class for the core emission of Spherical Tokamak
```
CoronelBuendia, line 1440: What is so different about an ST core w.r.t. a conventional one?

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,1752 @@
...
+
+        # Total line radiation loss along each flux tube
+        total_rad = np.sum(np.array(rad, dtype=object), axis=0).tolist()
+
+        return self.plot_2d_map(flux_tubes, total_rad)
+
+
+class STScrapeOffLayer(ScrapeOffLayerSector, ScrapeOffLayer, TwoPointModelTools):
+    """
+    Specific class for the scrape-off layer emission of Spherical Tokamak
```
CoronelBuendia, line 1531: Ditto here?

examples/radiation_transport/radiation_calculation.py
```diff
@@ -0,0 +1,194 @@
...
+        ["p_sol", "power entering the SoL", 300e6, "W", None, "Input"],
+    ]
+    # fmt: on
+)
+
+# %%[markdown]
+# Run core radiation source calculation for Spherical Tokamak
+
+# %%
+stcore = STCore(solver, impurity_content, impurity_data, plasma_params, rad_params)
```
CoronelBuendia, line 152: The discretisation used in this examples seems to be hard-coded to the grid of the equilibrium and some lin-spacing (hardcoded to 50 points) and looks extremely weird...

examples/radiation_transport/radiation_calculation.py
```diff
@@ -0,0 +1,194 @@
...
+# %%
+stcore.build_mp_rad_profile()
+stcore.build_core_radiation_map()
+
+# %%[markdown]
+# Run scrape-off layer radiation source calculation for Spherical Tokamak
+
+# %%
+stsol = STScrapeOffLayer(
+    solver, impurity_content, impurity_data, plasma_params, rad_params, fw_shape
```
CoronelBuendia, line 166: Why not calculate the values along flux surfaces rather than along this strange grid?

examples/radiation_transport/radiation_calculation.py
```diff
@@ -0,0 +1,194 @@
...
+fw_name = "DN_fw_shape.json"
+fw_name = os.sep.join([read_path, fw_name])
+fw_shape = Loop.from_file(fw_name)
+
+# %%[markdown]
+# Run particle solver
+
+# %%
+p_solver_params = ParameterFrame()
+solver = ChargedParticleSolver(p_solver_params, eq, dx_mp=0.001)
```
CoronelBuendia, line 63: Seems to me like you just need the equilibrium and one other thing that the `ChargedParticleSolver` calculates (i.e. dx_imp and the flux surfaces). Why not just use the equilibrium and the flux surfaces?

bluemira/base/config.py
```diff
@@ -94,6 +94,18 @@ class Configuration(ConfigurationSchema, ParameterFrame):
...
+        ["rho_ped_n", "Density pedestal r/a location", 0.94, "dimensionless", None, "Input"],
+        ["rho_ped_t", "Temperature pedestal r/a location", 0.976 , "dimensionless", None, "Input"],
+        ["n_el_ped", "Electron density pedestal height", 1.086e+20, "1/m^3", None, "Input"],
+        ["T_el_ped", "Electron temperature pedestal height", 3.74, "keV", None, "Input"],
+        ["alpha_n", "Density profile factor", 1.15, "dimensionless", None, "Input"],
+        ["alpha_t", "Temperature profile index", 1.905, "dimensionless", None, "Input"],
+        ["t_beta", "Temperature profile index beta", 2, "dimensionless", None, "Input"],
+        ["n_el_sep", "Electron density at separatrix", 1.5515e+19, "1/m^3", None, "Input"],
+        ["T_el_sep", "Electron temperature at separatrix", 0.48, "keV", None, "Input"],
+        ["rminor", "Minor radius", 2.183, "m", None, "Input"],
```
CoronelBuendia, line 108: This for sure already exists

CoronelBuendia (commented):
OK, so I've had a quick look at this and there's a lot here so apologies if I have misunderstood things. I have not yet looked at the physics. I think there are some problems with how the discretisation can be applied, as the results from the example look strange.

I'm also not clear what the intended outputs are in terms of a "portable" source term of radiation. Sure it plots, but how do I extract the source term?

Isn't the objective to have a radiation source term for a given equilibrium? Is a one-stop-shop interface for such an operation not a desirable thing to have? The issue is not clear on this point. Sure there are three parts to the source term, but I feel like if I want to stitch the results together I'm going to have a hard time.

Some documentation of the model would not go amiss!

I will have a look at the implementation in more detail down the line, but we need to discuss API and functionality first.

examples/radiation_transport/radiation_calculation.py
DarioV86: This discretisation comes from flux_surfaces.py and in turn from equilibrium.py.
I just wanted to be consistent with the discretisation that is used in advective_transport.py.

examples/radiation_transport/radiation_calculation.py
DarioV86: You mean as independent objects?
I call the ChargedParticleSolver to have the same set of flux surfaces

examples/radiation_transport/radiation_calculation.py
DarioV86: So, for sure having the option of chosing the number of points for the discretisation is a nice feature, but I think it would be useful just for a nicer plot, more continuous.
Here, I guess the goal is to calculate a function for the emission source that would serve as input for a tracing code (e.g. CHERAB). 
Indeed, the missing final step here (this would eventually anwer part of the question in the heading), is to generate a grid from which extrapolating the said function. 
Also, talking about grids, the scrape off layer one, is not an actual grid, or at least not made purposely. It is just the result of the points discretising each flux surface. I didn't set any grid, yet.

bluemira/base/config.py
DarioV86: I changed `p_sol` with the already existing` P_rad`, and I renamed `gamma` with `gamma_sheath`. Together with the description it should be quite explicit

bluemira/radiation_transport/radiation_profile.py
DarioV86: Having a connection length calculated correctly.
At the moment the calculation is very unstable near the separatrix. I need to move away from it by a random distance which varies according to the equilibrium.
An alternative would be a sort of iterator that moves further till the connetction length calculation "looks" correct. It doesn't seem worth, though.

bluemira/base/config.py
DarioV86: I removed `T_el_ped` and used the existing `T_e_ped`, and I removed `rminor` and used `A` and `R_0`

bluemira/radiation_transport/radiation_profile.py
DarioV86: Similarly. I collected everything in just one class, and left just what is strictly necessary for a DN in another class.

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -34,13 +34,627 @@
...
+    f_ev = constants.raw_uc(num_f / den_f, "J", "eV")
+    # Critical target temperature
+    t_crit = eps_cool / gamma
+    # Finding roots of the target temperature quadratic equation
+    coeff_2 = 2 * (eps_cool / gamma) - f_ev
+    coeff_3 = (eps_cool**2) / (gamma**2)
+    coeff = [1, coeff_2, coeff_3]
+    roots = np.roots(coeff)
+    if roots.dtype == complex:
+        t_tar = constants.raw_uc(f_ion_t, "keV", "eV")
```
je-cook, line 190: this logic here could be joined with line 193 making this just `t_tar = f_ion_t` and tabbing in l193

je-cook:
superceeded by #2224

PR #603, opened 2022-01-14
Shape design parameter validation and notebook conversion

DanShort12:
## Linked Issues

N/A

## Description

The component tree for the design_shape example was failing as the Design was being defined without a name. This PR looks to ensure that generic Designs are defined with all of the required parameters in place (most are handled by the validation on the resulting Builder classes). It adds the Name parameter to the example design, and converts the example to a Jupyter notebook.

## Interface Changes

Validation added to ensure required parameters are provided to a design.

## Checklist

### DIFF
--- a/examples/design/design_shapes.py
+++ b/examples/design/design_shapes.py
@@ -43,6 +43,7 @@ from bluemira.geometry.optimisation import GeometryOptimisationProblem
 # (since minimising the negative length will maximise the value of the absolute length).
 # Such an optimisation problem can be defined as below.
 
+
 # %%
 class MaximiseLength(GeometryOptimisationProblem):
     """
--- a/examples/design/design_shapes.py
+++ b/examples/design/design_shapes.py
@@ -23,12 +23,27 @@
 A basic tutorial for configuring and running a design with parameterised shapes.
 """
 
+# %%
 from bluemira.base.design import Design
 from bluemira.geometry.optimisation import GeometryOptimisationProblem
 
-# Make an example optimisation problem to be solved in the design.
+# %%[markdown]
+# # Configuring and Running a Simple Shape-Based Design
+#
+# This example shows how to set up and run a shape-based Design that uses a parameterised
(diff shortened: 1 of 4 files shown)


DanShort12:
FYI @hsaunders1904 - this should resolve the issue where viewing component trees was failing in Designs - it will enforce that you give the Design a Name parameter.

examples/design/design_shapes.py
```diff
@@ -59,8 +74,34 @@ def f_objective(self, x, grad):
...
+#     class.
+#   - maps the R_0 and A Design parameters to the r_0 and a shape parameters.
+#   - labels the resulting component as "Shape".
+#
+# - The TF Coils build stage:
+#   - uses the `MakeOptimisedShape` Builder class.
+#   - parameterises the TF Coils centerline shape using the `PrincetonD`
+#     GeometryParameterisation class.
+#   - maps the r_tf_in_centre and r_tf_out_centre Design parameters to the x1 and x2
+#     shape parameters. The dz shape parameter is set to 0 and fixed. The x1 shape
```
je-cook, line 96: Is there currently a way to view the optimisation variables without diving into the code? 
From a user perspective it would probably be nice to be either able to print that out or maybe better have a diagram explaining what the variables are in the docs (but that doesnt really work for all classes).

If not I think we should say something like the optimisation variables for PrinectonD are <vars> which modify <eg the z offset of the coil>

je-cook (commented):
Just one comment really

examples/design/design_shapes.py
DanShort12: I was thinking that for geometry parameterisation classes we could implement a function that allows the default settings of the class to be drawn with some annotations, then use a custom sphinx directive to automatically generate and include the diagram in our docs. That should be possible, so I'll put an issue in. I agree that it would be useful to be able to print the parameters used by a parameterisation, regardless of that auto-plotting, so I'll put another issue in for that.

DanShort12: See #607 #608

je-cook (approved):

PR #606, opened 2022-01-17
Add track tool for PF coil positioning

CoronelBuendia:
## Description
Adds a track decomposition tool for placing PF coils along a wire based on exclusion zones and a set of coils.

Adds support to `distance_to` to handle points
Adds `point_inside_shape` because I thought I needed it for this... Note that the FreeCAD version seems much less accurate on a 2-D shape than previous discretised methods.

## Interface Changes

`distance_to` can now handle iterables of length 3 and treat them as points

## Checklist

### DIFF
--- a/bluemira/builders/EUDEMO/pf_coils.py
+++ b/bluemira/builders/EUDEMO/pf_coils.py
@@ -288,16 +288,11 @@ def make_segments(track, exclusion_zones, coils):
             split_values = l_values[:-1] + 0.5 * np.diff(l_values)
             # Sub-divide into BSplines for now...
             # TODO: Actual primitive sub-division less fun...
-            split_values = np.append(split_values, 1.0)
+            split_values = np.concatenate([[0.0], split_values, [1.0]])
             sub_segs = []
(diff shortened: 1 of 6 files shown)


je-cook (commented):
Looks good. 

Is there a reason you see the `make_coil_mapper` DEMO specific? 
The only bit I see as specific is the creation of the PositionMapper object in the return statement

CoronelBuendia:
> Looks good.
> 
> Is there a reason you see the `make_coil_mapper` DEMO specific? The only bit I see as specific is the creation of the PositionMapper object in the return statement

I just didn't want to assume that it would be generic enough. Arguably, the whole approach of doing the coil position interpolation this way is just "one way of doing it", and the creation of individual segments for individual coils is equally questionable. STEP, I'm sure, will be using combinations of PathInterpolator and RegionInterpolator, and won't use this.

I'm also not 100% sure this is what the final form of this utility will end up looking like, I just wanted to break things down into small chunks before tackling the full PF position optimisation problem.

je-cook (approved):

PR #630, opened 2022-01-21
Create mapping edit helper

je-cook:
## Description

This makes it easier for users to modify the send/recv mappings of a code 

## Interface Changes

Additional `set_mappings` function. `ParameterMapping` now errors if you try to set other attributes to it

## Checklist

### DIFF
--- a/bluemira/base/parameter.py
+++ b/bluemira/base/parameter.py
@@ -94,6 +94,27 @@ class ParameterMapping:
         """
         return repr(self.to_dict())
 
+    def __setattr__(self, attr: str, value: Union[bool, str]):
+        """
+        Protect against additional attributes
+
+        Parameters
+        ----------
+        attr: str
+            Attribute to set (name can only be set on init)
+        value: Union[bool, str]
+            Value of attribute
+
+        """
+        if attr not in ["send", "recv", "name"] or (
+            hasattr(self, "name") and attr not in ["send", "recv"]
+        ):
+            raise KeyError(f"{attr} cannot be set for a {self.__class__.__name__}")
+        elif attr in ["send", "recv"] and not isinstance(value, bool):
+            raise ValueError(f"{attr} must be a bool")
+        else:
+            super().__setattr__(attr, value)
+
 
 class ParameterMappingEncoder(json.JSONEncoder):
     """
--- a/bluemira/codes/interface.py
+++ b/bluemira/codes/interface.py
@@ -333,7 +333,14 @@ class FileProgramInterface:
         Parameters
         ----------
         mappings: dict
-            A list of keys wi
+            A dictionary of variables to change mappings.
+
+        Notes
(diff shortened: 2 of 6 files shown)


bluemira/codes/interface.py
```diff
@@ -326,6 +326,24 @@ def _send_mapping(self) -> Dict[str, str]:
...
         return self.setup_obj._send_mapping
 
+    def modify_mappings(self, mappings: Dict[str, Dict[str, bool]]):
+        """
+        Modify the send/recieve mappings of a key
+
+        Parameters
+        ----------
+        mappings: dict
+            A list of keys wi
```
CoronelBuendia, line 336: wi... th what?

examples/codes/run_plasmod_example.py
```diff
@@ -249,3 +249,16 @@ def plot_profile(plasmod_solver, var_name, var_unit):
...
+# The mapping can be changed on a given parameter or set of parameters.
+# Notice how the value of `q_95` doesn't change in the output
+# even though its value has in the parameter (the previous value of 3.5 is used).
+
+# %%
+plasmod_solver.modify_mappings({"q_95": {"send": False}})
+plasmod_solver.params.q_95 = (5, "input")
+plasmod_solver.run()
+print_outputs(plasmod_solver)
+print("\nq_95 value history\n", plasmod_solver.params.q_95.history())
```
CoronelBuendia, line 264: I think it is fine if:
* when we modify a mapping to no longer send, the last value is kept by default
* we can then set a value on a parameter as shown above to overwrite the last value (if necessary)

bluemira/codes/interface.py
CoronelBuendia, line 336: wi.. th what?

bluemira/base/parameter.py
```diff
@@ -94,6 +94,27 @@ def __str__(self):
         """
         return repr(self.to_dict())
 
+    def __setattr__(self, attr: str, value: Union[bool, str]):
+        """
+        Protect against additional attributes
```
CoronelBuendia, line 99: dataclass with __slots__ comes in 3.10 :'(

CoronelBuendia (changes requested):
Just minor but I think I'm happy with the default logic here.

examples/codes/run_plasmod_example.py
je-cook: Yeah, for plasmod definitely that is what happens

bluemira/base/parameter.py
je-cook: ...yeah very annoying, but at least its coming

CoronelBuendia (approved):

DanShort12 (approved):
Great, gives a nice way to guide the user on how to adjust the mappings.

PR #636, opened 2022-01-24
Codes cleanup and finish generic wrapper

je-cook:
## Description

This does a bit of cleanup of some of codes and does some renaming to enable generic wrappers that came up from discussions and thoughts on Friday. Code interfaces can now be loaded on the fly from anywhere with builtin location in `bluemira.codes` the preference.

## Interface Changes

`plot_PROCESS` has been renamed to `plot_radial_build` and `run_systems_code` to `systems_code_solver`. `systems_code_solver` now returns the solver object which needs to be run externally (allowing flexibility and reruns)

## Checklist

### DIFF
--- a/BLUEPRINT/systems/plotting.py
+++ b/BLUEPRINT/systems/plotting.py
@@ -181,10 +181,9 @@ class ReactorPlotter:
         """
         Plots a 1-D vector of the radial build output from PROCESS
         """
-        # TODO: POSTprocess.py is archaic... need to refactor heavily and bring
-        # into the fold...
-        sys_code_dir = self.reactor.file_manager.generated_data_dirs["systems_code"]
-        codes.plot_radial_build(os.sep.join([sys_code_dir, "OUT.DAT"]), width=width)
+        codes.plot_radial_build(
+            self.reactor.file_manager.generated_data_dirs["systems_code"], width=width
+        )
 
     def plot_xz(self, x=None, z=None, show_eq=False, force=False):
         """
--- a/BLUEPRINT/systems/plotting.py
+++ b/BLUEPRINT/systems/plotting.py
@@ -184,7 +184,7 @@ class ReactorPlotter:
         # TODO: POSTprocess.py is archaic... need to refactor heavily and bring
         # into the fold...
         sys_code_dir = self.reactor.file_manager.generated_data_dirs["systems_code"]
-        codes.plot_PROCESS(os.sep.join([sys_code_dir, "OUT.DAT"]), width=width)
+        codes.plot_radial_build(os.sep.join([sys_code_dir, "OUT.DAT"]), width=width)
 
     def plot_xz(self, x=None, z=None, show_eq=False, force=False):
         """
--- a/BLUEPRINT/systems/plotting.py
+++ b/BLUEPRINT/systems/plotting.py
@@ -22,7 +22,6 @@
 """
 Plotting utilities for ReactorSystem objects
 """
-import os
 from itertools import cycle
 
 import matplotlib.pyplot as plt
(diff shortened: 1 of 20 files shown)


BLUEPRINT/systems/plotting.py
```diff
@@ -184,7 +184,7 @@ def plot_1D(self, width=1.0):
         # TODO: POSTprocess.py is archaic... need to refactor heavily and bring
         # into the fold...
         sys_code_dir = self.reactor.file_manager.generated_data_dirs["systems_code"]
-        codes.plot_PROCESS(os.sep.join([sys_code_dir, "OUT.DAT"]), width=width)
+        codes.plot_radial_build(os.sep.join([sys_code_dir, "OUT.DAT"]), width=width)
```
CoronelBuendia, line 187: This output filename is specific to PROCESS. If this was just in BLUEPRINT we could leave as is. I do wonder how we'd plot the PROCESS output in the EUDEMO reactor though. Not sure how best to handle this, but another systems code will likely how a different radial build plotting method. The one we have for PROCESS right now is really janky because it works on the OUT.DAT directly.

My instinct is always to add a plot method to a solver... Maybe that is bad

bluemira/builders/EUDEMO/reactor.py
```diff
@@ -78,13 +77,14 @@ def run_systems_code(self):
...
-        output: ParameterFrame = run_systems_code(
+        self.syscode_solver = systems_code_solver(
             self._params,
             config,
             self._file_manager.generated_data_dirs["systems_code"],
             self._file_manager.reference_data_dirs["systems_code"],
         )
-        self._params.update_kw_parameters(output.to_dict())
+        self.syscode_solver.run()
+        self._params.update_kw_parameters(self.syscode_solver.params.to_dict())
```
CoronelBuendia, line 87: So here for example, if I want to see the radial build of a reactor, I have to do:

```python
import os
from bluemira.codes.wrapper import plot_radial_build
sys_code_dir = reactor.file_manager.generated_data_dirs["systems_code"]
plot_radial_build(os.sep.join([sys_code_dir, "OUT.DAT"]), width=1.0)
```
I'm not sure that's right, but if we could have a neat interface to this somehow from the `Reactor` object that'd be great.

je-cook: Ok with the most recent push it would now look like 

```python
from bluemira.codes import plot_radial_build
plot_radial_build(reactor.file_manager.generated_data_dirs["systems_code"], width=1.0)
```

which is better. But it doesnt address the original point. Working on how best to formulate a response :slightly_smiling_face:

je-cook:
> This output filename is specific to PROCESS. If this was just in BLUEPRINT we could leave as is. I do wonder how we'd plot the PROCESS output in the EUDEMO reactor though. Not sure how best to handle this, but another systems code will likely how a different radial build plotting method. 
> 
> My instinct is always to add a plot method to a solver... Maybe that is bad

I think this goes back to whether the `systems_code_solver` function (and other similar possible functions) become a class themselves. 

Possibly the solver object should have plotters associated with it but it feels odd to me because it is not in my head at least part of the solver itself. 
It seems to me that it is post processing on the solver output which should be a separate thing

bluemira/builders/EUDEMO/reactor.py
CoronelBuendia: @DanShort12 and I had a chat about this too, off the back of #707. We should catch up and have a chat; Dan has an idea how to generalise this kind of "analysis stage" for a reactor, and access it in the same way we access build stages.

DanShort12: Yea, for builders we do `register_builder` and `get_builder`, so we could do the same for solvers (or generalise the register/get for builders to something like register/get build_stage and include solvers in that). As a more future-looking change we probably also want to (perhaps optionally) store history information for builders/solvers - like we do for parameters. Then we could say 'for my ith PROCESS run, what did my radial build look like'. It will be pretty memory intensive though, hence why it's probably an option to run when developing/investigating something.

CoronelBuendia:
@je-cook conflicts but looks good, I'd like to Dan to give it a quick look but then I think it can go in.

CoronelBuendia:
Ready for rebase now with #749  in.

DanShort12 (approved):
I agree with the comments about post-processing for adding some of the flexibility to support different plotting methods - that can go into a separate issue, and I also agree that it's going to be difficult to know what that looks like until we have other cases, so we can wait until we capture that requirement until it's needed. Otherwise a good enhancement that cleans things up and adds a good amount of flexibility.

Issue #629, opened 2022-01-21
Review of builder functionality


DanShort12:
## Description of issue / requirement to address

We've used the new builders /design framework for a while now, so it would be good to have a review of what is working well and what could do with a clean up or further improvements. I've put this ticket in to facilitate some discussion around that.

@CoronelBuendia - you've probably used the design / builders stuff the most other than me, so would appreciate any other thoughts you might have.

## Proposed solution

TBD

DanShort12:
1. We seem to be keeping to separation of concerns fairly well, and I like how the logic for making the geometry is separated out from the solving of optimisation problems / reading previous result / mocking. It could be useful to break some of the build functions down further so that there are individual functions for each component.

2. The callable logic on builders can be a bit fiddly, and as seen in #589 this can lead to some complications when arguments need to be passed to one method but not another. It may be worth unwinding this and just having the user make separate calls to the relevant methods (at least until there are enough examples to factor the problem back down into a form that works in all cases).

3. Related to 2., we're currently doing quite a bit that hides the real signatures of the builder methods. This leads to too much use of `**kwargs` and results in it being unclear how a method should be called because e.g. the IDE doesn't know what arguments can be passed in, or what help text is available. I think this is one of the main issues at the moment, and probably warrants some of this being unwound.

4. The run/read/mock can be a bit tricky to use when arguments need to be passed into the design problem. This leads tothe different modes needing to be called differently, which then ends up hiding what's needed for the actual method from the user (as for 3. this is seen in a resulting proliferation of `**kwargs` to cope with this).

I think my main gripe from the points I've listed so far is that we're doing quite a bit to hide functionality from the user, which results in the usage being obscured or at least hard to discover e.g. in an IDE.

CoronelBuendia:
I completely agree with your points 2, 3, and 4. The main problem is that it is at present difficult to unpick the logic of what arguments are getting used where. I fear doing everything in a `kwargs` dict is not going to be sustainable, long-term.

The real signatures are obfuscated, and a lot of logic happens quite deep under the hood in various abstractions. The result is (because I still unfamiliar with this arhitecture) that I have to click through all of the subclasses and private methods every time in order to check it will do what I hope.

I think the priority should remain "ease of use" for the average user, who might want to change some problem settings or parameterisation, etc. But we should consider "ease of developer use" in terms of writing new `Builders` and the like. Sadly, I don't have any immediate suggestions on how to improve things!

[...]

PR #650, opened 2022-01-26
Move arguments to init

DanShort12:
## Linked Issues

Related to #629 

## Description

This moves various kwargs usage, which is currently prevalent in many function signatures, to the init function. In this way, each builder can be set up and reinitialised as required, but is always called the same way. The usage of explicit arguments on the implementations of the init functions for concrete classes should also make it easier to follow what arguments need to be passed in.

## Interface Changes

The `Builder` classes now setup everything that they need in `__init__` and `reinitialise`. This means that the `build` and runmodes (`run`/`read`/`mock`) are called consistently without any arguments. The `_build_stage` in `Reactor` is now also called without arguments. `kwargs` are removed from places where they are not longer used.

## Checklist

### DIFF
--- a/bluemira/builders/EUDEMO/plasma.py
+++ b/bluemira/builders/EUDEMO/plasma.py
@@ -152,7 +152,6 @@ class PlasmaBuilder(Builder):
         bluemira_print("Running Plasma equilibrium design problem")
         eq = self._create_equilibrium()
         self._analyse_equilibrium(eq)
-        self._boundary = make_polygon(eq.get_LCFS().xyz, "LCFS")
 
     def read(self):
         """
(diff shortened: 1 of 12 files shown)


bluemira/builders/EUDEMO/plasma.py
```diff
@@ -165,7 +162,6 @@ def read(self):
         eq = self._read_equilibrium()
         self._analyse_equilibrium(eq)
         self._boundary = make_polygon(eq.get_LCFS().xyz, "LCFS")
-        return {"equilibrium": eq}
```
CoronelBuendia, line 168: Shouldn't we set `self._equilibrium` here?

CoronelBuendia (changes requested):
OK, looks neater and more explicit now. Keen to get this in so we can propagate these changes into some pending PRs.

bluemira/builders/EUDEMO/plasma.py
DanShort12: It's set in `_analyse_equilibrium` that way whenever the parameters are updated the equilibrium gets updated too... coming to think of it, the same logic also follows for updating the boundary in `_analyse_equilibrium` to ensure that is kept aligned too... I've added that in too.

CoronelBuendia (approved):

PR #653, opened 2022-01-27
Add `BluemiraPlane` from matrix

### LINKED ISSUE #652: Add a `from_matrix` class method to `BluemiraPlane`

I personally struggle with the (base, axis, angle) formulation of a `BluemiraPlane`, because it doesn't really do what I expect, even if that may be correct as a method. 

We already have `from_3_points` as a classmethod, enabling one to make a plane from 3 points, but I'm now running into the situation where I already know the rotation matrix of a plane and I just want to make a `BluemiraPlane` from a 4 x 4 matrix. 

This is possible in FreeCAD's `Placement`.

## Proposed solution

Add a classmethod to `BluemiraPlane` allowing the user to make a `BluemiraPlane.from_matrix`

## Alternative solutions

Part of the point of outsourcing the heavy lifting to FreeCAD is to not have to maintain these geometry transformations which honestly are pretty fiddly and different conventions for edge cases are possible. BUT, the alternative would be to write (or re-write) some of these transformations.

@ivanmaione

CoronelBuendia:
## Linked Issues

Closes #652 

## Description

Adds a classmethod to `BluemiraPlane` to instantiate from a matrix

## Interface Changes

Adds a classmethod to `BluemiraPlane` to instantiate from a matrix

## Checklist

### DIFF
--- a/bluemira/codes/_freecadapi.py
+++ b/bluemira/codes/_freecadapi.py
@@ -1162,6 +1162,11 @@ def make_plane_from_matrix(matrix):
     if not matrix.shape == (4, 4):
         raise FreeCADError(f"Matrix must be of shape (4, 4), not: {matrix.shape}")
 
+    for i in range(3):
+        row = matrix[i, :3]
+        matrix[i, :3] = row / np.linalg.norm(row)
+    matrix[-1, :] = [0, 0, 0, 1]
+
     matrix = Base.Matrix(*matrix.flat)
     return Base.Placement(matrix)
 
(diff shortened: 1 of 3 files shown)


CoronelBuendia:
> @CoronelBuendia as discussed, please add a test with a matrix that is not normalized (i.e not extracted by a placement). It should also work, but better to check.

Good spot... I'm quite surprised that FreeCAD does not do this normalisation. We also now enforce the last row to be (0, 0, 0, 1) silently - somehow I felt that was easier than warning about bad matrices.

ivanmaione (approved):

PR #646, opened 2022-01-26
Add squareness calculation to `ClosedFluxSurface`

CoronelBuendia:
## Linked Issues

c.f. STEP teams chat

## Description

Adds a calculation for the outer upper and lower squareness of a closed flux surface, following Holcomb et al.'s definition of it, see: DOI:10.1063/1.3125934

## Interface Changes

`eq.analyse_core` now returns `CoreResults` in which the squareness is also included

## Checklist

I confirm that I have completed the following checks:

- [x] Tests run locally and pass `pytest tests --reactor`
- [x] Code quality checks run locally and pass `flake8` and `black .`
- [x] Documentation built locally and checked `sphinx-build -W documentation/source documentation/build`

PR #656, opened 2022-01-27
Change flux surface convention and add documentation

### LINKED ISSUE #654: Check flux surface geometry conventions and write documentation describing the geometry in detail

Upon adding the squareness calculation, I remember that I had put another problem aside; that of the actual flux surface geometry definitions.

Many parameters are technically defined w.r.t. some centre point, the location of which is sometimes debatable. Most textbooks (e.g. Stangeby, Freidberg, Ariola + Pironti) do tend to define things like elongation and triangularity w.r.t to pure geometry, but it is at least in some cases considered reasonable to define these w.r.t. the magnetic axis.

image

I've asked Hartmut Zohm for advice because I have to say, I'm a bit worried about different definitions being used in different places.

## Proposed solution

At the moment `ClosedFluxSurface` has a bit of a mixed set of definitions here, so these must be standardised.

We also need to add some clear documentation describing these with some images.

CoronelBuendia:
## Linked Issues

Requires #646 
Closes #654 

Flux surface geometry now defined w.r.t. the centroid of the fictional reference ellipse, in line with most textbooks and seemingly general convention. Still an outstanding doubt related to the definition of upper and lower triangularity and vertical Shafranov shift where I've made an opinionated decision that they should be w.r.t. to the average vertical position of the radial extrema. 

It's still pretty debatable on how to handle the vertical position of the "centroid" however, especially when one considers things like NT, SF, or the general funky case where the elevations of the minimum and maximum radii of the shape are not nearly equal.

## Description

Clears up the convention on calculation of geometric flux surface quantities and adds documentation.

## Interface Changes

None, but flux surface analysis results will change.

## Checklist


je-cook (approved):
Looks good to me just waiting on #646

PR #658, opened 2022-01-28
Allow 3D builds through provided angle

DanShort12:
## Linked Issues

N/A

## Description

This PR allows 3D builds to be created through a specified angle. This is intended to allow components to be visualised without wrapping around the whole reactor, allowing for cut-through views.

## Interface Changes

The `build_xyz` methods are given an optional `degree` parameter, which is 360 by default.

## Checklist

### DIFF
--- a/bluemira/builders/EUDEMO/tf_coils.py
+++ b/bluemira/builders/EUDEMO/tf_coils.py
@@ -397,17 +397,32 @@ class TFCoilsBuilder(OptimisedShapeBuilder):
 
         return component
 
-    def build_xyz(self) -> Component:
+    def build_xyz(self, degree: float = 360.0) -> Component:
         """
         Build the x-y-z components of the TF coils.
+
+        Parameters
+        ----------
+        degree: float
+            The angle [°] around which to build the components, by default 360.0.
+
+        Returns
+        -------
+        component: Component
+            The component grouping the results in 3D (xyz).
         """
         component = Component("xyz")
+
+        # Minimum angle per TF coil
+        min_tf_deg = 360.0 / self._params.n_TF.value
+        n_tf_draw = min(int(degree // min_tf_deg) + 1, self._params.n_TF.value)
+        degree = min_tf_deg * n_tf_draw
 
         # Winding pack
         wp_solid = sweep_shape(self._wp_cross_section, self._centreline)
         winding_pack = PhysicalComponent("Winding pack", wp_solid)
         winding_pack.display_cad_options.color = BLUE_PALETTE["TF"][1]
-        sectors = circular_pattern_component(winding_pack, self._params.n_TF.value)
+        sectors = circular_pattern_component(winding_pack, n_tf_draw, degree=degree)
         component.add_children(sectors, merge_trees=True)
 
         # Insulation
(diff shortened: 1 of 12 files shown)


bluemira/builders/EUDEMO/tf_coils.py
```diff
@@ -397,17 +397,36 @@ def build_xy(self) -> Component:
...
 
+        # Minimum angle per TF coil (nudged by a tiny length since we start counting a
+        # sector at theta=0). This means we can draw a sector as 360 / n_TF and get one
+        # TF coil per sector. Python represents floats with 16 significant figures before
+        # getting round off, so adding on 1e-13 works here, in case someone sets n_TF
+        # to be 2.
+        min_tf_deg = (360.0 / self._params.n_TF.value) + 1e-13
+        n_tf_draw = min(int(degree // min_tf_deg) + 1, self._params.n_TF.value)
+        degree = min_tf_deg * n_tf_draw
+
```
DanShort12, line 424: I wasn't so sure what the expected behaviour should be here. I went with making sure that one coil is shown per sector, which I think would generally be what people would expect to happen, but in that building through 180 degrees will not give the coil opposite the first (which I guess could be another expected way that this would work).

CoronelBuendia (commented):
OK, so this will work fine, but I'm still hesitating here...

For most intents and purposes, we do want to build the full CAD model (all sectors and all components).

For viewing we typically only want to display a sub-set of the CAD model. So it feels weird to build only the half the CAD model for display purposes, because we will have to build the full CAD model for our technical use cases.

I'm just wondering how difficult it would be to build the full CAD model, and then only display certain parts of it with `[Reactor|Component].show_cad(view_mode="something")`. This could be done by selecting the sectors in the tree, for those components which are patterned, and by some boolean cuts (or alternative degree revolutions) for the few components that are full revolutions (right now, just plasma and PF coils).

Long-term it's also possible that we want to build "alternative" CAD models for a reactor, like a single sector, with sliced TF coils for modelling in axisymmetry. That use case is only tangentially related to this PR, but I'm just wondering if it's worth considering in terms of getting this "right" first time round.

DanShort12:
It would certainly be possible to write some logic into `show_cad` to view sectors and so on, or to have a broader view selector. I was thinking though that a build following the same steps as the full one would be more representative than effectively adding in additional build steps (cutting and so on) in the display. I was also thinking this way is more extensible, since other use cases can satisfy whatever view they want via their builder, rather than having to also write their own displayer.

I think the alternative models would be supported today by adding in alternative build methods (on top of `build_xz` etc.) - of course there may need to be some refactoring along with that when the need arises.

bluemira/builders/EUDEMO/tf_coils.py
CoronelBuendia: It looks "right" to me.

bluemira/builders/EUDEMO/tf_coils.py
```diff
@@ -397,17 +397,36 @@ def build_xy(self) -> Component:
...
         component = Component("xyz")
 
+        # Minimum angle per TF coil (nudged by a tiny length since we start counting a
+        # sector at theta=0). This means we can draw a sector as 360 / n_TF and get one
+        # TF coil per sector. Python represents floats with 16 significant figures before
+        # getting round off, so adding on 1e-13 works here, in case someone sets n_TF
+        # to be 2.
+        min_tf_deg = (360.0 / self._params.n_TF.value) + 1e-13
+        n_tf_draw = min(int(degree // min_tf_deg) + 1, self._params.n_TF.value)
+        degree = min_tf_deg * n_tf_draw
```
CoronelBuendia, line 423: just for my sanity, could we do:

```suggestion
        degree = (360.0 / self._params.n_TF.value) * n_tf_draw
```

CoronelBuendia (changes requested):
OK, also makes sense!

CoronelBuendia (approved):

Issue #675, opened 2022-01-31
Equilibrium analysis tools


CoronelBuendia:
## Description of issue / requirement to address

This is a general issue to gather requirements for a set of functionality relating to equilibrium queries. Once we have a fairly complete list of required functionality, we can then discuss implementation and do the work.

At present a wide range of different tools are available in the `equilibria` module and in the BLUEPRINT `StreamFlow` object. The latter has some functionality which is not present in the new module, but is set to be deprecated.

## Proposed solution

In my head, there are three generic categories of request for information on an `Equilibrium` object:
* Calculation of some floating point value, integral or otherwise: (e.g. psi at some point, internal inductance of the plasma, flux surface elongation)
* Queries as to where some physical feature lies in space: (e.g. finding the O, X-points, finding the LCFS, finding a flux surface based on some criterion)
* Queries related to the structure of the core plasma: (e.g. profiles, values plotted on a per-flux-surface basis)
Implement functionality for the following equilibrium queries:

At present, these bits of functionality are strewn about between `equilibria/physics.py`, `equilibria/find.py`, and `equilbria/flux_surfaces.py`. I propose to compile a list of desired functionality, restructure these tools in a clearer way, and implement any missing features.

## Desired functionality

| Description of functionality| Exists in `equilibria` |Name(s)  | New name | 
| ------------- | ------------- | ------------- | ------------- | 
| Finding the O and X points| &#9746; |`find_OX_points`, `eq.get_OX_points()` | |
| Finding the last closed flux surface | &#9746; |`find_LCFS_separatrix`, `eq.get_LCFS()` | |
| Finding the separatrix | &#9746; |`find_LCFS_separatrix`, `eq.get_separatrix()` | |
| Finding the separatrix legs | &#9746;  | `get_legs`| |
| Finding a flux surface from a specified point | &#9746;  | `find_flux_surface_through_point`| |
| Finding a flux surface(s) from a normalised psi vale| &#9746;  | `find_flux_surf`, `eq.get_flux_surface()`| |
| Finding the inboard midplane point | &#9744;  | | |
| Finding the outboard midplane point | &#9744;  | | |
| Calculating the radial field due to the equilibrium and coils | &#9746;  | `eq.Bx()`| |
| Calculating the vertical field due to the equilibrium and coils | &#9746;  | `eq.Bz()`| |
| Calculating the poloidal field due to the equilibrium and coils | &#9746;  | `eq.Bp()`| |
[...]

PR #677, opened 2022-02-02
Add get legs functionality to `equilibria`

CoronelBuendia:
## Linked Issues
Partially addresses #675 

## Description

Adds a `get_legs` function to `equilibria/find.py` to extract the separatrix legs from an equilibrium. Tested for SN and DN.

## Interface Changes

See above.

## Checklist

### DIFF
--- a/bluemira/equilibria/find.py
+++ b/bluemira/equilibria/find.py
@@ -816,7 +816,7 @@ def get_legs(equilibrium, n_layers: int = 1, dx_off: float = 0.0):
         dx_offsets = np.linspace(0, dx_off, n_layers)[1:]
 
     if isinstance(separatrix, Iterable):
-        # Double null (sort in/out top/bottom)
+        # Double null (sort in/out bottom/top)
         separatrix.sort(key=lambda half_sep: np.min(half_sep.x))
         x_points.sort(key=lambda x_point: x_point.z)
         legs = []
@@ -824,7 +824,7 @@ def get_legs(equilibrium, n_layers: int = 1, dx_off: float = 0.0):
             for x_p in x_points:
                 sep_leg = _extract_leg(half_sep, x_p.x, x_p.z, delta, o_point.z)
                 quadrant_legs = [sep_leg]
-                if dx_offsets:
+                if dx_offsets is not None:
                     quadrant_legs.extend(
                         _extract_offsets(
                             equilibrium, dx_offsets, sep_leg, direction, delta, o_point.z
@@ -844,7 +844,7 @@ def get_legs(equilibrium, n_layers: int = 1, dx_off: float = 0.0):
         legs.sort(key=lambda leg: leg.x[0])
         inner_leg, outer_leg = legs
         inner_legs, outer_legs = [inner_leg], [outer_leg]
-        if dx_offsets:
+        if dx_offsets is not None:
             inner_legs.extend(
                 _extract_offsets(
                     equilibrium, dx_offsets, inner_leg, -1, delta, o_point.z
--- a/bluemira/equilibria/find.py
+++ b/bluemira/equilibria/find.py
@@ -744,12 +744,7 @@ def get_legs(equilibrium, n_layers: int, dx_off: float):
     Returns
     -------
     legs: Dict[str, List[Loop]]
-        Geometries of the legs, sorted as:
-            [lower inner,  # if present
-             lower outer,  # if present
-             upper inner,  # if present
-             upper outer,  # if present
-            ]
+        Dictionary of the legs with each key containing a list of geometries
 
     Raises
     ------
@@ -759,15 +754,16 @@ def get_legs(equilibrium, n_layers: int, dx_off: float):
     -----
     Will return two legs in the case of a single null
     Will return four legs in the case of a double null
+
+    We can't rely on the X-point being contained within the two legs, due to interpolation
+    and local minimum finding tolerances.
     """
 
-    def extract_leg(flux_line, x_point):
-        radial_line = Loop(
-            x=[x_point.x - delta, x_point.x + delta], z=[x_point.z, x_point.z]
-        )
+    def extract_leg(flux_line, x_cut, z_cut):
+        radial_line = Loop(x=[x_cut - delta, x_cut + delta], z=[z_cut, z_cut])
         arg_inters = join_intersect(flux_line, radial_line, get_arg=True)
         arg_inters.sort()
-        if x_point.z < o_p.z:
+        if z_cut < o_point.z:
             # Lower null
             func = operator.lt
         else:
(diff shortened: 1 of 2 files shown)


bluemira/equilibria/find.py
```diff
@@ -725,6 +728,90 @@ def get_flux_loop(psi_norm):
...
+            if leg.argmin(x_point[:2]) > 3:
+                leg.reverse()
+
+            flux_legs.append(leg)
+        if len(flux_legs) == 1:
+            flux_legs = flux_legs[0]
+        return flux_legs
+
+    o_points, x_points = equilibrium.get_OX_points()
+    o_p = o_points[0]
```
hsaunders1904, line 795: Would making `o_p` an argument in `extract_leg` work? It took some digging to find where this definition was being used.

bluemira/equilibria/find.py
```diff
@@ -725,6 +728,90 @@ def get_flux_loop(psi_norm):
...
+            for x_p in x_points[:2]:
+                leg = extract_leg(half_sep, x_p)
+                legs.append(leg)
+    else:
+        # Single null
+        legs = extract_leg(separatrix, x_points[0])
+
+    # Sort [lower inner, lower outer, upper inner, upper outer]
+    legs.sort(key=lambda leg: leg.x[0] + leg.z[0])
+    return legs
```
hsaunders1904, line 812: How would you feel about using a `dict` instead of a list as a return type here? Then, when we're using the return value, it's immediately clear what leg we're accessing, e.g.,

```python
legs = get_legs(eq)
Divetor.make_target(legs["LOWER_INNER"])
```

As we have a known set of keys, we could also consider using an `Enum`. Then the typing tells us exactly what keys are allowed:

```python
class Leg(Enum):
    LOWER_INNER = "lower_inner"
    LOWER_OUTER = "lower_outer"
    ...

def get_legs(equilibrium) -> Dict[Leg, Loop]:
    ...
```

hsaunders1904 (commented):
Am I right in thinking that this outputs a single loop for each leg? The old `StreamFlow` implementation outputted several loops (scrape off layers?). Are those other layers no longer needed?

I've left a couple of comments on the code, let me know what you think

bluemira/equilibria/find.py
```diff
@@ -725,6 +728,90 @@ def get_flux_loop(psi_norm):
...
+            x=[x_point.x - delta, x_point.x + delta], z=[x_point.z, x_point.z]
+        )
+        arg_inters = join_intersect(flux_line, radial_line, get_arg=True)
+        arg_inters.sort()
+        if x_point.z < o_p.z:
+            # Lower null
+            func = operator.lt
+        else:
+            # Upper null
+            func = operator.gt
```
je-cook, line 771:
```suggestion
        # Lower null == Lessthan,  Upper null == Morethan
        func = operator.lt if x_point.z < o_p.z else operator.gt
```
as the if else here is just defining a single short variable

je-cook (commented):
Not much to add really, I agree with Harry's comments

bluemira/equilibria/find.py
je-cook: With the enum it would be possible to keep the list and just set `LOWER_INNER = 0` etc. No preference however

CoronelBuendia: > Am I right in thinking that this outputs a single loop for each leg? The old `StreamFlow` implementation outputted several loops (scrape off layers?). Are those other layers no longer needed?
> 
> I've left a couple of comments on the code, let me know what you think

Yes, this would just get you the legs of the actual separatrix. Do you need more layers? Can add them in, I can certainly see how it might help divertor design, just wasn't sure if you needed them for what is already there.

CoronelBuendia: > With the enum it would be possible to keep the list and just set `LOWER_INNER = 0` etc. No preference however

Happy to put in a dict and enum here, yep

bluemira/equilibria/find.py
CoronelBuendia: it's the same O-point every time, I will rename o_p and x_p though to make it clearer. Sonarcloud thinks this is too complicated anyway, so I will try and refactor a bit.

bluemira/equilibria/find.py
```diff
@@ -725,6 +728,124 @@ def get_flux_loop(psi_norm):
...
+            )
+            offset_legs.append(extract_leg(Loop(x=xl, z=zl), x, z))
+        return offset_legs
+
+    o_points, x_points = equilibrium.get_OX_points()
+    o_point = o_points[0]
+    x_points = x_points[:2]
+    separatrix = equilibrium.get_separatrix()
+    delta = equilibrium.grid.dx
+    n_layers = max(2, n_layers)
```
hsaunders1904, line 811: Should we be honouring the number of layers that were passed into the function? I'd probably be a bit surprised if I'd asked for a single layer but got two - especially as this is not documented in the doc string.

bluemira/equilibria/find.py
```diff
@@ -725,6 +728,124 @@ def get_flux_loop(psi_norm):
...
+            "upper_inner": legs[1],
+            "upper_outer": legs[3],
+        }
+    else:
+        # Single null
+        x_point = x_points[0]
+        legs = extract_leg(separatrix, x_point.x, x_point.z)
+        legs.sort(key=lambda leg: leg.x[0])
+        inner_leg, outer_leg = legs
+        inner_legs, outer_legs = [inner_leg], [outer_leg]
```
hsaunders1904, line 837:
```suggestion
        inner_legs, outer_legs = ([leg] for leg in legs)
```

hsaunders1904 (commented):
Works well for me 👍 . I noticed that, for the case I'm testing, the legs don't extend all the way to the x-point. Is this OK/to be expected?

bluemira/equilibria/find.py
CoronelBuendia: yeah fair does, actually probably one is OK too, but will require a bunch of checks. I will try and get this less smelly too.

CoronelBuendia:
> Works well for me +1 . I noticed that, for the case I'm testing, the legs don't extend all the way to the x-point. Is this OK/to be expected?
> 
> 

Yeah so this is due to the way we get the separatrix at present, with the old approach in `StreamFlow` doing a bunch of cute stuff under the hood to upsample things, without direct knowledge of the field.

With this way we start from the more or less "true"  position of the X-point, which due to interpolation the separatrix does not pass through at the moment. Because we use some interpolation on psi and that means that any turning points occur on a grid point, which the X-point is almost never on. It's a known issue with `find_LCFS_separatrix` that the extrema are not totally correctly captured.

je-cook (approved):
I'm happy @hsaunders1904 any more to add?

hsaunders1904 (approved):
I'm happy 👍

Issue #642, opened 2022-01-25
Add builder for EU-DEMO first wall


hsaunders1904:
## Description of issue / requirement to address

A first wall (with divertor) builder needs to be written and added to the EU-DEMO reactor build process.

## Proposed solution

- Add a `FirstWallBuilder` (or similar) class to
  - build an approximate first wall shape: following flux surface, with a geometric offset.
  - build some divertor shape. An approximate shape is fine for now.
  - cut the wall shape below the x-point, and join with the divertor.
- Add the builder to the `EUDEMOReactor`'s build steps

Further work will involve adding an optimisation step to the first wall builder, but this may be a separate ticket.

je-cook:
Done

PR #680, opened 2022-02-03
Write initial first wall builder

hsaunders1904:
## Linked Issues
#642

## Description

Implements `FirstWallBuilder` class and adds the "First wall" component to the `EUDEMOReactor`.

Three builder classes have been implemented:
- `WallBuilder`: builds a closed wall shape, independent of divertor
- `DivertorBuilder`: builds the divertor components
- `FirstWallBuilder`: runs `WallBuilder`, cuts it off below the x-point, and uses `DivertorBuilder` to make a divertor in the gap.

Also added is the `MinimiseLength` geometry optimisation class, which minimises the length of a shape parameterisation with some constraints.

Some incomplete parts:
- Divertor targets are currently always vertical
- Divertor baffles are currently straight lines
- Only single-null plasmas with x-point at lower end are supported

Another note for reviewers, the function `bluemira.geometry.tools.find_point_along_wire_at_length` is currently implemented using discretisation. Is there a way to do this using primitives?

## Checklist

### DIFF
--- a/bluemira/base/builder.py
+++ b/bluemira/base/builder.py
@@ -28,7 +28,7 @@ from __future__ import annotations
 import abc
 import enum
 import string
-from typing import Dict, List, Literal, Optional, Union
+from typing import Any, Dict, List, Literal, Optional, Union
 
 from bluemira.base.components import Component
 from bluemira.base.error import BuilderError
@@ -88,7 +88,7 @@ class Builder(abc.ABC):
     _params: ParameterFrame
     _design_problem = None
 
-    def __init__(self, params, build_config: BuildConfig, **kwargs):
+    def __init__(self, params: Dict[str, Any], build_config: BuildConfig, **kwargs):
         self._name = build_config["name"]
 
         self._validate_config(build_config)
--- a/bluemira/base/builder.py
+++ b/bluemira/base/builder.py
@@ -96,7 +96,7 @@ class Builder(abc.ABC):
         self._params = ParameterFrame.from_template(self._required_params)
         self.reinitialise(params, **kwargs)
 
-    def __call__(self, **kwargs) -> Component:
+    def __call__(self) -> Component:
         """
         Perform the full build process using the current parameterisation of the builder.
 
@@ -109,7 +109,7 @@ class Builder(abc.ABC):
         if hasattr(self, "_runmode"):
             bluemira_print(f"Designing {self.name} using runmode: {self.runmode}")
             self._runmode(self)
-        return self.build(**kwargs)
+        return self.build()
 
     @abc.abstractmethod
     def reinitialise(self, params, **kwargs) -> None:
--- a/bluemira/base/builder.py
+++ b/bluemira/base/builder.py
@@ -96,7 +96,7 @@ class Builder(abc.ABC):
         self._params = ParameterFrame.from_template(self._required_params)
         self.reinitialise(params, **kwargs)
 
-    def __call__(self) -> Component:
+    def __call__(self, **kwargs) -> Component:
         """
         Perform the full build process using the current parameterisation of the builder.
 
@@ -109,7 +109,7 @@ class Builder(abc.ABC):
         if hasattr(self, "_runmode"):
             bluemira_print(f"Designing {self.name} using runmode: {self.runmode}")
             self._runmode(self)
-        return self.build()
+        return self.build(**kwargs)
 
     @abc.abstractmethod
     def reinitialise(self, params, **kwargs) -> None:
--- a/bluemira/builders/EUDEMO/first_wall/divertor.py
+++ b/bluemira/builders/EUDEMO/first_wall/divertor.py
@@ -23,7 +23,8 @@ Define builder for divertor
 """
 
 import enum
-from typing import Any, Dict, Iterable, List, Tuple
+import operator
+from typing import Any, Callable, Dict, Iterable, List, Tuple
 
 import numpy as np
 
@@ -31,10 +32,10 @@ from bluemira.base.builder import BuildConfig, Builder, Component
 from bluemira.base.components import PhysicalComponent
 from bluemira.base.config import Configuration
 from bluemira.equilibria.equilibrium import Equilibrium
-from bluemira.equilibria.find import find_flux_surface_through_point
+from bluemira.equilibria.find import find_flux_surface_through_point, get_legs
+from bluemira.geometry._deprecated_loop import Loop
 from bluemira.geometry.tools import find_point_along_wire_at_length, make_polygon
 from bluemira.geometry.wire import BluemiraWire
-from BLUEPRINT.nova.stream import StreamFlow
 
 
 class LegPosition(enum.Enum):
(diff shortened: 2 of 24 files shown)


bluemira/builders/EUDEMO/first_wall/divertor.py
```diff
@@ -0,0 +1,329 @@
...
+
+    _required_params = [
+        "div_L2D_ib",
+        "div_L2D_ob",
+        "div_Ltarg",
+        "div_open",
+    ]
+    _required_config: List[str] = []
+    _params: Configuration
+    _default_runmode: str = "mock"
```
DanShort12, line 84: If we're not running a design problem then you shouldn't need to set the default run mode (and shouldn't need to add implementations for run / read /mock).

bluemira/builders/EUDEMO/first_wall/divertor.py
```diff
@@ -0,0 +1,329 @@
...
+            The position of the leg the baffle should join to (e.g.,
+            inner).
+        start: Sequence
+            The position (in x-z) to start drawing the baffle from,
+            e.g., the outside end of a target.
+        end: Sequence
+            The position (in x-z) to stop drawing the baffle, e.g., the
+            postion to the upper part of the first wall.
+        """
+        # Just generate the most basic baffle: a straight line
```
DanShort12, line 244: Perhaps this comment can go into the docstring for the function? Then the docs will also reflect that this is currently making a baffle using straight lines?

bluemira/builders/EUDEMO/first_wall/divertor.py
```diff
@@ -0,0 +1,329 @@
...
+                [point[0], point[0]],
+                [point[1], point[1]],
+                [point[2] - target_length / 2, point[2] + target_length / 2],
+            ]
+        )
+        return PhysicalComponent(label, make_polygon(target_coords))
+
+    def make_dome(self, start: Sequence[float], end: Sequence[float], label: str):
+        """
+        Make a dome between the two given points
```
DanShort12, line 182: Add a bit of a description here that the dome follows the flux lines from the inner target end points.

bluemira/builders/EUDEMO/first_wall/first_wall.py
```diff
@@ -0,0 +1,255 @@
...
+        first_wall (Component)
+        └── xz (Component)
+            └── wall (Component)
+                └── wall_boundary (PhysicalComponent)
+            └── divertor (Component)
+                ├── inner_target (PhysicalComponent)
+                ├── outer_target (PhysicalComponent)
+                ├── dome (PhysicalComponent)
+                ├── inner_baffle (PhysicalComponent)
+                └── outer_baffle (PhysicalComponent)
```
DanShort12, line 93: I like this, we should consider putting this in for other builders.

bluemira/builders/EUDEMO/first_wall/wall.py
```diff
@@ -0,0 +1,231 @@
...
+        self.adjust_variable("lower", lower, lower_bound=0.2, upper_bound=0.7)
+        self.adjust_variable("bottom", bottom, lower_bound=0.05, upper_bound=0.75)
+
+        # Fix 'flat' to avoid drawing the PolySpline's outer straight.
+        # The straight is often optimised to near-zero length, which
+        # causes an error when CAD tries to draw it
+        self.fix_variable("flat", 0)
+
+
+class WallPrincetonD(PrincetonD):
```
DanShort12, line 103: I've ummed and ahhed a bit over these builder-specific shapes. On one hand I like that they are tuned to the details of the build, but in principle this could be done from the configuration and tuned there... perhaps we can leave this in and see if it works well as an approach (particularly if the bounds need to be relative (if they do, then it could be worth us considering allowing derived replacements in the config)).

bluemira/builders/EUDEMO/first_wall/wall.py
```diff
@@ -0,0 +1,231 @@
...
+                self.COMPONENT_WALL_BOUNDARY,
+                BluemiraWire(self.boundary, label=self.COMPONENT_WALL_BOUNDARY),
+            )
+        )
+        component.plot_options.wire_options["color"] = BLUE_PALETTE["DIV"]
+        bm_plot_tools.set_component_plane(component, "xz")
+        return component
+
+    # TODO(hsaunders1904): 'height' is not a parameter for all shapes.
+    # Does it make sense to derive it within this class?
```
DanShort12, line 218: I think this links to my point above about supporting derived / calculated values dynamically in the config. It might also be that some problems need an offset if e.g. this polyspline height ends up within the keep out zone.

bluemira/geometry/wire.py
```diff
@@ -220,3 +220,17 @@ def change_placement(self, placement):
                 change_placement(o, placement._shape)
             else:
                 o.change_placement(placement)
+
+    def start_point(self) -> np.ndarray:
+        """
+        Get the coordinates of the start of the wire.
+        """
+        point = self.boundary[0].Vertexes[0].Point
```
DanShort12, line 228: I'll have a think about this. We're trying to keep all the FreeCAD bits within the FreeCAD API (Vertexes and Point are FreeCAD objects, so in principle may not exist if we were to use some other CAD backend).

DanShort12 (commented):
I know there's more to come, but thought to get things rolling with some initial comments while this is getting towards done.

bluemira/builders/EUDEMO/first_wall/divertor.py
hsaunders1904: I'd added those methods because I was getting `NotImplementedError`s. I see now that that's because the `runmode` for the `FirstWallBuilder` was being propagated to the divertor builder. 

It could be handled by just popping 'runmode` before we pass the config through. It feels like there should be a more automatic way. I'll have a think

bluemira/builders/EUDEMO/reactor.py
```diff
@@ -42,13 +43,21 @@ class EUDEMOReactor(Reactor):
...
         """
         Run the EU-DEMO reactor build process. Performs the following tasks:
 
         - Run the (PROCESS) systems code
-        - Build the Plasma
-        - Build the TF Coils
+        - Build the 'Plasma'
+        - Build the 'TF Coils'
+        - Build the 'PF Coils'
+        - Build the 'First Wall'
```
CoronelBuendia, line 60: Do we need this? The code is actually pretty clear in this method.

CoronelBuendia (commented):
A few comments for now.. still havent found the time to play around with it - sorry!

bluemira/builders/EUDEMO/first_wall/wall.py
CoronelBuendia: spun out in #765

bluemira/builders/EUDEMO/first_wall/wall.py
CoronelBuendia: spun out in #765

bluemira/base/builder.py
```diff
@@ -88,7 +88,7 @@ class Builder(abc.ABC):
     _params: ParameterFrame
     _design_problem = None
 
-    def __init__(self, params, build_config: BuildConfig, **kwargs):
+    def __init__(self, params: Dict[str, Any], build_config: BuildConfig, **kwargs):
```
CoronelBuendia, line 91: still kwargs in here; is this right?

CoronelBuendia: just this now from my side; otherwise all good!

hsaunders1904: Things break when I remove them. The issue is that the constructors in the concrete builder classes delegate most of the work to reinitialise. Without the kwargs, they can't pass through the parameters.

Dan and I have talked about this and he's going to open an issue to look into alternatives to `reinitialise`. 

Are we happy to leave this as is for the moment, then remove the kwargs when we address what to do with `reinitialise`? It's proving a bit of a rabbit-hole implementing a work-around.

CoronelBuendia: OK, fair does

CoronelBuendia (approved):
Several issues spun off but best to get this in

PR #681, opened 2022-02-03
Stop json files from being continually generated in CLI tests

DanShort12:
## Linked Issues

N/A

## Description

Moves config generation into a class setup. Clears up the files when done. Sorry the diff is rather opaque!

## Interface Changes

N/A

## Checklist


je-cook (changes requested):
You weren't kidding that diff is nigh on impenetrable!

Mostly collecting the tests into one class other than the json regen?

There are quite a few repetitions of 
`flags = ["-i", INDIR, "-ri", REACTORNAME, "-o", outdir_flag]` maybe with the variable name slightly different followed by 
`runner.invoke(..)` and the assert. Can we factor that out into a function?

DanShort12:
I used the collection of the tests into a class to handle controlling the regeneration of the json files, but can break that out into separate actions. I'll redo this so that it's easier to see how things move around.

Good suggestion on breaking out the invocation though.

je-cook:
> I used the collection of the tests into a class to handle controlling the regeneration of the json files, but can break that out into separate actions. I'll redo this so that it's easier to see how things move around.
> 

I prefer the class handler to be honest as they are all doing work on the same set of objects, if you want to thats ok but I think I got there in the end.

Just the invocation change and I think its mergeable

DanShort12:
Ah, sorry, was part way through breaking down the commits yesterday but didn't get round to pushing them. Should be easier to review per commit now, and it's also now got the refactor to clean up how the runner gets invoked.

tests/BLUEPRINT/test_blueprint_cli.py
```diff
@@ -106,425 +100,461 @@ def wrapper_grouped_decorator(*args, **kwargs):
...
+    def test_datadir(
+        self,
+        mock_rclass,
+        mock_build,
+        mock_plot_xz,
+        mock_plot_xy,
+        mock_save_CAD_model,
+        tempdir,
+    ):
+        runner = CliRunner()
```
je-cook, line 538: This isnt used (sonarcloud not me).

je-cook (approved):
Just one minor and then I'll merge

tests/BLUEPRINT/test_blueprint_cli.py
DanShort12: Should be sorted in the latest push (I wonder if there's a way to configure sonarcloud to exclude patterns in its regex for variable / method names).

je-cook: Maybe somehow we can parse the flake8 config because it seems like they're the same name issues as flake has

PR #684, opened 2022-02-03
Geometry module  __all__ for some files

CoronelBuendia:
## Linked Issues

Reduces a bit of namespace noise in IDEs and should protect namespaces better in general

## Description

Adds __all__ to some files in geometry in which only a single class resides. Such that now one should not see `from bluemira.geometry.face import [BluemiraWire] as a suggestion.

## Interface Changes
N/A

## Checklist

### DIFF
--- a/bluemira/geometry/bound_box.py
+++ b/bluemira/geometry/bound_box.py
@@ -26,6 +26,8 @@ from dataclasses import dataclass
 
 import numpy as np
 
+__all__ = ["BoundingBox"]
+
 
 @dataclass
 class BoundingBox:
--- a/bluemira/geometry/face.py
+++ b/bluemira/geometry/face.py
@@ -38,6 +38,8 @@ from bluemira.geometry.base import BluemiraGeo
 from bluemira.geometry.error import DisjointedFace, NotClosedWire
 from bluemira.geometry.wire import BluemiraWire
 
+__all__ = ["BluemiraFace"]
+
 
 class BluemiraFace(BluemiraGeo):
     """Bluemira Face class."""
--- a/bluemira/geometry/plane.py
+++ b/bluemira/geometry/plane.py
@@ -30,6 +30,8 @@ import numpy as np
 import bluemira.codes._freecadapi as cadapi
 from bluemira.geometry.error import GeometryError
 
+__all__ = ["BluemiraPlane"]
+
 
 class BluemiraPlane:
     """
--- a/bluemira/geometry/shell.py
+++ b/bluemira/geometry/shell.py
@@ -32,6 +32,8 @@ import bluemira.codes._freecadapi as cadapi
 from bluemira.geometry.base import BluemiraGeo
 from bluemira.geometry.face import BluemiraFace
 
+__all__ = ["BluemiraShell"]
+
 
 class BluemiraShell(BluemiraGeo):
     """Bluemira Shell class."""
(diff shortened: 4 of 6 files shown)


je-cook (approved):

ivanmaione:
@CoronelBuendia, @je-cook, @DanShort12, not sure if it is a good approach to suggest a user (or a developer) import BluemiraWire from face.py (just to refer to the example reported in the description of this PR).
My preferred solution would instead be to use the __init__.py of each module to create the conventional public API exposed by the module. In this specific case, I would prefer to add in geometry/__init__.py all the necessary imports:
```
from .bound_box import BoundingBox
from .coordinates import Coordinates
from .wire import BluemiraWire
from .face import BluemiraFace
from .shell import BluemiraShell
from .plane import BluemiraPlane
...
```
and then, when it is necessary, do the import as:
1) for internal files
```
from . import BluemiraWire
```
2) for external files
```
import bluemira.geometry as geo
bmwire = geo.BluemiraWire(.....)
```
I know that the suggest approach could bring to the usual cyclic import problems, but probably we can solve with the TYPE_CHECKING.

Another alternative solution, if the problem is just for geometry, would be to join all the files related to geometries (i.e. wires.py, face.py, etc.) into the base.py (so, all in a file, like we have for the Components).

What are your thoughts?

CoronelBuendia:
The only thing that matters for me is that the user can import things from a certain module without needing to know exactly where such a thing might live inside the module.

So, I also think that certain key classes and functions should be imported inside `__init__.py` such that they are accessible at the module level.

We are however a long way off sorting out these circular import issues, and until such a time as we can actually expose everything at the module level, all this PR is doing is protecting the namespaces from `from somewhere import *`. We'd probably keep these `__all__` fields even with importing everything at the module level. Pretty sure most modern IDEs will pick them up when you change the name of a class or function, too.

I was hoping that defining __all__ would also stop IDEs from showing that you can import `BluemiraGeo` from e.g. `bluemira.geometry.face`, but sadly my VS Code doesn't seem to check all before showing me the options of what I can import from a file. Perhaps I have my settings wrong..

I think that relative imports inside modules are unnecessary and we actually had to get rid of them in the past because they were causing a lot of pain when we were moving things around. In practice, from the user perspective, it makes no difference if we use full path imports or relative imports. One is more explicit and the other less so. With modern IDEs I honestly don't see the disadvantage of full path imports. The user certainly will not care; it's about what is easier for us to manage, and we have been burnt trying to use relative imports everywhere.

I don't have any strong opinions on `from somewhere import something as sw` + `sw.Something(...)` or `from somewhere import Something` + `Something(...)`

je-cook:
It looks like there is a typo in the description. Looking at the code it should read `from bluemira.geometry.face import [BluemiraFace]`

I agree with @CoronelBuendia and also think relative imports reduce readability for the sake of a few characters and have no major benefit.
Imports in `__init__.py` should be only added when we feel the module is complete and only in very specific circumstances. Its very easy to get to a state where imports through init become circular especially in modules used in many places. I see them as  directions for how to use the module for users (advanced users maybe), not really developers

ivanmaione:
@je-cook, as also clarified with @CoronelBuendia, there was a misunderstood due to the typo in the description (that suggested maintaining the behavior that, in reality, Matti was trying to avoid :) ). I still think that, due to the short length of all the geometry classes, we could join all the files in just one, but maybe we can discuss it another time. In conclusion, nothing in contrary to this PR. Going to merge it.

PR #686, opened 2022-02-03
Add unit conversions

je-cook:
## Description

Add in unit conversions ~to/from codes~ into and out of parameters and make sure our constants are the right values (according to pint).
Also collects all unit conversion functions (may be to be put in its own file?)

Update: This PR makes Parameter and ParameterFrame unit aware. I'm going to open a new PR for the codes stuff otherwise this will be huge!

## Interface Changes

N/A

## Checklist


bluemira/base/parameter.py
```diff
@@ -326,6 +344,18 @@ def source_history(self):
...
+    @property
+    def unit(self):
+        """
+        The unit of the parameter
+        """
+        return self._unit
+
+    def _unit_setup(self, unit: Union[Unit, str, None]):
+        if isinstance(unit, None):
+            bluemira_warn(f"{self.var} has no unit")
```
CoronelBuendia, line 356: This is going to happen a lot... Could we define a NO_UNIT unit or something? We have so many fractions, ratios, strings, and the like, and I think it will be quite easy to lose consistency between ["n/a", "N/A" and whatever]. Perhaps you've already got something planned!

bluemira/base/constants.py
```diff
@@ -42,68 +59,192 @@
...
+    Returns
+    -------
+    flow_in_mols: Union[float, np.array]
+        The flow in mol/s
+
+    Notes
+    -----
+    At 273.15 K for a diatomic gas
+    """
+    return flow_in_pam3_s / 2270
```
CoronelBuendia, line 225: I will try and find my workings for this gem... probably buried in a document somewhere, sorry!

requirements.txt
```diff
@@ -55,7 +55,9 @@ parso==0.8.3
 pexpect==4.8.0
 pickleshare==0.7.5
 Pillow==9.0.0
+Pint==0.18
 pluggy==1.0.0
+periodictable==1.6.0
```
CoronelBuendia, line 60: Just a note to our future selves: we're effectively importing constants from these repositories, and any changes to those constants (which are likely to be small, but nevertheless!) will change our results.

We should exercise caution when accepting updates to these modules, to the point where I'm almost considering adding a test so that we can detect any changes... I know that sounds mad, but these changes here have the potential to cause less than negligible variations in results.

bluemira/base/constants.py
```diff
@@ -42,68 +59,192 @@
...
+    ureg.Quantity("m_e").to("g") * ureg.Quantity("avogadro_constant").to_base_units()
+).magnitude
 
-# Deuterium molar mass
-D_MOLAR_MASS = 2.014102  # [u] or [g/mol]
+# Tritium half-life
+# https://www.nist.gov/pml/radiation-physics/radioactivity/radionuclide-half-life-measurements
+# http://www.lnhb.fr/nuclear-data/nuclear-data-table/
+# http://www.lnhb.fr/nuclides/H-3_tables.pdf
+T_HALFLIFE = 12.312  # [yr]
```
CoronelBuendia, line 92: Ah ha! Interesting... this will change results, although I don't think we have any proper regression tests on the fuel cycle model at present.

CoronelBuendia (commented):
Nice to see some effort go into this; it's going to be pretty important, I reckon!

bluemira/base/parameter.py
je-cook: There is a builtin `'dimensionless'` keyword that is equivalent to `""`. This was mean't to be a protection against no unit specified at all. I havent converted existing units yet

bluemira/equilibria/file.py
```diff
@@ -91,18 +91,18 @@ class EQDSKInterface:
         ["bcentre", "Magnetic field at the reference radius", None, "T", None, "eqfile"],
         ["cplasma", "Plasma current", None, "A", None, "eqfile"],
         ["fpol", "Poloidal current function f = R*B on 1-D flux grid", None, "T.m", None, "eqfile"],
-        ["pressure", "Plasma pressure function on 1-D flux grid", None, "n.t/m^2", None, "eqfile"],
+        ["pressure", "Plasma pressure function on 1-D flux grid", None, "N.t/m^2", None, "eqfile"],
         ["ffprime", "FF' function on 1-D flux grid", None, "m.T^2/V.s/rad", None, "eqfile"],
-        ["pprime", "P' function on 1-D flux grid", None, "n.t/m^2/V.s/rad", None, "eqfile"],
+        ["pprime", "P' function on 1-D flux grid", None, "N.t/m^2/V.s/rad", None, "eqfile"],
```
je-cook, line 96: `<Unit('metric_ton * newton * second / meter ** 2 / radian / volt')>`

BLUEPRINT/systems/plasma.py
```diff
@@ -321,7 +321,7 @@ def calc_p_lh(self):
             The L-H transition power [MW]
         """
         a = self.params.R_0 / self.params.A
-        return P_LH(self.params.rho * 1e20, self.params.B_0, a, self.params.R_0)
+        return P_LH(self.params.rho * 1e21, self.params.B_0, a, self.params.R_0)
```
hsaunders1904, line 324: Apologies if my maths is wrong, but are we off by a factor of ten here?

We used to have:
```
0.8 . 1e20 . 1e20 m^-3 
  = 8 . 1e39 m^-3
```
we now have:
```
80 . 1e21 . um^-3 = 8 . 1e22 . 1e18 . m^-3
  = 8 . 1e40 m^-3
```

bluemira/base/constants.py
```diff
@@ -42,68 +64,218 @@
...
+
+
+def from_keV(
+    value: Union[float, np.array, List[float]],
+    to: str = "celsius",
+    *,
+    _from: str = "keV"
+):
+    """
+    Convert a temperature in Celsius to Kelvin.
```
hsaunders1904, line 185:
```suggestion
    Convert a temperature in keV to Celsius.
```

Out of interest, why do we default to `Celsius` here rather than `Kelvin`?

bluemira/base/constants.py
```diff
@@ -42,68 +64,218 @@
...
+YR_TO_S = ureg.Quantity(1, ureg.year).to("s").magnitude
 
 # Seconds to years
-S_TO_YR = 1 / YR_TO_S
+S_TO_YR = ureg.Quantity(1, ureg.second).to("year").magnitude
+
+
+def to_celsius(kelvin: Union[float, np.array, List[float]]) -> Union[float, np.array]:
+    """
+    Convert a temperature in Celsius to Kelvin.
```
hsaunders1904, line 142:
```suggestion
    Convert a temperature in Kelvin to Celsius.
```

bluemira/base/constants.py
```diff
@@ -42,68 +64,218 @@
...
+    Notes
+    -----
+    At 273.15 K for a diatomic gas
+    """
+    return flow_in_pam3_s / 2270
+
+
+def mols_to_pam3s(flow_in_mols):  # noqa :N802
+    """
+    Convert a flow in Pa.m^3/s to a flow in mols.
```
hsaunders1904, line 261:
```suggestion
    Convert a flow in mols to a flow in Pa.m^3/s.
```

bluemira/base/parameter.py
```diff
@@ -239,6 +274,36 @@ def __init__(
             self._value_history = []
             self._source_history = []
 
+    def __dir__(self):
+        """
+        Add missing methods
+        """
+        return sorted(
```
hsaunders1904, line 281: Aren't `set`s already sorted? Would a straight conversion to `list` work?

bluemira/equilibria/equilibrium.py
```diff
@@ -524,7 +524,7 @@ class Equilibrium(MHDState):
     Ip: float (optional) default = 0
         Plasma current [A]
     li: None or float (default None)
-        Normalised plasma internal inductance [N/A]
+        Normalised plasma internal inductance [-]
```
hsaunders1904, line 527: What does `-` mean in this context? Is it different from `'dimensionless'`?

hsaunders1904 (commented):
Looks good to me 👍. A couple of minor things I spotted

bluemira/base/parameter.py
je-cook: yes you're right, I will just use list

bluemira/base/constants.py
je-cook: The reason is really just because, I'm wondering whether we change it to kelvin generally, it seems more of a standard thing to use. @CoronelBuendia any thoughts?

je-cook: The reason is really just because, I'm wondering whether we change it to kelvin generally, it seems more of a standard thing to use. @CoronelBuendia any thoughts?

BLUEPRINT/systems/plasma.py
je-cook: Yep it should still be 1e20 to keep the same thing, weird conversion in the first place as we would now be in `1/pm^-3`

je-cook: sorry I think it should be 1e18 I think because I am storing it under a different unit (so I can use a prefix) therefore the unit itself doesnt come into play

je-cook: probably should just do this: `ureg.Quantity(80, "um^-3").to("m^-3").magnitude`. I will add a converter function somewhere

je-cook: in the previous version the units were specified as a string `10**20 / m^3`  which is not the easiest thing to do in pint because it is not a multiple of 3^3 -> 21 (for the prefix) and pint doesnt usually allow a raw number in a unit (although possible if you want to muck about a bit). 

```python
In [1]: import pint

In [2]: ureg = pint.UnitRegistry(fmt_locale="en_GB")

In [3]: (ureg.Quantity(0.8, "m^-3") * 1e20).to("um^-3")
Out[3]: 79.99999999999999 <Unit('1 / micrometer ** 3')>  # floating point but its the same
```
But the actual value that was passed into the above function was `0.8*1e20` which needs to be the same raw number for the function to act the same. As `self.params.rho == 80` in the new version we need to convert the raw number. Therefore `80*1e18 == 0.8*1e20`. The conversion happens to be the same as `ureg.Quantity(80, "um^-3").to("m^-3").magnitude`

Does that make sense or am I rambling :slightly_smiling_face: ?

hsaunders1904: Yeah, that makes sense. The physical value of `rho` hasn't changed, but the value passed to the function is 100 times more, so we need to multiply by 100 times less here

hsaunders1904 (approved):
There's still the question of whether `from_` temperature converters should default to Celsius/Kelvin, but that's potentially not a priority. I'd be tempted to require that users of the function specify what unit to convert to, then it's explicit at the point of use.

Code changes look good to me ✔️

bluemira/base/config.py
```diff
@@ -38,43 +38,43 @@ class Configuration(ConfigurationSchema, ParameterFrame):
...
-        ['n_PF', 'Number of PF coils', 6, 'N/A', None, 'Input'],
-        ['n_CS', 'Number of CS coil divisions', 5, 'N/A', None, 'Input'],
-        ['TF_ripple_limit', 'TF coil ripple limit', 0.6, '%', None, 'Input'],
-        ['Av', 'Reactor availability', 0.3, 'N/A', None, 'Input'],
-        ['A', 'Plasma aspect ratio', 3.1, 'N/A', None, 'Input'],
+        ['blanket_type', 'Blanket type', 'HCPB', 'dimensionless', None, 'Input'],
+        ['n_TF', 'Number of TF coils', 16, 'dimensionless', None, 'Input'],
+        ['n_PF', 'Number of PF coils', 6, 'dimensionless', None, 'Input'],
+        ['n_CS', 'Number of CS coil divisions', 5, 'dimensionless', None, 'Input'],
+        ['TF_ripple_limit', 'TF coil ripple limit', 0.6, 'dimensionless', None, 'Input'],
```
CoronelBuendia, line 54: and here

bluemira/fuel_cycle/lifecycle.py
```diff
@@ -63,8 +63,8 @@ class LifeCycle:
         ["blk_2_dpa", "Second blanket life limit (EUROfer)", 50, "dpa", "https://iopscience.iop.org/article/10.1088/1741-4326/57/9/092002/pdf", "Input"],
         ["div_dpa", "Divertor life limit (CuCrZr)", 5, "dpa", "https://iopscience.iop.org/article/10.1088/1741-4326/57/9/092002/pdf", "Input"],
         ["vv_dpa", "Vacuum vessel life limit (SS316-LN-IG)", 3.25, "dpa", "RCC-Mx or whatever it is called", "Input"],
-        ["tf_fluence", "Insulation fluence limit for ITER equivalent to 10 MGy", 3.2e21, "n/m^2", "https://ieeexplore.ieee.org/document/6374236/", "Input"],
-        ["tf_ins_nflux", "TF insulation peak neutron flux", 1.4e13, "n/m^2/s", "Pavel Pereslavtsev sent me an email 20/02/2017", "Input"],
+        ["tf_fluence", "Insulation fluence limit for ITER equivalent to 10 MGy", 3.2e21, "N/m^2", "https://ieeexplore.ieee.org/document/6374236/", "Input"],
+        ["tf_ins_nflux", "TF insulation peak neutron flux", 1.4e13, "N/m^2/s", "Pavel Pereslavtsev sent me an email 20/02/2017", "Input"],
```
CoronelBuendia, line 67: These are in fact neutrons, not Newtons. Can simplify to just 1/m^2 and 1/m^2/s

examples/BLUEPRINT/nova/curved_picture_frame_example.py
```diff
@@ -38,23 +38,23 @@
...
-    ["n_TF", "Number of TF coils", 12, "N/A", None, "Input"],
+    ["n_TF", "Number of TF coils", 12, "dimensionless", None, "Input"],
     ["tk_tf_nose", "TF coil inboard nose thickness", 0.17, "m", None, "Input"],
     ['tk_tf_side', 'TF coil inboard case minimum side wall thickness', 0.02, 'm', None, 'Input'],
     ["tk_tf_wp", "TF coil winding pack thickness", 0.569, "m", None, "PROCESS"],
     ["tk_tf_ins", "TF coil ground insulation thickness", 0.008, "m", None, "Input"],
     ["tk_tf_insgap", "TF coil WP insertion gap", 1.0E-7, "m", "Backfilled with epoxy resin (impregnation)", "Input"],
     ["r_tf_in", "Inboard radius of the TF coil inboard leg", 0.148, "m", None, "PROCESS"],
-    ["TF_ripple_limit", "Ripple limit constraint", 0.65, "%", None, "Input"],
+    ["TF_ripple_limit", "Ripple limit constraint", 0.65, "dimensionless", None, "Input"],
```
CoronelBuendia, line 48: 0.0065 (note that we treat the number in % in the actual solvers)

examples/BLUEPRINT/nova/how_to_design_tfcoils.py
```diff
@@ -59,15 +59,15 @@
...
+    ["n_TF", "Number of TF coils", 16, "dimensionless", None, "Input"],
     ["rho_j", "TF coil WP current density", 18.25, "MA/m^2", None, "Input"],
     ["tk_tf_nose", "TF coil inboard nose thickness", 0.6, "m", None, "Input"],
     ["tk_tf_wp", "TF coil winding pack thickness", 0.5, "m", None, "PROCESS"],
     ["tk_tf_front_ib", "TF coil inboard steel front plasma-facing", 0.04, "m", None, "Input"],
     ["tk_tf_ins", "TF coil ground insulation thickness", 0.08, "m", None, "Input"],
     ["tk_tf_insgap", "TF coil WP insertion gap", 0.1, "m", "Backfilled with epoxy resin (impregnation)", "Input"],
     ["r_tf_in", "Inboard radius of the TF coil inboard leg", 3.2, "m", None, "PROCESS"],
-    ["ripple_limit", "Ripple limit constraint", 0.6, "%", None, "Input"],
+    ["ripple_limit", "Ripple limit constraint", 0.6, "dimensionless", None, "Input"],
```
CoronelBuendia, line 70: .

examples/fuel_cycle/EUDEMO_fuelcycle.py
```diff
@@ -83,7 +83,7 @@
     ["div_dpa", "Divertor life limit (CuCrZr)", 5, "dpa",
      "https://iopscience.iop.org/article/10.1088/1741-4326/57/9/092002/pdf", "Input"],
     ["vv_dpa", "Vacuum vessel life limit (SS316-LN-IG)", 3.25, "dpa", "RCC-Mx or whatever it is called", "Input"],
-    ["tf_fluence", "Insulation fluence limit for ITER equivalent to 10 MGy", 3.2e21, "n/m^2",
+    ["tf_fluence", "Insulation fluence limit for ITER equivalent to 10 MGy", 3.2e21, "N/m^2",
```
CoronelBuendia, line 86: 1/m^2

CoronelBuendia (changes requested):
Just some issues with % and neutron vs Newton.

I'd like to check the flux function units again

I also see sweeping changes in parameter that I haven't had a chance to look at.

Thought I'd get these comments in first

BLUEPRINT/systems/tfcoils.py
```diff
@@ -75,18 +75,18 @@ class ToroidalFieldCoils(Meshable, ReactorSystem):
         ['tk_tf_case_out_out', 'TF coil case thickness on the outboard outside', 0.4, 'm', None, 'Calc'],
         ['tf_wp_width', 'TF coil winding pack radial width', 0.76, 'm', 'Including insulation', 'PROCESS'],
         ['tf_wp_depth', 'TF coil winding pack depth (in y)', 1.05, 'm', 'Including insulation', 'PROCESS'],
-        ['tk_tf_outboard', 'TF coil outboard thickness', 1, 'm', None, 'Input', 'PROCESS'],
+        ['tk_tf_outboard', 'TF coil outboard thickness', 1, 'm', None, 'PROCESS'],
         ['r_tf_in', 'Inboard radius of the TF coil inboard leg', 3.2, 'm', None, 'PROCESS'],
-        ['TF_ripple_limit', 'TF coil ripple limit', 0.6, '%', None, 'Input'],
+        ['TF_ripple_limit', 'TF coil ripple limit', 0.6, 'dimensionless', None, 'Input'],
```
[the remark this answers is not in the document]
je-cook: I wondered how many times that will happen.. I'll change some stuff about

examples/BLUEPRINT/nova/curved_picture_frame_example.py
je-cook: ok good to know

bluemira/base/parameter.py
```diff
@@ -872,15 +1023,18 @@ def add_parameters(self, record_list, source=None):
...
+                    self.add_parameter(self.modify_source(param, source))
                 else:
-                    self.add_parameter(*self.modify_source(source, param))
+                    plen = len(Parameter.__slots__)
+                    if len(param) not in [plen - 1, plen - 2, plen + 1]:
+                        raise ValueError
+                    self.add_parameter(*self.modify_source(param, source))
 
-    def set_parameter(self, var, value, source=None):
+    def set_parameter(self, var, value, unit=None, source=None):
```
je-cook, line 1037: Note for me, I havent set this up to be used

tests/BLUEPRINT/cad/test_shieldCAD.py
```diff
@@ -40,7 +40,7 @@ def test_thermalshield(self):
         v = ThermalShieldCAD(self.thermal_shield, neutronics=False)
         volume = v.get_properties()["thermal_shield"]["Volume"]
         # This will change if you change the build pattern in TSCAD
-        assert np.isclose(round(volume, 2), 6.02)
+        assert np.isclose(round(volume, 2), 5.96)
```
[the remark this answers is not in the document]
je-cook: With percentage changes this is now closer (same for vesselCAD) still not the same though

bluemira/equilibria/file.py
CoronelBuendia: To be checked [action matti]

bluemira/base/constants.py
CoronelBuendia: To be checked by Tiago

tplobo: @je-cook 
Hey James, here's the conversion you requested:
Screenshot 2022-03-01 at 14 56 10

Basically, [Pa.m3] can be converted to [mol] using a normalization at "normal temperature", which I had to check. But that's 0ºC and you already have that in the script (T_N = 273.15K). This makes the conversion factor to be:

8.3144598 * 273.15 ~ 2271.1

je-cook: Thanks very much!

bluemira/equilibria/file.py
CoronelBuendia: yeah not sure where the t came in. Looks to me like it should be N/m^2/V.s/rad

tests/BLUEPRINT/cad/test_shieldCAD.py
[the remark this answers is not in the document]
CoronelBuendia: Yeah I'm a little concerned these changes are not just butterfly effects. I would have thought that they wouldn't change at all. I have to say though, I don't think it is worth digging down into these (a bit of a mammoth task) because this is for old BLUEPRINT which we're moving away from. They are also not massive so perhaps we can let them slide... I was kind of hoping the change to percent for ripple would fix this.

bluemira/base/constants.py
CoronelBuendia: Agreed as K across the board

CoronelBuendia (approved):

bluemira/base/constants.py
tplobo: Shouldn't this say "mol/s" insteal of "mols"?

je-cook: Yep, I'll add it in #748

PR #589, opened 2022-01-11
Add EU-DEMO cryostat thermal shield

CoronelBuendia:
## Linked Issues
Requires #702 
Closes #582 (will pick up VVTS and ports in a future issue

## Description

Adds a thermal shield builder, which at present just builds the cryostat TS.

## Interface Changes

N/A

## Checklist

I confirm that I have completed the following checks:

- [x] Tests run locally and pass `pytest tests --reactor`
- [x] Code quality checks run locally and pass `flake8` and `black .`
- [x] Documentation built locally and checked `sphinx-build -W documentation/source documentation/build`

PR #687, opened 2022-02-04
Add `fix_face` and add into `BluemiraFace` creation

### LINKED ISSUE #682: Add `fix_face` and include when faces are created

We've come across an issue in #589 in which a `BluemiraFace` is created from an allegedly valid `BluemiraWire` but the resulting face is not valid. It was found that calling `.fix(float, float, float)` on this fixed the face.

## Proposed solution

Add a `fix_face` function in `_freecadapi.py` and use it (under the hood) when `BluemiraFace`s are created. Warn if the resulting face required fixing.

I am not proposing to make this a user-facing function for the time being.

## Alternative solutions

Haven't come across any good alternatives. Given that this can happen with a wire that is supposedly valid means we will need some form of detection and protection against these cases sooner or later.

CoronelBuendia:
## Linked Issues

Closes #682 

## Description

Adds `fix_face` and puts into `BluemiraFace._create` with a warning if a face is invalid and another one if the fix didn't help. A bit noisy probably because we're recreating the faces each time.

Removes a test (perhaps controversially) on a mixed face maker which just fails to be fixed because the underlying wire is so bad:

image
image

## Interface Changes

None, really.. but under the hood behavioural changes.

## Checklist


bluemira/codes/_freecadapi.py
```diff
@@ -1123,6 +1123,22 @@ def fix_wire(wire, precision=EPS, min_length=MINIMUM_LENGTH):
     wire.fix(precision, min_length, min_length)
 
 
+def fix_face(face, precision=EPS, min_length=MINIMUM_LENGTH):
```
DanShort12, line 1126: I don't think we need to be specific about this fixing a face - the `fix` method lives on `Part.Shape`, so any of the apiWire/apiFace/apiShell/apiSolid would work in this case. Perhaps we can just rename the one above to `fix_shape` and update the docstring to accept any shape?

tests/bluemira/geometry/test_data/div_test_mfm.json
```diff
@@ -1 +0,0 @@
....214413818903267, -6.190877659711987, -6.190877655521035, -6.176851005293429, -6.169908673502505, -6.153458732645959, -6.129952391143888, -6.112467097584158, -6.10947066033259, -6.10052112583071, -6.10052112210542, -6.05489255907014, -6.032431241124868, -6.006110856775194, -5.992417314089835, -5.970814513973892, -5.970814512111247, -5.960382487159222, -5.952377453446388, -5.952377453446388, -5.932107432279736, -5.932107430417091, -5.9150836849585176, -5.913261932786554, -5.908507171086967, -5.908507169224322, -5.889630299527198, -5.88963029673323, -5.8707534284330904, -5.870753427967429, -5.863856318872422, -5.858658609446138, -5.842298134230077, -5.842298134230077, -5.824656302575022, -5.824656302109361, -5.8126614657230675, -5.804456404410303, -5.791096241679043, -5.774437479209155, -5.774437476880848, -5.537956531625241, -5.325361884664744, -5.332005923148245, -5.344397844746709, -5.3565479423850775, -5.368418048135936, -5.3800504435785115, -5.3913968205451965, -5.402500879019499, -5.413324547465891, -5.423896882217377, -5.434193668887019, -5.444232070352882, -5.453998758923262, -5.463502215221524, -5.4727365602739155, -5.481705261860043, -5.478971095290035, -5.491773025598377]}
```
DanShort12, line 1: I guess this test no longer works with the face being fixed if invalid?

DanShort12 (commented):
Noticed this looked ready but hadn't been assigned for a review yet. Since it's mentioned in #589 I figured I'd put a few comments to it if that helps to move it along.

tests/bluemira/geometry/test_data/div_test_mfm.json
CoronelBuendia: Yeah this Loop is just bad (it has overlapping lines) so a valid face can't be made from it.

je-cook (approved):
Happy here

ivanmaione (approved):

je-cook:
may not be needed after the refactor of geometry

CoronelBuendia:
Closed because will be obsolete with #1650

PR #701, opened 2022-02-08
Install dependencies from source

### LINKED ISSUE #504: Environment encoding sometimes set implicitly to ASCII instead of utf8

DanShort12:
## Linked Issues

Closes #504 (hopefully)

## Description

Provides scripts to build and install dependencies from source. This allows us to use a more recent PySide2 version than is available on package managers, which in turn should hopefully fix some encoding issues people have seen, amongst other conda-related issues.

## Interface Changes

N/A

## Checklist


DanShort12:
Scripts should be pretty much ready for a first pass - works locally on a clean env.

A few thoughts on improvements:

- [ ] add an option to set the version of the code being installed
- [ ] add an option to set whether debug or release build is used
- [ ] add a check whether running on linux or osx
- [ ] add a Dockerfile (more likely to follow in a separate PR

je-cook (commented):
Just looking through without trying to install it looks good (and easy to follow!). 

What was the reason behind sticking with qt 5.14.2?

Also I see that when we (very occasionally :crossed_fingers: ) have to update this we should probably note somewhere what the version are and where they are set.

I think once we're happy with it all a Docker container and linking to the CI would be good. Because then we can get everyone to be using the same env.
Possibility of supporting both conda and Docker will probably happen in the short term. Although I would hope that wouldnt be for too long...

DanShort12:
TBH I haven't tried a more recent qt version yet. I see that the latest qt5 version is 5.15.2 so could run a build with that and see what breaks. I could also put in an update so that the version of the dependency being checked out is an argument to the scripts - that way the version would be explicit whenever the script is run.

I've pretty much got a first pass of a working docker image. It's pretty huge and I haven't got the python OCC stuff in there yet, and will grow when we end up with gmsh, fenics, etc. as further dependencies. I'd quite like to make it so that only the necessary runtime libraries are passed into the final image and had a bit of a go at that this morning, but while all the linking paths seemed fine I end up with a hard crash in the C++. Possibly will be a case of initially testing peoples internet connections and then trimming down in the future.

je-cook (commented):
Could be nice to have a docker-compose file to include the required flags to eg forward X11

DanShort12:
Needs updating to include new dependencies (meshio, gmsh, fenics)

je-cook:
Updated to latest Ubuntu.

Ubuntu 22.04 no longer packages sub versions of python so this is built on the default python (3.10), there are some hacks at the end because of this that need to be done with a couple of packages that aren't updated for 3.10 yet.

I've also patched the qt5 build in the same way that kde have because of a gcc compilation problem. Possilbe other solution is to pull from their repo.

Main remaining issue other than that is getting pivy to install properly, it builds but crashes when I try to run any CAD viewer (same as #1020), pivy==0.6.5 (#1048) has some run issues about circular importing. If I pull across the #1033 I can view things however so it feels like the last hurdle.

As far as I'm aware all other dependencies are now included, I havent run pytest yet though, just the EUDEMO example.
This should stay WIP until the hacks bash script can be removed.

je-cook:
docker commands for the interested. Will add a docker compose file with this is at some point:
`sudo docker build -t bluemira:dev .`
`sudo docker run --rm -it --volume `pwd`:/home/user/bluemira -v `pwd`/../process:/home/user/process -v /tmp/.X11-unix:/tmp/.X11-unix -v $HOME/.Xauthority:/user/.Xauthority -e DISPLAY=$DISPLAY --network host bluemira:dev /bin/bash`

je-cook:
only failing tests locally are related to dolfinx and the lack of nmm

je-cook:
Container is now ~ 6.3GB which could probably be shrunk slightly more with some effort but maybe only 10%. 

Ascii error has reared its ugly head again (#504). I havent found a way around it, tried among other things setting locale's etc in the container. I can avoid it using the normal export annoyance which I will put in the bashrc when we are ready to merge this.

This is now only waiting on:

- [ ] dolfinx conversion
- [ ] how we want to use the docker container in the CI

it is unlikely #1102 will be needed now, the tests pass locally and all runtime dependencies have been updated for python 3.10

docker/env.Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+FROM base as release
+COPY --from=build_deps /usr /usr
+COPY --from=build_deps /etc /etc
+RUN pip install git+https://github.com/Fusion-Power-Plant-Framework/bluemira.git@v0.1.0
+
+RUN useradd -ms /bin/bash user
+COPY --from=build_deps --chown=user /opt/venv/ /opt/venv/
+RUN chown user:user /opt/venv
+USER user
+WORKDIR /home/user
```
je-cook, line 87: All of the below lines (apart from the bluemira install) need to be accessible from both develop and release
```suggestion
COPY --from=build_deps /usr /usr
COPY --from=build_deps /etc /etc
RUN pip install git+https://github.com/Fusion-Power-Plant-Framework/bluemira.git@v0.1.0

RUN useradd -ms /bin/bash user
COPY --from=build_deps --chown=user /opt/venv/ /opt/venv/
RUN chown user:user /opt/venv
USER user
WORKDIR /home/user
```

.pre-commit-config.yaml
```diff
@@ -4,12 +4,15 @@ repos:
     hooks:
     - id: trailing-whitespace
       exclude: test_data
+      exclude_types: ["diff"]
```
je-cook, line 7: the changes in this file are not needed anymore as I removed all the patch files.

Maybe its useful to keep this here anyway, up to you

docker/env.Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+
+# QT5 has some not standard lib locations which freecad install doesnt remember
+ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/Qt-5.15.5/lib
+# Dolfin needs help finding Boost runtime libaries
+ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/lib
+
+FROM base as release
+COPY --from=build_deps /usr /usr
+COPY --from=build_deps /etc /etc
+RUN pip install git+https://github.com/Fusion-Power-Plant-Framework/bluemira.git@v0.1.0
```
je-cook, line 81: So we always have the latest release
```suggestion
RUN pip install git+https://github.com/Fusion-Power-Plant-Framework/bluemira.git@main
```

scripts/occ/install-occ-deps.sh
```diff
@@ -0,0 +1,4 @@
+set -e
```
je-cook, line 1: can be removed no longer used

scripts/occ/install-occ.sh
```diff
@@ -0,0 +1,41 @@
+set -e
```
je-cook, line 1: can be remove no longer used

je-cook (changes requested):
Testing the release build with the below build command atm otherwise looks fine for the release. The develop build needs a bit of adjustment I think

```bash
docker build -t bluemira:latest -f docker/env.Dockerfile --target=release ./
```

hsaunders1904:
Frustratingly, I can't test this on my work laptop as I keep running out of disk space 😡. I'll work from home tomorrow and test with my personal laptop

docker/env.Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+COPY scripts/fenics ./scripts/fenics/
+RUN bash scripts/fenics/install-fenics-deps.sh
+RUN bash scripts/fenics/install-fenics.sh
+
+# QT5 has some not standard lib locations which freecad install doesnt remember
+ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/Qt-5.15.5/lib
+# Dolfin needs help finding Boost runtime libaries
+ENV LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/usr/local/lib
+
+RUN useradd -ms /bin/bash user
```
je-cook, line 78: I think you need another layer otherwise it will be a ~16GB image
```suggestion
FROM base as user_base
RUN useradd -ms /bin/bash user
```

docker/env.Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+COPY --from=build_deps /usr /usr
+COPY --from=build_deps /etc /etc
+COPY --from=build_deps --chown=user /opt/venv/ /opt/venv/
+RUN chown user:user /opt/venv
+USER user
+WORKDIR /home/user
+
+FROM base as release
```
je-cook, line 86:
```suggestion
FROM user_base as release
```

docker/env.Dockerfile
```diff
@@ -0,0 +1,93 @@
...
+RUN pip install git+https://github.com/Fusion-Power-Plant-Framework/bluemira.git@main
+
+FROM base as develop
```
je-cook, line 89:
```suggestion
FROM user_base as develop
```

je-cook (changes requested):

requirements.txt
```diff
@@ -3,7 +3,7 @@ asteval==0.9.27
 Babel==2.11.0
 click==8.1.3
 contourpy==1.0.6
-CoolProp==6.4.1
+CoolProp==6.4.2.dev0
```
je-cook, line 6: Its out now so we can just use it directly
```suggestion
CoolProp==6.4.2
```

scripts/h5py/install-h5py.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+
+set -e
+
+apt-get update
+apt-get install -y libhdf5-dev
+HDF5_DIR=HDF5_DIR=/usr/lib/x86_64-linux-gnu/hdf5/serial/ python -m pip install h5py
```
je-cook, line 7:
```suggestion
HDF5_DIR=/usr/lib/x86_64-linux-gnu/hdf5/serial/ python -m pip install h5py
```

scripts/README.md
```diff
@@ -0,0 +1,49 @@
+# Install without conda
```
je-cook, line 1: This one isnt up to date we've changed a lot of things since. Either just remove it and we make a new one in a follow up or update it. Up to you

je-cook (approved):
Looks good, well done :slightly_smiling_face:

scripts/h5py/install-h5py.sh
hsaunders1904: I'm hoping this is why it's not quite working at the moment 🤞

Currently rebuilding...

PR #708, opened 2022-02-11
✨ Coils Refactor

### LINKED ISSUE #210: SymmetricCircuit and force_symmetry assume symmetry about z = 0
In GitLab by @alexanderianblair on Jun 28, 2021, 10:30

Currently users are restricted to systems symmetric about z = 0 when using `SymmetricCircuit` objects and/or the `force_symmetry` option in `Equilibria`, as discussed [here](https://gitlab.com/CoronelBuendia/BLUEPRINT/-/merge_requests/393#note_611696545).

This is the usual use case, but it is possible users may want to create a system symmetric off-axis (about z != 0). Symmetry could be made an attribute of the `Grid` object itself, and all uses of it can construct symmetrised objects as needed. Alternatively, the solve on a symmetric `Grid` object could be entirely performed in the half-plane when symmetry is applied, without the need to expand to full-length arrays after every solve.

### LINKED ISSUE #322: CoilGroup methods failing with SymmetricCircuit members
## Description

ForceField methods have also been broken by the changes to SymmetricCircuit, due to the assumption that the number of independently controlled coil currents (coilset._ccoils) is equal in size to the total number of coils (coilset.n_coils), which is not true when the coilset contains SymmetricCircuit members.

## Proposal

Rationalise how the number of "controlled"  coils and the real number of coils are treated in `CoilSet`.

`coilset.n_coils` should be the true number of physical coils in a `CoilSet`
`coilset.n_control_coils` should be the number of individually controlled current circuits in a `CoilSet`.

might be as simple as:

```python
@property    
def n_control_coils(self):
     return len(self._ccoils)
```

We need also to address how ForceField calculates force and field and jacobians in the case of symmetric circuits.

### LINKED ISSUE #337: Improve default `Coil` meshing approach and make it modifiable

In a few places in `equilibria` the `CoilSet` is meshed with a hard-coded default value that cannot be overriden. This behaviour should be changed such that, as a minimum, the meshing size can be overriden.

## Proposed solution

A code review of how "meshing" is handled in equilibria, and expose the functionality such that it can be overriden by the user at appropriate points.

The meshing in `PicardAbsIterator` is only use in position optimisation, and the meshing should be moved into the `PositionOptimiser` such that the iterator does not directly change the state of the `CoilSet`.

## Alternative solutions

Forever live with `d_coil=0.4`. THis is not desirable

## Additional context

Spin off from #314 @alexanderianblair

### LINKED ISSUE #346: Refactor `equilibria/coils.py`

Implement the general refactor of `coils.py` as discussed on MIRO:

image

The general ideas are as follows:
* `CoilGroup` shall be the new base class object, which `Coil`, `Circuit`, `SymmetricCircuit` and `CoilSet` inherit from
* A new mixin class shall be used to isolate all physical calculations associated with the existence of a `CoilGroup` (psi, Bx, Bz, etc.)
* `CoilGroup` shall remain ignorant of how they are being controlled or size-fixed
* `CoilSet` will be responsible for the control and size-fixing of coils, and have some utility methods for extracting useful subsets of coils inside the underlying `CoilGroup`

Now `SymmetricCircuit` inherits from `CoilGroup` objects, `CoilGroup` objects now may contain a mixture of `Coil` and `CoilGroup` objects within them, which each have their own (distinct) set of methods and attributes. This has introduced bugs such as #322 arising from calling methods and attributes existing in `Coil` objects that do not exist in `CoilGroup `objects. 

## Proposed solution

If `Coil` is made to inherit from `CoilGroup` as a special case (of size 1), then it can be assumed that all `CoilGroup` objects contain only `CoilGroup` objects of smaller size.

## Alternative solutions
`Coil` properties could be added separately to `SymmetricCircuit`, or all existing implementations updated to allow for either `Coil` or `CoilGroup` objects being present in `CoilGroup` objects. This would make the code more fragile when changes are made to how the coilset is handled.

### LINKED ISSUE #679: Expand set_positions Coilset function
Before requesting a new feature, please check the Issues to see if your idea is already being discussed.

## Requirement to address

The Coilset function `set_positions` only set positions for control coils. 

Is it possible to make the control coil focus as an option? 

## Proposed solution

Modify the `set_positions` function adding a optional argument `control_coil_only=True`. 
If this option is set to `False` the position setting would be done on all coils (not only the control ones)

@alexanderianblair, @CoronelBuendia and @je-cook

je-cook:
## Linked Issues
This closes or will close when finished, possibly others to be included

Closes #346 
Closes #322
Closes #679 
Closes #337

## Description

This is a long lived feature branch to refactor coils per #346
@alexanderianblair @CoronelBuendia 

Update 2/12/22

This is now in an MVP state. Still todo:

- [x] Remake the plotting routines
- [x] Update coils examples and examples generally
- [x] Add more tests to get at edge cases
- [x] My sanity check (I've gone a bit code blind now...so best I've got atm)
- [x] Documentation

I'm opening this up for comment now again @CoronelBuendia @hsaunders1904, sorry it is so large

## Interface Changes

Probably some, TODO

## Checklist


bluemira/equilibria/coils.py
```diff
@@ -85,27 +92,318 @@ def generate_name(coil: Any, idx: Optional[int] = None):
...
+        atol = X_TOLERANCE
+        x_min, x_max = (
+            self._quad_x - self._quad_dx - atol,
+            self._quad_x + self._quad_dx + atol,
+        )
+        z_min, z_max = (
+            self._quad_z - self._quad_dz - atol,
+            self._quad_z + self._quad_dz + atol,
+        )
+        return (x >= x_min) & (x <= x_max) & (z >= z_min) & (z <= z_max)
```
CoronelBuendia, line 326: Is this still just checking whether the point is inside the full coil? Looks like it might be checking on all the quad points

bluemira/equilibria/coils.py
```diff
@@ -85,27 +92,318 @@ def generate_name(coil: Any, idx: Optional[int] = None):
...
+        current using semi-analytic method.
+        """
+        return semianalytic_Bx(self.x, self.z, x, z, d_xc=self.dx, d_zc=self.dz)
+
+    def _control_Bz_analytical(self, x, z):
+        """
+        Calculate vertical magnetic field Bz response at (x, z) due to a unit
+        current using semi-analytic method.
+        """
+        return semianalytic_Bz(self.x, self.z, x, z, d_xc=self.dx, d_zc=self.dz)
```
CoronelBuendia, line 363: x, and z don't exist anymore

CoronelBuendia (commented):
I appreciate this is still WIP. Just a couple of quick ones looking at the diff.  I don't know how kosher it is to make a mixin rely on attributes from a subclass.

Would it be possible to get this in a state where it passes the tests before #530?

je-cook:
> I appreciate this is still WIP. Just a couple of quick ones looking at the diff. I don't know how kosher it is to make a mixin rely on attributes from a subclass.
> 
> Would it be possible to get this in a state where it passes the tests before #530?

I'm on the fence about that but because it is a mixin it is _never_ (or should never) be used in isolation. Therefore anything that inherits needs to specifiy the extra arguments. The other way will lead to having two copies of a attribute within the same class I would have thought

bluemira/equilibria/coils.py
je-cook: It is point inside a given quadrature. For most B field calculations I thought that would be what you'd want because if _quad_dz or _quad_dx = 0 then we can just use the greens function?

bluemira/equilibria/coils.py
je-cook: this was taken over from the old coils because it made no distinction to subcoils so I assumed it was the same calculation everywhere.
Also see below comment about existence

je-cook:
Sonarcloud errors are because lots in unimplemented.

I expect the same for bluemira pytest stuff

CoronelBuendia (changes requested):
OK so I've had a quick play, and I'm trying just to focus on the API and usability stuff.

Had a few problems trying to initialise a `CoilGroup`, as I have to make lists for everything including default values.

Seems to be an issue with one of the boolean checks so I didn't get much further

For Coil it seems weird that one can initialise it with just x and z, and I'm not sure if the coil-sizing stuff is robust enough right now. I'm not sure we really want a flag_sizefix on a coil, but until we get to CoilSet I'm not sure how best to treat that..!

I think a few next steps are required to continue ironing out this new case:
* make initialisation of `CoilGroup` a bit more user friendly. (In my head, I thought one could make a CoilGroup from a bunch of CoilGroups). Vectors are fine but the defaults need to auto-map to the length
* implement a simple quadrature approach and test it (want to make sure this structure can neatly handle quadd-ed coilgroups)
* implement circuit and symmetriccircuit (same reasons as above)

Definitely a good starting basis. Big job to get this all right!

bluemira/equilibria/coils.py
```diff
@@ -223,1992 +546,674 @@ def get_max_current(self, coil):
...
-            raise TypeError("Coil positions type should be either list or dict")
-
-    def _set_positions_list(self, positions):
-        for coil, position in zip(self._ccoils, positions):
-            coil.set_position(*position)
+        ret_str = ""
+        for ind, name in enumerate(self.name):
+            np.array
+            ret_str += f"{name}: X={self.x[ind][0]:.2f} m, Z={self.z[ind][0]:.2f} m, I={self.current[ind][0]/1e6:.2f} MA"
+        return ret_str[:-1]
```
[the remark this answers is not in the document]
je-cook: That looks like something half finished...
will fix

bluemira/equilibria/coils.py
```diff
@@ -223,1992 +546,674 @@ def get_max_current(self, coil):
...
-        for coil in self.coils.values():
-            coil.plot(ax=ax, subcoil=subcoil, **kwargs)
+        xx, zz = np.ones((1, 4)) * x_c, np.ones((1, 4)) * z_c
+        x_boundary = xx + dx * np.atleast_2d([-1, 1, 1, -1])
+        z_boundary = zz + dz * np.atleast_2d([-1, -1, 1, 1])
+        return x_boundary, z_boundary
 
 
-class SymmetricCircuit(Circuit):
+class CoilGroup(CoilFieldsMixin, abc.ABC):
```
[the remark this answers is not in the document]
je-cook: My initial thoughts were that CoilGroup wasnt really used directly but Coil and any circuit were the main entry points, maybe in Coilset you could do what you're asking. Either way the interface does need a bit of cleaning

bluemira/equilibria/coils.py
```diff
@@ -223,1992 +546,674 @@ def get_max_current(self, coil):
...
+
+        self._name_map = {
+            f"{self._ctype[en].name}_{ind}" if n is None else n: ind
+            for en, (n, ind) in enumerate(zip(_inputs["name"], self._index))
+        }
 
-        super().__init__([coil, mirror])
+        self._flag_sizefix = False
+        self.__sizer = CoilSizer(self)
+        self.__sizer(self)
```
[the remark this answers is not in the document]
je-cook: ahh that shouldnt be possible. It is because `is_num(np.nan) == True`

bluemira/equilibria/coils.py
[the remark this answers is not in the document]
je-cook: Ok this now works, need to add a test

bluemira/equilibria/coils.py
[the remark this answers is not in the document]
je-cook: Also I now error in the above situation instead of setting a default

bluemira/equilibria/coils.py
```diff
@@ -85,634 +108,631 @@ def generate_name(coil: Any, idx: Optional[int] = None):
...
+            ),
+            self._quad_weighting[None],
+        )
+
+    def Bx(self, x, z):
+        """
+        Calculate radial magnetic field Bx at (x, z)
         """
-        Sets the position of the Coil object
+        return self.control_Bx(x, z) * self.current
```
CoronelBuendia, line 321: I'm not sure here, but it looks like the Green's functions are weighted with a weight of 1 (?). So just from looking at this, it seems as though a discretised coil would have its full current multiplied at all the quadrature points? 

Update: It seems like this is not the case, though, so clearly I'm not following it right!

This einsum stuff I don't understand well, so I'm not sure if it's a dot product or a hadamard product... Is the padding to facilitate an einsum somehow?

bluemira/equilibria/coils.py
```diff
@@ -723,1492 +743,1226 @@ def F(self, eqcoil):  # noqa :N802
...
 
-    def Bz_greens(self, bgreen):
-        """
-        Uses the Greens mapped dict to quickly compute the Bx
-        """
-        bz_coils = 0
-        for name, coil in self.coils.items():
-            bz_coils += coil.Bz_greens(bgreen[name])
-        return bz_coils
+    A group of coils is defined as shaing a property eg current
```
CoronelBuendia, line 968: sharing

bluemira/equilibria/coils.py
```diff
@@ -723,1492 +743,1226 @@ def F(self, eqcoil):  # noqa :N802
...
-        super().remove_coil(coilname)
-        self._classify_control()
+        if isinstance(new_current, Iterable):
+            new_current = new_current[0]
+        self._current[:] = new_current
+
+
+class SymmetricCircuit(Circuit):
+    """
+    Positionally symmetric coils everything else the same
```
CoronelBuendia, line 1579: We should be a bit more specific here: symmetric in z about the x=0 plane

bluemira/equilibria/coils.py
```diff
@@ -723,1492 +743,1226 @@ def F(self, eqcoil):  # noqa :N802
...
+    b_max: Optional[float]
+        Maximum magnetic field at the coil [T]
+
+    """
+
+    __slots__ = ("_uv", "_symmetry_point", "_point")
+
+    def __init__(
+        self,
+        symmetry_line: np.ndarray[[float, float], [float, float]],
```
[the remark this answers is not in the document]
CoronelBuendia: Funny.. it's come round now!

bluemira/equilibria/coils.py
```diff
@@ -723,1492 +743,1226 @@ def F(self, eqcoil):  # noqa :N802
...
         """
-        for coil, d_i in zip(self._ccoils, current_change):
-            coil.adjust_current(d_i)
+        raise EquilibriaError("class not designed to be subclassed")
 
-    def set_control_currents(self, currents, update_size=True):
+    def __str__(self) -> str:
         """
-        Sets the currents in the coils being controlled
+        Pretty pront Coilset
```
CoronelBuendia, line 1758: print

bluemira/equilibria/coils.py
```diff
@@ -22,42 +22,74 @@
...
-CS_COIL_NAME = "CS_{}"
-NO_COIL_NAME = "Unclassified_{}"
 
+__ITERABLE_FLOAT = Union[float, Iterable[float]]
+__ITERABLE_COILTYPE = Union[str, CoilType, Iterable[Union[str, CoilType]]]
+__ANY_ITERABLE = Union[__ITERABLE_COILTYPE, __ITERABLE_FLOAT]
 
-class CoilNamer:
+
+class CoilNumber:
```
CoronelBuendia, line 92: This is quite minor, but I've found that if you initialise a Coil incorrectly, it still increments the counters. This was probably the case before too, so we can forget it for now.

bluemira/equilibria/coils.py
```diff
@@ -85,634 +108,631 @@ def generate_name(coil: Any, idx: Optional[int] = None):
...
-            self.flag_sizefix = False
+    def __init__(self, weighting=None):
+        if type(self) == CoilFieldsMixin:
+            raise TypeError("Can't be initialised directly")
 
-        coil.flag_sizefix = self.flag_sizefix
+        self.discretise(weighting)
 
-    def __call__(self, coil, current=None):
+    def _pad_discretisation(
```
CoronelBuendia, line 166: I don't understand why we need to pad at all... why is this?

bluemira/equilibria/coils.py
```diff
@@ -85,634 +108,631 @@ def generate_name(coil: Any, idx: Optional[int] = None):
...
         x, z = np.ascontiguousarray(x), np.ascontiguousarray(z)
-        # Add an offset, to ensure points very near the edge are counted as
-        # being on the edge of a coil
-        atol = X_TOLERANCE
-        x_min, x_max = self.x - self.dx - atol, self.x + self.dx + atol
-        z_min, z_max = self.z - self.dz - atol, self.z + self.dz + atol
-        return (x >= x_min) & (x <= x_max) & (z >= z_min) & (z <= z_max)
 
-    def assign_material(self, j_max=NBTI_J_MAX, b_max=NBTI_B_MAX):
+        zero_coil_size = np.logical_or(self._dx == 0, self._dz == 0)
```
CoronelBuendia, line 413: these have now changed to `np.isclose` and probably should here too

bluemira/equilibria/coils.py
```diff
@@ -723,1492 +743,1226 @@ def F(self, eqcoil):  # noqa :N802
...
+    def dz(self) -> np.ndarray:
         """
-        Set the size of the coils in the Circuit.
+        Get dz coordinate of each coil
         """
-        for coil in self.coils.values():
-            coil.make_size(current=current)
+        _dz = self._dz[:]
+        _dz.flags.writeable = False
+        return _dz
```
CoronelBuendia, line 1350: Woah.. what's this a read-only view? Is that not safer as a copy? I'm not sure what the reasons for this are, but I think a comment would be good!

CoronelBuendia (changes requested):
OK, I've had a look over things a little. I think it's getting there, but unfortunately it's kind of hard to tell without a fully functional PR.

I don't follow the einsum maths (as usual! :D) but it seems to be working as I'd expect, along with the discretisation and control mixing logic.

I reckon it is best just to plough on through until such a point that `CoilSet` control logic and the like is required so that at least an Equilibrium can be generated or one of the examples run. 

I don't mind the API change for `CoilSet` from `[coils]` to `*coils`, although we did promise we would avoid API changes... Let's see where we get to once it is fully functional and decide whether we have another round of API changes. If it is just this, then I'd tend towards keeping things as is, but I guess it's likely we accumulate more stuff. `SymmetricCircuit` instantiation, for example, is likely to be different.

All things said, this will be a lot easier to understand once all the tests pass. I think if you have any doubts on the way we can talk about them live, and rope Alex in too.

P.S.: I vaguely remember us saying that `flag_sizefix` should be handled inside a `CoilSet` and not `Coil`, but I guess that will be on the MIRO somewhere.

bluemira/equilibria/coils.py
je-cook: Einsum is all hadamard products with possible summations over specific axes. A dot product is just a hadamard product of two matrices where you sum over the 'middle' axes eg the last axis of the first array and the first axis of the last.

eg.

```python
In [14]: a = np.ones((4,3)) * 4 

In [15]: b = np.ones((3,4)) * 6

In [16]: np.dot(a, b)
Out[16]: 
array([[72., 72., 72., 72.],
       [72., 72., 72., 72.],
       [72., 72., 72., 72.],
       [72., 72., 72., 72.]])

In [17]: np.einsum("ij, jk -> ijk", a, b)
Out[17]: 
array([[[24., 24., 24., 24.],
        [24., 24., 24., 24.],
        [24., 24., 24., 24.]],

       [[24., 24., 24., 24.],
        [24., 24., 24., 24.],
        [24., 24., 24., 24.]],

       [[24., 24., 24., 24.],
        [24., 24., 24., 24.],
        [24., 24., 24., 24.]],

       [[24., 24., 24., 24.],
        [24., 24., 24., 24.],
        [24., 24., 24., 24.]]])

In [18]: np.einsum("ij, jk -> ik", a, b)
Out[18]: 
array([[72., 72., 72., 72.],
       [72., 72., 72., 72.],
       [72., 72., 72., 72.],
       [72., 72., 72., 72.]])
```

I'll comment on the padding in your other comment

bluemira/equilibria/coils.py
je-cook: I think its important to keep this. That way dont have to do a lot of catching in order to have unique coil numbers even if we don't necessarily have unique coil names.

I havent yet done it but maybe in the next PR the region and path links should be done with the coil numbers (also anything else that internally uses the names for access)

bluemira/equilibria/coils.py
je-cook: The padding is so we can use the qudrature properties in one array even though they may be of different lengthed vectors due to coils eg having the same discretisation but are different sizes. No extra calculations are done on these zeros as of [17af92d](https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/708/commits/17af92dbf3be1c959aff6ae6e41b41adfbb9c9a6). The user also is unlikely to interact with them.

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,1090 @@
...
+    ----------
+    coils: Union[Coil, CoilGroup[Coil]]
+
+    """
+
+    def __init__(self, *coils: Union[Coil, CoilGroup[Coil]]):
+        if isinstance(coils[0], Iterable):
+            raise TypeError("Iterable is not a Coil or CoilGroup")
+        self._coils = coils
+        self._pad_discretisation(self.__list_getter("_quad_x"))
```
je-cook, line 107:
```suggestion
```
now padded on access

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,1090 @@
...
+        """
+        Fix coils sizes
+        """
+        for fx in self.__list_getter("fix_size"):
+            fx()
+
+    def add_coil(self, *coils: Union[Coil, CoilGroup[Coil]]):
+        """Add coils to the coil group"""
+        self._coils = (*self._coils, *coils)
+        self._pad_discretisation(self.__list_getter("_quad_x"))
```
je-cook, line 149:
```suggestion
```

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,1090 @@
...
+        coils = np.array(self._coils, dtype=object)
+        mask = np.full(coils.size, True, dtype=bool)
+        mask[to_remove] = False
+
+        for no, c in enumerate(self._coils):
+            if c.name == []:
+                mask[no] = False
+
+        self._coils = tuple(coils[mask])
+        self._pad_discretisation(self.__list_getter("_quad_x"))
```
je-cook, line 197:
```suggestion
```

bluemira/equilibria/opt_problems.py
```diff
@@ -143,8 +143,8 @@ def read_coilset_state(coilset, current_scale):
             Number of substates (blocks) in the state vector.
         """
         substates = 3
-        x, z = coilset.get_positions()
-        currents = coilset.get_control_currents() / current_scale
+        x, z = coilset.position
+        currents = coilset.current / current_scale
```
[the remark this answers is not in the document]
je-cook: unsure here because x,z is not done with control coils originally

bluemira/equilibria/opt_problems.py
```diff
@@ -168,14 +168,9 @@ def set_coilset_state(coilset, coilset_state, current_scale):
...
-        # SymmetricCircuits, it appears...
-        # positions = list(zip(x, z))
-        # self.coilset.set_positions(positions)
-        for i, coil in enumerate(coilset.coils.values()):
-            coil.x = x[i]
-            coil.z = z[i]
-        coilset.set_control_currents(currents * current_scale)
+        coilset.x = x
+        coilset.z = z
+        coilset.current = currents * current_scale
```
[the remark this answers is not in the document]
je-cook: likewise here

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,1088 @@
...
+        if not _top_level:
+            return removed_names + to_remove
+
+    @classmethod
+    def from_group_vecs(cls, eqdsk: EQDSKInterface):
+        """
+        Initialises an instance of CoilSet from group vectors. This has been
+        implemented as a dict operation, because it will occur for eqdsks only.
+        Future dict instantiation methods will likely differ, hence the
+        confusing name of this method.
```
hsaunders1904, line 206:
```suggestion
        Initialises an instance of CoilSet from group vectors. 
        
        This has been implemented as a dict operation, because it will 
        occur for eqdsks only.
        Future dict instantiation methods will likely differ, hence the
        confusing name of this method.
```

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,1088 @@
...
+        """
+        pfcoils = []
+        cscoils = []
+        passivecoils = []
+        for i in range(eqdsk.ncoil):
+            dx = eqdsk.dxc[i]
+            dz = eqdsk.dzc[i]
+            if abs(eqdsk.Ic[i]) < I_MIN:
+                # Catch CREATE's crap 0's
```
hsaunders1904, line 215: Any more context for what 'crap 0s' means?

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,1088 @@
...
+                        dx=dx,
+                        dz=dz,
+                        ctype="PF",
+                    )
+                    coil.fix_size()  # Oh ja
+                    pfcoils.append(coil)
+
+        coils = pfcoils
+        if len(cscoils) != 0:
+            coils.extend(cscoils)
```
hsaunders1904, line 253: Is the `if` necessary here? 
```suggestion
        coils.extend(cscoils)
```

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,1088 @@
...
+        for c in self._coils:
+            if isinstance(c, CoilGroup):
+                try:
+                    return c._find_coil(name)
+                except KeyError:
+                    pass
+            elif c.name == name:
+                return c
+        else:
+            raise KeyError("Coil not found in Group")
```
hsaunders1904, line 386:
```suggestion
            raise KeyError(f"Coil '{name}' not found in Group")
```

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,1088 @@
...
+        if isinstance(ctype, str):
+            ctype = CoilType[ctype]
+        for c in self._coils:
+            if isinstance(c, CoilGroup):
+                coils.extend(c._get_coiltype(ctype))
+            elif c.ctype == ctype:
+                coils.append(c)
+        return coils
+
+    def get_coiltype(self, ctype):
```
hsaunders1904, line 400:
```suggestion
    def get_coiltype(self, ctype: Union[str, CoilType]):
```

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,1088 @@
...
+        return np.where(
+            np.isnan(self.j_max),
+            max_current,
+            get_max_current(self.dx, self.dz, self.j_max),
+        )
+
+    @property
+    def name(self) -> np.ndarray:
+        """Get coil names"""
+        return self.__getter("name").tolist()
```
hsaunders1904, line 435: Is this returning a list or array?

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,1088 @@
...
+
+    @property
+    def position(self):
+        """Get coil x, z positions"""
+        return np.array([self.x, self.z])
+
+    @property
+    def ctype(self) -> np.ndarray:
+        """Get coil types"""
+        return self.__getter("ctype").tolist()
```
hsaunders1904, line 455: Ditto

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,1088 @@
...
+
+    @n_turns.setter
+    def n_turns(self, values: Union[float, Iterable[float]]):
+        """Set coil number of turns"""
+        self.__setter("n_turns", values)
+
+
+class Circuit(CoilGroup):
+    """
+    A CoilGroup with the same currents
```
hsaunders1904, line 620: Does this mean, 'where all coils have the same current'?

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,1088 @@
...
+        The coils to be added to the set
+    control_names: Optional[List]
+        List of coil names to be controlled
+
+    """
+
+    def __init__(
+        self,
+        *coils: Union[Coil, CoilGroup[Coil]],
+        control_names: Union[bool, List] = True,
```
hsaunders1904, line 796: What happens if this is `False`? Might it be better to treat `None` (i.e. no names given) as all coils, then an empty list to indicate no control coils (if that's allowed). Then the `Optional[List[str]]` in the docstring will be correct.

hsaunders1904 (changes requested):
I'm about halfway through, I think, so, getting there. Here are my comments so far, I'll have a read through of the next half tomorrow!

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,1088 @@
...
+        control: bool
+            operations on control coils only
+
+        Returns
+        -------
+        np.ndarray of response
+        """
+        return self._sum(super().Bz(x, z), sum_coils=sum_coils, control=control)
+
+    def unit_psi(
```
CoronelBuendia, line 944: Not really a fan of `unit_..` as a name for these methods. The whole point is to get the response at a location for a unit _current_ in a given coil or grouping thereof for [psi, Bx, Bz, Bp]. 

As such, e.g. `unit_psi(x, z)` is effectively doing `get_response_to_a_unit_current_in_coil_for_psi_at_location(x, z)`. I feel that the previous formulation of `control_psi(x, z)` or e.g. `get_psi_response(x, z)` are clearer short-forms for this.

The `unit` aspect is perhaps the least interesting part of this, and is more readily confused with e.g. [V.s], [Wb/rad], etc.

bluemira/equilibria/coils/_coil.py
```diff
@@ -0,0 +1,521 @@
...
+        current: float = 0,
+        j_max: float = np.nan,
+        b_max: float = np.nan,
+        discretisation: float = 1,
+        n_turns: int = 1,
+    ):
+        self._dx = None
+        self._dz = None
+        self._discretisation = 1
+        self._flag_sizefix = None not in (dx, dz)
```
CoronelBuendia, line 151: Perhaps fiddly, but would be great to make this something that lives on `CoilSet` in the same way `control` now does.

Not necessary for this PR

bluemira/equilibria/coils/_field.py
```diff
@@ -0,0 +1,770 @@
...
+    def unit_psi(self, x: Union[float, np.ndarray], z: Union[float, np.ndarray]):
+        """
+        Calculate poloidal flux at (x, z) due to a unit current
+        """
+        x, z = np.ascontiguousarray(x), np.ascontiguousarray(z)
+
+        ind = np.where(self._quad_weighting != 0)
+        out = np.zeros((*x.shape, *self._quad_x.shape))
+
+        out[(*(slice(None) for _ in x.shape), *ind)] = greens_psi.py_func(
```
CoronelBuendia, line 79: Ouch, what's the performance hit?

bluemira/equilibria/coils/_field.py
```diff
@@ -0,0 +1,770 @@
...
+        semianalytic_func: Callable,
+    ):
+        """
+        Boiler-plate helper function to mixed the Green's function responses
+        with the semi-analytic function responses, as a function of position
+        outside/inside the coil boundary.
+
+        Parameters
+        ----------
+        x,z: Union[float, int, np.array]
```
CoronelBuendia, line 184:
```suggestion
        x: Union[float, int, np.array]
            The x values at which to calculate the response at
        z: Union[float, int, np.array]
            The z values at which to calculate the response at
```

bluemira/equilibria/coils/_field.py
```diff
@@ -0,0 +1,770 @@
...
+
+        Used for situation where there are calculation points both inside and
+        outside the coil boundaries.
+
+        Parameters
+        ----------
+        inside: np.ndarray[bool]
+            array of if the point is inside a coil
+        x,z: Union[float, int, np.array]
+            Points to calculate field at
```
CoronelBuendia, line 239:
```suggestion
        x: Union[float, int, np.array]
            The x values at which to calculate the response at
        z: Union[float, int, np.array]
            The z values at which to calculate the response at
```

bluemira/equilibria/coils/_field.py
```diff
@@ -0,0 +1,770 @@
...
+        """
+        Calculate radial magnetic field B respose at (x, z) due to a unit
+        current using Green's functions.
+
+        Parameters
+        ----------
+        greens: Callable
+            greens function
+        x,z: Union[float, int, np.array]
+            Points to calculate field at
```
CoronelBuendia, line 336:
```suggestion
        x: Union[float, int, np.array]
            The x values at which to calculate the response at
        z: Union[float, int, np.array]
            The z values at which to calculate the response at
```

bluemira/equilibria/coils/_field.py
```diff
@@ -0,0 +1,770 @@
...
+    ):
+        """
+        Calculate radial magnetic field Bx respose at (x, z) due to a unit
+        current using Green's functions.
+
+        Parameters
+        ----------
+        greens: Callable
+            greens function
+        x,z: Union[float, int, np.array]
```
CoronelBuendia, line 391:
```suggestion
        x: Union[float, int, np.array]
            The x values at which to calculate the Bx response
        z: Union[float, int, np.array]
            The z values at which to calculate the Bx response
```

bluemira/equilibria/coils/_field.py
```diff
@@ -0,0 +1,770 @@
...
+        _quad_weight: Optional[np.ndarray] = None,
+    ):
+        """
+        Calculate vertical magnetic field Bz at (x, z) due to a unit current
+
+        Parameters
+        ----------
+        greens: Callable
+            greens function
+        x,z: Union[float, int, np.array]
```
CoronelBuendia, line 427:
```suggestion
        x: Union[float, int, np.array]
            The x values at which to calculate the Bz response
        z: Union[float, int, np.array]
            The z values at which to calculate the Bz response
```

bluemira/equilibria/coils/_field.py
```diff
@@ -0,0 +1,770 @@
...
+        """
+        Calculate radial magnetic field Bx response at (x, z) due to a unit
+        current using semi-analytic method.
+
+        Parameters
+        ----------
+        semianalytic: Callable
+            semianalytic function
+        x,z: Union[float, int, np.array]
+            Points to calculate field at
```
CoronelBuendia, line 467:
```suggestion
        x: Union[float, int, np.array]
            The x values at which to calculate the response at
        z: Union[float, int, np.array]
            The z values at which to calculate the response at
```

bluemira/equilibria/coils/_field.py
```diff
@@ -0,0 +1,770 @@
...
+        coil_dz: Optional[np.ndarray] = None,
+    ):
+        """
+        Calculate vertical magnetic field Bx response at (x, z) due to a unit
+        current using semi-analytic method.
+
+        Parameters
+        ----------
+        x,z: Union[float, int, np.array]
+            Points to calculate field at
```
CoronelBuendia, line 518:
```suggestion
        x: Union[float, int, np.array]
            The x values at which to calculate the Bx response
        z: Union[float, int, np.array]
            The z values at which to calculate the Bx response
```

bluemira/equilibria/coils/_field.py
```diff
@@ -0,0 +1,770 @@
...
+        x: np.ndarray,
+        z: np.ndarray,
+        split: bool = False,
+        coil_x: Optional[np.ndarray] = None,
+        coil_z: Optional[np.ndarray] = None,
+        coil_dx: Optional[np.ndarray] = None,
+        coil_dz: Optional[np.ndarray] = None,
+    ):
+        """
+        Calculate radial magnetic field Bx response at (x, z) due to a unit
```
CoronelBuendia, line 459:
```suggestion
        Calculate radial magnetic field response at (x, z) due to a unit
```

bluemira/equilibria/coils/_field.py
```diff
@@ -0,0 +1,770 @@
...
+        coil_dz: Optional[np.ndarray] = None,
+    ):
+        """
+        Calculate vertical magnetic field Bz response at (x, z) due to a unit
+        current using semi-analytic method.
+
+        Parameters
+        ----------
+        x,z: Union[float, int, np.array]
+            Points to calculate field at
```
CoronelBuendia, line 555:
```suggestion
        x: Union[float, int, np.array]
            The x values at which to calculate the Bz response
        z: Union[float, int, np.array]
            The z values at which to calculate the Bz response
```

bluemira/equilibria/coils/_field.py
```diff
@@ -0,0 +1,770 @@
...
+                Bz = coil2.unit_Bz(x, z)
+                Bx = coil2.unit_Bx(x, z)
+            # 1 cross B
+            response[:, j, :] = (
+                2 * np.pi * x[:, None] * np.squeeze(np.array([Bz, -Bx]).T)
+            )
+        return response
+
+
+class CoilFieldsMixin(CoilGroupFieldsMixin):
```
CoronelBuendia, line 652: Lot of duplication here, is this needed? Only get rid of it if the increased coupling is not too bad..

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,1088 @@
...
+    Notes
+    -----
+    Although two groups can be defined any movement of the coils is
+    achieved by offsets to the mean position of the coilgroups
+    """
+
+    def __init__(
+        self,
+        *coils: Union[Coil, CoilGroup[Coil]],
+        symmetry_line: Union[Tuple, np.ndarray] = ((0, 0), (1, 0)),
```
CoronelBuendia, line 680: At present, symmetry will only work about the z=0 line. I would say don't expose this as an option until we bother to address https://github.com/Fusion-Power-Plant-Framework/bluemira/issues/210 (low priority, seeing as it's not that difficult to shift the grid).

bluemira/equilibria/positioner.py
```diff
@@ -174,15 +167,31 @@ def equispace_PF(self, track, n_PF):
         pos = np.linspace(0, 1, n_PF)
         xint, zint = interp1d(l_norm, tf_loop.x)(pos), interp1d(l_norm, tf_loop.z)(pos)
         return [
-            Coil(xint[i], zint[i], name=PF_COIL_NAME.format(n_PF - i), j_max=NBTI_J_MAX)
-            for i in range(n_PF)
+            Coil(xint[i], zint[i], ctype="PF", j_max=NBTI_J_MAX) for i in range(n_PF)
         ]
 
-    def equispace_CS(self, x_cs, tk_cs, z_min, z_max, n_CS):
+    def equispace_CS(self, x_cs, tk_cs, z_min, z_max, n_CS, j_max=12.5):
```
CoronelBuendia, line 173:
```suggestion
    def equispace_CS(self, x_cs, tk_cs, z_min, z_max, n_CS, j_max=NB3SN_J_MAX):
```

bluemira/equilibria/positioner.py
```diff
@@ -33,14 +33,7 @@
...
-from bluemira.equilibria.coils import (
-    CS_COIL_NAME,
-    PF_COIL_NAME,
-    Coil,
-    CoilSet,
-    Solenoid,
-    get_max_current,
-)
+from bluemira.equilibria.coils import Coil, CoilSet, get_max_current
 from bluemira.equilibria.constants import NBTI_J_MAX
```
CoronelBuendia, line 37:
```suggestion
from bluemira.equilibria.constants import NB3SN_J_MAX, NBTI_J_MAX
```

bluemira/equilibria/coils/_field.py
```diff
@@ -0,0 +1,770 @@
...
+        atol: Optional[float]
+            Add an offset, to ensure points very near the edge are counted as
+            being on the edge of a coil
+
+        Returns
+        -------
+        inside: np.array(dtype=bool)
+            The Boolean array of point indices inside/outside the coil boundary
+        """
+        x, z = np.ascontiguousarray(x)[..., None], np.ascontiguousarray(z)[..., None]
```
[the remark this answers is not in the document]
je-cook: @CoronelBuendia added this on the old version and I can't remember why... possibly something speed related

CoronelBuendia (changes requested):
OK just a code review for now, will get started on whacking this around this afternoon hopefully.

bluemira/equilibria/coils/_field.py
```diff
@@ -0,0 +1,770 @@
...
+                # self inductance
+                xxw = xw[same_pos]
+                cr = self._current_radius[xxw]
+                Bz = np.zeros((x.size, 1))
+                Bx = Bz.copy()  # Should be 0 anyway
+                mask = np.zeros_like(Bz, dtype=bool)
+                mask[same_pos] = True
+                if any(cr != 0):
+                    cr_ind = np.where(cr != 0)
+                    Bz[mask][cr_ind] = (
```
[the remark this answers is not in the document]
je-cook: its just because its used in multiple places in the calculation of Bz

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,1088 @@
...
+    get_max_current,
+)
+from bluemira.equilibria.constants import I_MIN
+from bluemira.equilibria.error import EquilibriaError
+from bluemira.equilibria.file import EQDSKInterface
+from bluemira.equilibria.plotting import CoilGroupPlotter
+from bluemira.utilities.tools import flatten_iterable, yintercept
+
+
+def symmetrise_coilset(coilset):
```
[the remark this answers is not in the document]
je-cook: I tried that but we get a load of dependency loops if I do that because of the `isinstance` checks

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,1088 @@
...
+        """
+        if ctype is None:
+            return len(self.x)
+
+        if not isinstance(ctype, CoilType):
+            ctype = CoilType[ctype]
+
+        return Counter(self.ctype)[ctype]
+
+    def plot(self, ax=None, subcoil=True, label=False, force=None, **kwarg):
```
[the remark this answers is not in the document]
je-cook: I can add a comment. They all get fed directly to mpl

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,1088 @@
...
+    def fix_sizes(self):
+        """
+        Fix coils sizes
+        """
+        for fx in self.__list_getter("fix_size"):
+            fx()
+
+    def add_coil(self, *coils: Union[Coil, CoilGroup[Coil]]):
+        """Add coils to the coil group"""
+        self._coils = (*self._coils, *coils)
```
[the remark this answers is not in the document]
je-cook: I wanted to make it difficult to modify the internal object. If its a tuple most will fail at the first hurdle and use the builtin method. It also means we don't have to do a type conversion when we set the class up as `*coils` makes it already a tuple. I do see your point however. What do you think?

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,1088 @@
...
+                c_names.append(c.name)
+
+        to_remove = [names.index(c_n) for c_n in c_names]
+
+        coils = np.array(self._coils, dtype=object)
+        mask = np.full(coils.size, True, dtype=bool)
+        mask[to_remove] = False
+
+        for no, c in enumerate(self._coils):
+            if c.name == []:
```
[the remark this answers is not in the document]
je-cook: Its a list if c is a coils group because coilgroups can be nested. The idea was to remove empty groups. I will add a comment

bluemira/equilibria/coils/_grouping.py
je-cook: @CoronelBuendia may be able to say more here :laughing:. Its copypasted from the old coils

bluemira/equilibria/coils/_grouping.py
je-cook: should be a list

bluemira/equilibria/coils/_grouping.py
je-cook: yep, list

bluemira/equilibria/coils/_coil.py
```diff
@@ -0,0 +1,521 @@
...
+        self.j_max = j_max
+        self.b_max = b_max
+        self.ctype = ctype
+        self.name = name
+        self.n_turns = n_turns
+
+        self._number = CoilNumber.generate(self.ctype)
+        if self.name is None:
+            self.name = f"{self._ctype.name}_{self._number}"
+
```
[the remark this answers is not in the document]
je-cook: The number is a unique identifier its should be used for other things then the name doesnt matter if for whatever reason is duplicated

bluemira/equilibria/coils/_coil.py
```diff
@@ -0,0 +1,521 @@
...
+        """
+        Plot a Coil
+        """
+        return CoilGroupPlotter(
+            self, ax=ax, subcoil=subcoil, label=label, force=force, **kwarg
+        )
+
+    def n_coils(self):
+        """Number of coils in coil"""
+        return 1
```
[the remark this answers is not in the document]
je-cook: Because coils and coilgroups can be nested next to each other I thought it was important to be able to access the number of coils in both instances. Will add a note

bluemira/equilibria/coils/_coil.py
```diff
@@ -0,0 +1,521 @@
...
+        self._dz = None if value is None else float(value)
+        if None not in (self.dx, self.dz):
+            self._discretise()
+            self._set_coil_attributes()
+
+    @current.setter
+    def current(self, value: float):
+        """Set coil current"""
+        self._current = float(value)
+        if None not in (self.dx, self.dz) and not self._flag_sizefix:
```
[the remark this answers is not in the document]
je-cook: ine of them could be zeros in initialisation of the class so I'm making sure that they are both set before trying to resize the coil.

bluemira/equilibria/coils/_coil.py
```diff
@@ -0,0 +1,521 @@
...
+            Overwrite default constant material max field [T]
+
+        """
+        self.j_max = j_max
+        self.b_max = b_max
+
+    def get_max_current(self):
+        """Get max current"""
+        return (
+            np.infty
```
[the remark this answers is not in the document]
je-cook: Likewise, I only knew about inf and infty

bluemira/equilibria/coils/_field.py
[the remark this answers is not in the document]
je-cook: yep debugging miss

bluemira/equilibria/coils/_grouping.py
je-cook: I had that before but thought it could be confusing. If you've jumped to that too I will put it back

bluemira/equilibria/coils/_field.py
je-cook: Debugging miss :)

bluemira/equilibria/coils/_grouping.py
je-cook: control then is confused with the concept of control coils which is different. I could change the to `<>_response` possibly. Get makes me think that there should be a set

bluemira/equilibria/coils/_grouping.py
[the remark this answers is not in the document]
hsaunders1904: It is a private attribute, so users are responsible if they make edits to it. I do get that you're also trying to prevent developers from accidentally editing it, too. It's not too big of a deal as long as `self._coils` is small, if it gets large, then a `list` will be much more efficient.

bluemira/equilibria/coils/_grouping.py
hsaunders1904: Seems a reasonable API I think, but we should make sure it's spelled out in the docstring

bluemira/equilibria/coils/_field.py
je-cook: This was the minimal I could do while storing the coordinates of a single coils as floats otherwise we will sometimes have arrays and sometimes floats and its hard to track. The first two functions are unfortunate dupes but the third is just a wrapper.

bluemira/equilibria/coils/_grouping.py
je-cook: Would it not be possible to do eg x=5 as the symmetry line (or anything perpendicular to z, symmetric about something similar to the plasma centre)?

bluemira/equilibria/coils/_coil.py
```diff
@@ -0,0 +1,521 @@
...
+        x: float,
+        z: float,
+        dx: float = None,
+        dz: float = None,
+        name: Optional[str] = None,
+        ctype: Union[str, CoilType] = CoilType.NONE,
+        current: float = 0,
+        j_max: float = np.nan,
+        b_max: float = np.nan,
+        discretisation: float = 1,
```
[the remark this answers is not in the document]
je-cook: 1 = no discretisation 0= infinite discretisation

bluemira/equilibria/coils/_coil.py
```diff
@@ -0,0 +1,521 @@
...
+    ctype: Union[str, CoilType]
+        Type of coil as defined in CoilType
+    current: float (default = 0)
+        Coil current [A]
+    j_max: float
+        Maximum current density in the coil [A/m^2]
+    b_max: float
+        Maximum magnetic field at the coil [T]
+    discretisation: float
+        discretise the coil (between 0 and 1)
```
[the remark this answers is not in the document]
je-cook: I will expand the comment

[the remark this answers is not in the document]
je-cook: expanded the description. Its the fraction in width and height eg 0.5 will result in 4 filaments

bluemira/equilibria/coils/_coil.py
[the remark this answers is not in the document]
hsaunders1904: Ah, I see. Would a `has_size` method make that intent clearer? I thought I saw this logic repeated in other places, so could be worth refactoring into a function anyway

bluemira/equilibria/coils/_grouping.py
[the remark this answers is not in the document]
je-cook: I imagine it will always be a relatively small number of coils and also an uncommon operation (adding or removing coils). Something to keep an eye on I think but I don't think worth changing for now.

Alternatively I could do:

```python
coils_list = list(self._coils)
coils_list.extend(coils)
self._coils = tuple(coils_list)
```

bluemira/equilibria/coils/_coil.py
je-cook: I think its possible but will require a bit of work as the call to resize would need to be overridden on each coil at the moment.

bluemira/equilibria/coils/_coil.py
```diff
@@ -0,0 +1,521 @@
...
+    Parameters
+    ----------
+    x: float
+        Coil geometric centre x coordinate [m]
+    z: float
+        Coil geometric centre z coordinate [m]
+    dx: float
+        Coil radial half-width [m] from coil centre to edge (either side)
+    dz: float
+        Coil vertical half-width [m] from coil centre to edge (either side)
```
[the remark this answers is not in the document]
je-cook: At the moment its implicit from specifying dx and dz maybe as @CoronelBuendia  mentioned below this could be dealt with in a new PR

bluemira/equilibria/coils/_grouping.py
CoronelBuendia: Ach some EQDSK formats pad a vector ~Xe-12 - Xe-22 currents for like 100 coils or so. So if you read them you get like 90 coils with quasi-zero current popping up

bluemira/equilibria/coils/_grouping.py
je-cook: Removed

je-cook:
Ok I'm done once through with those comments

bluemira/equilibria/coils/_coil.py
```diff
@@ -0,0 +1,521 @@
...
+
+    __PF_counter: int = 1
+    __CS_counter: int = 1
+    __no_counter: int = 1
+
+    @staticmethod
+    def generate(ctype: CoilType) -> int:
+        """
+        Generate a coil name based on its type and indexing if specified. If no index is
+        specified, an encapsulated global counter assigns an index.
```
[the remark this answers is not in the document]
hsaunders1904: Should the '_If no index is specified_' part of the docstring be removed then?

bluemira/equilibria/coils/_coil.py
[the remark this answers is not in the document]
hsaunders1904: I think we should make this explicit in the docstring. It's not necessarily what a user might expect. Especially as we're not allowing resizing, and not currently throwing any warnings or error when a resize is attempted and not allowed. I can imagine this being very frustrating from a usability point-of-view.

bluemira/equilibria/coils/_coil.py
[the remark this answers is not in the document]
CoronelBuendia: But it is only used for naming?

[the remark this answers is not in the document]
je-cook: We use the names in eg the region mapping and I would like to not use name as a unique identifier there as it will possibly restrict the user if they have dupe names with the region mapper. In a follow up PR I want to change that to use the unique identifier to link regions with specific coils.

bluemira/equilibria/coils/_coil.py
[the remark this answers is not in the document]
je-cook: oh yes definitely

bluemira/equilibria/coils/_coil.py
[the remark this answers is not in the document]
je-cook: Done

hsaunders1904 (changes requested):
OK, that's everything from me, I think

bluemira/equilibria/coils/_grouping.py
je-cook: A `'NONE'` type coil is meant to be used for passive coils that can never be controlled eg a iron insert or similar. In the code it just means they are separated from other types of coil and are less likely to accidentally be used for things. There are no actual protections in the code to modify their functionality so functionally they work the same as the other two types.

je-cook: It probably should also be noted that `'NONE'` coils are the default.

bluemira/equilibria/coils/_coil.py
```diff
@@ -0,0 +1,521 @@
...
+        Fixes the size of all coils
+        """
+        self._flag_sizefix = True
+
+    def resize(self, current: Optional[float] = None):
+        """Resize coil given a current"""
+        if not self._flag_sizefix:
+            # Adjust the size of the coil
+            self.dx, self.dz = self._make_size(current)
+            self._set_coil_attributes()
```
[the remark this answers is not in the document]
je-cook: Fair point. I will add a message to the initialisation if dx and dz are set that way at least there is one info message. The init dx and dz just stops the recalculation of the coil size if the current changes.

It almost feels like a debug message to me though because if you set the coil size its a definitive action that has been taken.

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,897 @@
...
+        return self._control
+
+    @control.setter
+    def control(self, control_names: Optional[Union[List, bool]] = None):
+        """Set control coils"""
+        names = self.name
+        if isinstance(control_names, List):
+            self._control_ind = np.arange(len([names.index(c) for c in control_names]))
+        elif control_names or control_names is None:
+            self._control_ind = np.arange(len(names)).tolist()
```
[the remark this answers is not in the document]
je-cook: thanks, list-ifted

tests/equilibria/test_coils.py
```diff
@@ -41,42 +40,64 @@
 from bluemira.equilibria.grid import Grid
 
 
+def callable_tester(f_callable, coils=1):
+    """
+    Checks that all different field calls (with different inputs) behave
+    as expected
```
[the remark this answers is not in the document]
je-cook: Copied this, @CoronelBuendia what does behave as expected mean?

bluemira/equilibria/positioner.py
```diff
@@ -446,6 +461,7 @@ def _segment_tracks(lb, ub):
         sub-track into two, so that two coils don't end up on top of each other
         """
         # beware of np.zeros_like!
+        # TODO this feels tempramental
```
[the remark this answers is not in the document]
je-cook: #1709 1709

bluemira/utilities/tools.py
```diff
@@ -561,16 +561,8 @@ def consec_repeat_elem(arr: np.ndarray, num_rep: int) -> np.ndarray:
...
-    x2 = arr[1, 0]
-    y2 = arr[1, 1]
-    m = 0
-    b = x2 - x1
-    d = y2 - y1
-    if b != 0:
-        m = d / b
-    return m
+    b = arr[1, 0] - arr[0, 0]
+    return np.inf if b == 0 else arr[1, 1] - arr[0, 1] / b
```
hsaunders1904, line 565:
```suggestion
    return np.inf if b == 0 else (arr[1, 1] - arr[0, 1]) / b
```

tests/equilibria/test_coils.py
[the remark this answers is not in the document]
CoronelBuendia: Just that the different calls with float, arrays of varying length etc, all return the same result

CoronelBuendia:
OK here is a summary of stuff I have found while playing around with it so far:
* `CoilSet` has no `__repr__`
* `Coil.plot` does not work:
```python
TypeError                                 Traceback (most recent call last)
/home/matti/code/bluemira/examples/equilibria/eudemo_2017.py in line 1
----> [230](file:///home/matti/code/bluemira/examples/equilibria/eudemo_2017.py?line=229) coil.plot()

File ~/code/bluemira/bluemira/equilibria/coils/_coil.py:224, in Coil.plot(self, ax, subcoil, label, force, **kwargs)
    [199](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_coil.py?line=198) def plot(
    [200](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_coil.py?line=199)     self,
    [201](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_coil.py?line=200)     ax=None,
   (...)
    [205](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_coil.py?line=204)     **kwargs,
    [206](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_coil.py?line=205) ):
    [207](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_coil.py?line=206)     """
    [208](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_coil.py?line=207)     Plot a Coil
    [209](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_coil.py?line=208) 
   (...)
    [222](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_coil.py?line=221) 
    [223](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_coil.py?line=222)     """
--> [224](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_coil.py?line=223)     return CoilGroupPlotter(
    [225](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_coil.py?line=224)         self, ax=ax, subcoil=subcoil, label=label, force=force, **kwargs
    [226](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_coil.py?line=225)     )

File ~/code/bluemira/bluemira/equilibria/plotting.py:220, in CoilGroupPlotter.__init__(self, coil, ax, subcoil, label, force, **kwargs)
    [217](file:///home/matti/code/bluemira/bluemira/equilibria/plotting.py?line=216)     if isinstance(kwargs["alpha"], list):
    [218](file:///home/matti/code/bluemira/bluemira/equilibria/plotting.py?line=217)         kwargs["alpha"] = alpha[0]
...
    [257](file:///home/matti/code/bluemira/bluemira/equilibria/plotting.py?line=256) ):
    [258](file:///home/matti/code/bluemira/bluemira/equilibria/plotting.py?line=257)     if control_ind[i]:
    [259](file:///home/matti/code/bluemira/bluemira/equilibria/plotting.py?line=258)         if self.colors is not None:

TypeError: 'float' object is not iterable
```
* I can't discretise a `Coil` once it has been instantiated?
* I can't toggle control on members of a `CoilSet`
* If I remove a `Coil` from a `Coilset` it doesn't plot any more:
```python
IndexError                                Traceback (most recent call last)
Cell In [38], line 1
----> 1 coilset.plot()

File ~/code/bluemira/bluemira/equilibria/coils/_grouping.py:155, in CoilGroup.plot(self, ax, subcoil, label, force, **kwargs)
    [131](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_grouping.py?line=130) def plot(
    [132](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_grouping.py?line=131)     self,
    [133](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_grouping.py?line=132)     ax=None,
   (...)
    [137](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_grouping.py?line=136)     **kwargs,
    [138](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_grouping.py?line=137) ):
    [139](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_grouping.py?line=138)     """
    [140](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_grouping.py?line=139)     Plot a CoilGroup
    [141](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_grouping.py?line=140) 
   (...)
    [153](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_grouping.py?line=152)         passed to matplotlib's Axes.plot
    [154](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_grouping.py?line=153)     """
--> [155](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_grouping.py?line=154)     return CoilGroupPlotter(
    [156](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_grouping.py?line=155)         self, ax=ax, subcoil=subcoil, label=label, force=force, **kwargs
    [157](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_grouping.py?line=156)     )

File ~/code/bluemira/bluemira/equilibria/plotting.py:220, in CoilGroupPlotter.__init__(self, coil, ax, subcoil, label, force, **kwargs)
    [217](file:///home/matti/code/bluemira/bluemira/equilibria/plotting.py?line=216)     if isinstance(kwargs["alpha"], list):
    [218](file:///home/matti/code/bluemira/bluemira/equilibria/plotting.py?line=217)         kwargs["alpha"] = alpha[0]
...
--> [247](file:///home/matti/code/bluemira/bluemira/equilibria/plotting.py?line=246) control_ind[control] = True
    [249](file:///home/matti/code/bluemira/bluemira/equilibria/plotting.py?line=248) if subcoil:
    [250](file:///home/matti/code/bluemira/bluemira/equilibria/plotting.py?line=249)     qb = self._cg._quad_boundary

IndexError: index 10 is out of bounds for axis 0 with size 10
``` 
* I suspect a fair few other things may not work after removing a coil
* Probably `**_greens` should be private methods, as one has to feed a really specific input to these methods. They probably should not be exposed to the user (not really related to this PR in all fairness!)
* Not sure why I'd want to but making a `SymmetricCircuit` from a `CoilGroup` does not work:
```python
ValueError                                Traceback (most recent call last)
Cell In [63], line 1
----> 1 sym2 = SymmetricCircuit(cg)

File ~/code/bluemira/bluemira/equilibria/coils/_grouping.py:719, in SymmetricCircuit.__init__(self, *coils)
    [717](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_grouping.py?line=716) self.modify_symmetry(symmetry_line)
    [718](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_grouping.py?line=717) diff = self._symmetrise()
--> [719](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_grouping.py?line=718) self._coils[1].x -= diff[0]
    [720](file:///home/matti/code/bluemira/bluemira/equilibria/coils/_grouping.py?line=719) self._coils[1].z -= diff[1]

ValueError: output array is read-only
```
would probably be fine to remove `CoilGroup` from `SymmetricCircuit` but perhaps this construction option is something people with perfectly up-down symmetrical tokamaks would want.
* The output of `SymmetricCircuit.psi(xx, zz)` is a 3-D array somehow, same with Bx and probably Bz, and Bp if we still have it

I will try to keep bashing things around this afternoon

je-cook:
> I can't discretise a Coil once it has been instantiated?

You can by just setting the discretisation:

```python
In [1]: from bluemira.equilibria.coils import Coil,

In [2]: cc = Coil(1,2, dx=3,dz=4, current=1e5, discretisation=0.1)

In [3]: cc._quad_x
Out[3]: array([-1.95, -1.85, -1.75, ...,  3.75,  3.85,  3.95])

In [4]: cc._quad_x.size
Out[4]: 4800

In [5]: cc.discretisation = 0.2

In [6]: cc._quad_x.size
Out[6]: 1200
```

> I can't toggle control on members of a CoilSet

I could add a toggle function if you want but:

```python
control_list = coilset.control
print(control_list)
control_list.pop()
coilset.control = control_list
print(coilset.control) 
```

The rest of them don't look too bad to fix

je-cook:
there is possibly a bug in the amount of discretisation, it seems like a very high no of points is produced. Either way the docstring doesnt reflect the functionality atm

CoronelBuendia:
OK for the discretisation property. Toggling control is likely to be rare but if one can also assign somehow I guess it is OK.

When I had written the discretisation back in BLUEPRINT the number was in [m], so 0.1 for a dx=3 x dz= 4 coil would be 30 * 40 * 2 * 2 = 4800. Not sure if you have reimplemented it, but clearly the docstring does not describe what is going on.

Incidentally, this:
```python
cc = Coil(1,2, dx=3,dz=4, current=1e5, discretisation=0.1)
```
we should probably want to fail as it would cross in negative x!

je-cook:
Anything else here @CoronelBuendia @hsaunders1904 ?

bluemira/equilibria/coils/_coil.py
```diff
@@ -0,0 +1,575 @@
...
+
+    def _set_coil_attributes(self):
+        self._current_radius = 0.5 * np.hypot(self.dx, self.dz)
+        self._x_boundary, self._z_boundary = self._make_boundary(
+            self.x, self.z, self.dx, self.dz
+        )
+
+    def _rectangular_discretisation(self):
+        """
+        Discretise a coil into smaller rectangles based on the length
```
CoronelBuendia, line 477: This is more of a "filament" approach, as each sub-rectangle (as plotted) just has one current filament at its centre. There is probably a mathematical fancy name for something like this. Maybe like "uniform quadrature" except not that.

bluemira/equilibria/coils/_grouping.py
```diff
@@ -0,0 +1,917 @@
...
+        kwargs:
+            passed to matplotlib's Axes.plot
+        """
+        return CoilGroupPlotter(
+            self, ax=ax, subcoil=subcoil, label=label, force=force, **kwargs
+        )
+
+    def fix_sizes(self):
+        """
+        Fix coils sizes
```
CoronelBuendia, line 170:
```suggestion
        Fix the sizes of coils in CoilGroup
```

bluemira/equilibria/coils/_coil.py
```diff
@@ -0,0 +1,575 @@
...
+    n_turns: int
+        Number of turns
+
+    Notes
+    -----
+    If dx and dz are specified the coil size is fixed when modifying j_max or current
+
+    """
+
+    def __init__(
```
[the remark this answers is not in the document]
je-cook: well I've done it anyway was quite quick

CoronelBuendia (approved):

hsaunders1904 (approved):

PR #724, opened 2022-02-16
Revert "Refactor build stage"

sebkahn:
Reverts Fusion-Power-Plant-Framework/bluemira#624


je-cook:
I think I have a fix instead of reverting the whole PR, just cleaning up/double checking

PR #726, opened 2022-02-16
Creates a register of plotting classes

### LINKED ISSUE #531: Proposal: Reorganise displayers and plotters

The various displayers and plotters currently reside in the `display` module. This often leads to circular imports because the functionality to retrieve the plotter e.g. `_get_plotter_class` needs to know about the classes that can be plotted/displayed. This inter-dependency suggests that in reality the plotter/displayer should be in the same module as the thing that it can plot / display.

## Proposed solution

Move the specific plotters and displayers into the same modules as the things that are being plotted/displayed. It should also be possible to register the class to be plotted / displayed with its corresponding plotter without hard coding them into `_get_plotter_class` and equivalents. This could be done with a decorator to explicitly say what plotter should be used for the class being registered.

## Alternative solutions

We could probably achieve something similar by using `__init_subclass__` and directly assigning the plotter / displayer class to the class to be plotted/displayed.

## Additional context

This will likely leave the `display` module as even smaller than it currently is. It may then be worth considering moving this into `base` to reduce the number of separate modules that we have. However, as `displayer.py` depends on an external code (cadapi) this could be bad for future independence of core modules from external codes (although so much will depend on the cadapi that perhaps this is less of a concern).

DanShort12:
## Linked Issues

Closes #531 

## Description

Looks to make plotters more closely associated with their corresponding classes via the following changes:

- Creates a registry class to store mapping between class and plotter
- Helper methods to avoid exposing registry
- Move plotting classes to be along side corresponding geometry / component classes
- Register plotting classes within module

## Interface Changes

No external changes. Will allow users to associate their own plotters with their own classes (e.g. for specific components).

## Checklist

### DIFF
--- a/bluemira/base/components.py
+++ b/bluemira/base/components.py
@@ -32,7 +32,12 @@ from anytree import NodeMixin, RenderTree
 
 from bluemira.base.error import ComponentError
 from bluemira.display.displayer import DisplayableCAD
-from bluemira.display.plotter import Plottable
+from bluemira.display.plotter import (
+    BasePlotter,
+    Plottable,
+    get_plotter,
+    register_plotter,
+)
 
 if TYPE_CHECKING:
     from bluemira.geometry.base import BluemiraGeo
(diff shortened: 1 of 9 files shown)


DanShort12:
Still needs work, particularly for displayers, but @ivanmaione figured you might want to have an early look at the sorts of changes that I was suggesting.

CoronelBuendia:
Whilst the issue still remains, this solution is now too out of date to consider.

PR #788, opened 2022-03-08
Removing flux surface that doesn't intersect the wall

DarioV86:
The SPR-045 showed that in some cases the last open flux surface at the lfs nearly overlaps the first wall.
In such case no intersection are detected.
To avoid the raising error, the flux surface is removed. Being the farthest from the separatrix, it doesn't carry a relevant amount of energy.


je-cook (commented):
Seems reasonable to me. Can you add a test for this?

bluemira/radiation_transport/advective_transport.py
```diff
@@ -322,8 +322,18 @@ def _analyse_DN(self, first_wall):
...
+        for no, flux_surface in enumerate(self.flux_surfaces):
+            try:
+                flux_surface.clip(first_wall)
+            except ValueError:
+                bluemira_warn(
+                    f"No intersection detected betwenn wall and flux_surface {no}"
+                )
+                if no > len(self.flux_surfaces_ib_lfs) and no < len(
+                    self.flux_surfaces_ob_lfs
+                ):
```
sebkahn, line 334: Why the if statement ?

bluemira/radiation_transport/advective_transport.py
```diff
@@ -322,8 +322,18 @@ def _analyse_DN(self, first_wall):
...
+                    self.flux_surfaces_ob_lfs.pop(no)
+                    self.flux_surfaces_ob_hfs.pop(no)
```
sebkahn, line 336: Is this fix working for the case of a non intersecting ob flux line ?

bluemira/radiation_transport/advective_transport.py
DarioV86: In the solver, there are 4 groups of flux lines: 1) LFS_down, LFS_up, HFS_down, HFS_up.

Afterwards the flux lines from these different groups are grouped together so 
`len(flux_surfaces) = len(self.flux_surfaces_ob_lfs) + len(self.flux_surfaces_ob_hfs) + len(self.flux_surfaces_ib_lfs) + len(self.flux_surfaces_ib_hfs)`

As we know that the problem of not finding intersections happens only at the outboard, and as in this specific function we are picking flux lines from both the big pool flux_surfaces and the 4 sub-pools, I use this if statement to only check the flux lines at the outboard. 
Not sure it's clear :(

sebkahn:
Hello, here are my comments 

- There is missing inboard if statement
- Using indexes to select inboard and outboard is unsetteling me.

bluemira/radiation_transport/advective_transport.py
DarioV86: Here lfs and hfs are indicating (not correctly) lower divertor and higher divertor. So, we are still at the low field side, aka outboard, aka right, and I am popping both lower and upper part of the last open flux surface. The one that doesn't find intersections

bluemira/radiation_transport/advective_transport.py
sebkahn: This is dangerously dependent on the order of the flux lines additions.

I think it is better to replace this logic by comparing the fluxline themselves (if np.compare(flux_surface_ob_xxx[no]), flux_surface[no])  pop this thing out!

sebkahn: Else try another logic or something ...

sebkahn: Better 

Keep it like that but add a specific test !

sebkahn (approved):

je-cook (approved):

PR #798, opened 2022-03-11
⬆️ flake8-bandit upgrade

je-cook:
## Description

flake8-bandit has had a new release allowing us to upgrade bandit too

Additionally pinning versions implicitly is not possible in pre-commit at the moment: 
https://github.com/pre-commit/pre-commit    issue 582
## Interface Changes

None

## Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,36 +14,36 @@ short = "An integrated inter-disciplinary design tool for future fusion "
 with open("README.md", "r") as f:
     long = f.read()
 
-install_requires = [  # PYL = Version limited by python version
-    "asteval",  # 0.9.25
+install_requires = [
+    "asteval",
     "Babel",
-    "click",  # 8.0.3
-    "CoolProp",  # 6.4.1
-    "fortranformat",  # 1.1.1
-    "imageio",  # 2.13.5
-    "ipykernel",  # 6.6.1
-    "matplotlib",  # 3.3.4   PYL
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -14,36 +14,36 @@
...
     "neutronics-material-maker==0.1.11",  # Crash on upgrade
-    "nlopt",  # 2.7.0
-    "numba",  # 0.53.1
-    "numba-scipy",  # 0.3.0
-    "numpy",  # 1.19.5   PYL
-    "pandas",  # 1.3.5    PYL
+    "nlopt",
+    "numba",
+    "numba-scipy",
+    "numpy=1.21.5",  # numba's highest numpy
```
DanShort12, line 31: Should be as below (I guess the tests worked because we don't do an install using setuptools):

```suggestion
    "numpy==1.21.5",  # numba's highest numpy
```

DanShort12: Also perhaps more generally these should be <=x.y.z? Then people with dependencies on lower versions won't get stuck... although perhaps things will break on our side in that case so maybe best to go with fixed versions for now

DanShort12 (changes requested):
Needs a fix because conda doesn't use python standard versioning strings (so we can't copy/paste)

setup.py
DanShort12: Looks fine after fixing this issue though.

PR #747, opened 2022-02-24
Shail fix parameterisation ordering

shaildesai96:
## Linked Issues

Does this PR close or fix any Issues? Remember to create an Issue before starting work so that your fix / proposal can be addressed by the team.

e.g. Closes #{ID}

## Description

Gets wires 'pointing' in same direction for PictureFrame shapes

## Interface Changes

If you've had to update an interface or introduce a new interface as part of your change then let us know here.

## Checklist

I confirm that I have completed the following checks:

- [ ] Tests run locally and pass `pytest tests --reactor`
- [ ] Code quality checks run locally and pass `flake8` and `black .`
- [ ] Documentation built locally and checked `sphinx-build -W documentation/source documentation/build`

PR #850, opened 2022-03-23
Primitive path interpolation

### LINKED ISSUE #849: `PathInterpolator` still using discretised geometry length interpolation

`PathInterpolator` still uses a discretised interpolation of a `BluemiraWire`

## Proposed solution

Leverage new `value_at` functionality to remove discretisation in `PathInterpolator`

## Alternative solutions
It's not a big deal if we leave things as is, but it would be more accurate to use primitive geometry here.

CoronelBuendia:
## Linked Issues

Closes #849 

## Description

Adds a `parameter_at(point, tolerance)` method to `BluemiraWire`
Adds a `split_wire(wire, vertex, tolerance)` function to `geometry/tools.py` (which is used above)
Moves the `PathInterpolator` and the rather crappy `make_coil_mapper` tools to pure primitive interpolation.

Note that a coil mapping test fails because I believe the `TaperedPictureFrame` is not correctly ordered (which should be fixed in #747)

## Interface Changes
See above, no changes to existing APIs.

Apologies for the rather noisy diff.. I've recently been shown the value of parametrising tests (apologies for taking so long to come round..)

## Checklist

### DIFF
--- a/bluemira/builders/EUDEMO/pf_coils.py
+++ b/bluemira/builders/EUDEMO/pf_coils.py
@@ -298,16 +298,12 @@ def make_coil_mapper(track, exclusion_zones, coils):
         elif len(bin) == 1:
             new_segments.append(PathInterpolator(segment))
         else:
-            # Split segment: Sub-divide into BSplines for now...
-            # TODO: Actual primitive sub-division less fun...
             coils = [coils[i] for i in bin]
             l_values = [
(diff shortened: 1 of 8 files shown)


je-cook (approved):
This looks fine to me (Mainly for utilities but I had a look at it all :slightly_smiling_face: )

ivanmaione:
Fine also for me.

PR #856, opened 2022-03-24
Implement variable thickness offset function

### LINKED ISSUE #763: Implement variable thickness offset function

A function to generate a shape with a variable offset to another shape is required. The aim of the function is to enable building shapes with a variable thickness based on given offsets. This is required by components like blankets and the vacuum vessel.

A discretised version of this is implemented in `BLUEPRINT.geometry.offset.varied_angular_offset`. The functionality should be replicated using primitives.

The desired outcome is something along these lines:
* For a given planar wire, produce a variable thickness offset ranging from the inboard to the outboard about the mid-plane of the shape (centered at center of mass)
* Two thickness are specified: one on the inboard, one on the outboard
* The radius of the offset shape should never be less than the thickness at the midplane
* The transition of the offset should be continuous and smooth
* The poloidal angle at which the offset operation starts should be specified as an input
* Up-down symmetry about the mid-plane of the offset thickness may be assumed, but the shape itself may not be up-down symmetric
* 
image

hsaunders1904:
## Linked Issues

Closes #763 

## Description

Implements a variable offset function for `BluemiraWire`s. 

The function is implemented to offset wires in the xz-plane, but this could fairly easily be changed to any of the 2D planes. This will only sensibly work with convex shapes.

The function takes a minimum offset, maximum offset and two angles, between which the offset increases linearly between the minimum and maximum offset. 

It works by discretizing the wire in xz, taking the normals to the tangents at each point, then adding the offsets to the original points along the normal vectors. The new points are joined using a B-spline to make the new wire.

image

## Checklist


bluemira/geometry/varied_offset.py
```diff
@@ -0,0 +1,165 @@
...
+        The angle at which the variable offset should end, in degrees.
+    num_points: int
+        The number of points to use in the discretization of the input
+        wire.
+
+    Returns
+    -------
+    offset_wire: BluemiraWire
+        New wire at a variable offset to the input.
+    """
```
CoronelBuendia, line 72: check for acceptable input wires:
* closed
* planar  (x-z) 
* convex (??)

And raise GeometryError or warn/modify if not.

bluemira/geometry/varied_offset.py
```diff
@@ -0,0 +1,165 @@
...
+from bluemira.geometry.tools import find_clockwise_angle_2d, make_bspline
+from bluemira.geometry.wire import BluemiraWire
+
+
+def varied_offset(
+    wire: BluemiraWire,
+    min_offset: float,
+    max_offset: float,
+    min_offset_angle: float,
+    max_offset_angle: float,
```
CoronelBuendia, line 36: Probably should call these inboard / outboard rather than min and max

tests/bluemira/geometry/test_varied_offset.py
```diff
@@ -0,0 +1,142 @@
...
+            "max_offset_angle": 160,  # degrees
+        },
+        {
+            "wire": make_circle(axis=(0, 1, 0)),
+            "min_offset": 1,
+            "max_offset": 3,
+            "min_offset_angle": 90,  # degrees
+            "max_offset_angle": 140,  # degrees
+        },
+    ]
```
CoronelBuendia, line 53: Probably need some more challenging shapes here; PrincetonD, PolySpline spring to mind (almost the more likely use-case parameterisations).

bluemira/geometry/varied_offset.py
```diff
@@ -0,0 +1,165 @@
...
+    num_points: int = 200,
+) -> BluemiraWire:
```
CoronelBuendia, line 38: Debating about whether it would be nice to perform the angle transitions (optionally) relative to the center-point of the widest x-z slice, rather than the COG. This is ill-defined though, so this isn't an actionable comment - just a musing!

CoronelBuendia (changes requested):
Good to see this, certainly looks like it's working for your tests and in the pic. I've yet to have a play so no comments on the implementation. I will bash it around tomorrow and see if I can break it ;)

bluemira/geometry/varied_offset.py
CoronelBuendia: Just for fun, I tried it with a straight line. Amazingly, no errors, but probably not what one mine expect :D:
image

bluemira/geometry/varied_offset.py
CoronelBuendia: also, change `_angle` to `_degree` for consistency with other stuff.

bluemira/geometry/varied_offset.py
```diff
@@ -0,0 +1,165 @@
...
+Module containing functions to generate variable offset curves
+"""
+
+import numpy as np
+
+from bluemira.geometry.tools import find_clockwise_angle_2d, make_bspline
+from bluemira.geometry.wire import BluemiraWire
+
+
+def varied_offset(
```
CoronelBuendia, line 31: I think this should live in `tools.py` next to `offset_wire`. Alternatively both `offset` functions could live in `geometry/offset.py`; I have a feeling these won't be the only offset tools we end up with.

Thoughts on where this should live, @ivanmaione ?

bluemira/geometry/varied_offset.py
```diff
@@ -0,0 +1,165 @@
...
+    num_points: int
+        The number of points to use in the discretization of the input
+        wire.
+
+    Returns
+    -------
+    offset_wire: BluemiraWire
+        New wire at a variable offset to the input.
+    """
+    wire_coords = wire.discretize(num_points).xz
```
CoronelBuendia, line 73: So this isn't the _most_ primitive approach one could take to this problem, but given that one way or another one ends up make bsplines from a set of points, I think this is totally fine for the time being. Especially given our time pressures to produce some CAD..

CoronelBuendia (changes requested):

bluemira/geometry/varied_offset.py
hsaunders1904: 😆 I'll add a check that the wire's closed

bluemira/geometry/varied_offset.py
CoronelBuendia: @hsaunders1904 sorry to keep moving the goalposts; I think given that this function is pretty specific, it would probably best live in `builders/EUDEMO/tools.py` for now - until such a time that (also, if) we can find a completely generic variable offset.

ivanmaione: IMHO this function is not so general to be placed in geometry. Its arguments are very specific to the build of a BB cross-section in a tokamak (or similar) and that strongly limits the generality of the intended function.
For the time being, I would propose to place this function "somewhere" in builders (since it just helps in building a particular component) and, in the future, implement something a little bit more general (at least for the input parameters) in geometry.

bluemira/geometry/varied_offset.py
ivanmaione: In the case of variable distance, we cannot really escape from a bspline output due to shape topological transformation (e.g. an arc of circle could be transformed into an ellipse, at best, or another curve depending on the defined offset variation). I am thus fine with the bspline approach (unless until we don't implement a very stable offset function... then, if we have time, we can add improvements to consider some particular cases for which primitives shapes are conserved).

CoronelBuendia (commented):
Looking good, might need to redo a couple of bits down the line but we have some times pressures

CoronelBuendia (approved):

PR #857, opened 2022-03-25
🥅 less noisy parameters

je-cook:
## Description

Parameters have been quite noisy recently, this makes them less so, up for discussion. I've also considered adding a trace logging that this could go into. Thoughts welcome, will assign someone at some point after a discussion.

To test try running the EUDEMO example with and without this

## Interface Changes

## Checklist

### DIFF
--- a/bluemira/base/parameter.py
+++ b/bluemira/base/parameter.py
@@ -1335,7 +1335,7 @@ class ParameterFrame:
             and len(value) in [2, 3]
             and isinstance(value[1], (str, type(None)))
         ):
-            bluemira_debug("Consider using a dictionary for ordering safety")
+            # bluemira_debug("Consider using a dictionary for ordering safety")
             unit = value[2] if len(value) == 3 else None
             source = value[1]
             value = value[0]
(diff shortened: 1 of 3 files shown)


CoronelBuendia (approved):

bluemira/base/parameter.py
```diff
@@ -1335,7 +1335,7 @@ def _from_iterable(value):
             and len(value) in [2, 3]
             and isinstance(value[1], (str, type(None)))
         ):
-            bluemira_debug("Consider using a dictionary for ordering safety")
+            # bluemira_debug("Consider using a dictionary for ordering safety")
```
hsaunders1904, line 1338: If this is being commented, can it be removed?

hsaunders1904 (commented):
Output is far less noisy 👍 . Just a comment on a comment

bluemira/base/parameter.py
je-cook: I left it there as a note to any future developer but in principle yes, it can be removed

hsaunders1904: A note like that I would probably put in the docstring, rather than in the form of commented code

hsaunders1904 (approved):

PR #251, opened 2021-09-20
Draft: JETTO integration into bluemira

DanShort12:
In GitLab by @je-cook on Jul 22, 2021, 12:36

Closes #147

This is the first stages on enabling JETTO to be kicked off from bluemira and the data returned from it.

Possibilities

- Setup a JETTO run
  - [ ] create a template file
  - [ ] lookuptables (TODO understand what this entails)
  - [ ] ...
- Run JETTO
  - [ ] on prominence
  - [ ] in a Docker container
  - [ ] on a batch system (heimdall)
  - [ ] 'locally' where locally could be a user logged driectly into an HPC system
  - [ ] In a mocking mode as we do for PROCESS currently
- Get from JETTO run
  - [x] eqdsk
  - [ ] ...

Outside the scope of this merge is the feedback loops between JETTO, PROCESS and bluemira. This is just setting a base level for enabling functionality.

PR #903, opened 2022-04-01
✨ prominence downloader

je-cook:
## Description

Strips prominence downloader out of #251. As we talked about the other day @CoronelBuendia @DanShort12 

This may be better placed in `codes` as with the freecad and nlopt api's but I'm unsure. I can also add more to it if required but I think that could wait until we use it at some point. The main reason for this is there is no download function in the main prominence module, it only exists in the 'binary' file. 

## Interface Changes

None

## Checklist

### DIFF
--- a/bluemira/utilities/prominence.py
+++ b/bluemira/utilities/prominence.py
@@ -1,0 +1,127 @@
+# bluemira is an integrated inter-disciplinary design tool for future fusion
+# reactors. It incorporates several modules, some of which rely on other
+# codes, to carry out a range of typical conceptual fusion reactor design
+# activities.
+#
+# Copyright (C) 2021 M. Coleman, J. Cook, F. Franza, I.A. Maione, S. McIntosh, J. Morris,
+#                    D. Short
+#
+# bluemira is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# bluemira is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with bluemira; if not, see <https://www.gnu.org/licenses/>.
+"""
+Prominence API
+"""
+
+import sys
+from pathlib import Path
+from unittest.mock import patch
+
+from bluemira.base.look_and_feel import bluemira_debug, bluemira_print, bluemira_warn
+from bluemira.utilities.tools import get_module
+
+
+class ProminenceDownloader:
+    """
+    Object to import prominence's binary file to enable job data downloading.
+
+    the binary file has a lot of python code directly in it that is
+    not accessible from the prominence module.
+
+    Parameters
+    ----------
+    jobid: int
+      prominence jobid
+    save_dir: str
+        directory to save the jetto output
+    force: bool
+        overwrite existing files
+
+    """
+
+    def __init__(self, jobid, save_dir, force=False):
+        self.id = jobid
+        self.dir = False
+        self.force = force
+
+        self._save_dir = save_dir
+        self._old_open = open
+
+        self._prom_bin = self._get_binary()
+
+    def __call__(self):
+        """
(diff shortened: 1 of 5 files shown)


bluemira/utilities/prominence.py
```diff
@@ -0,0 +1,127 @@
...
+from bluemira.base.look_and_feel import bluemira_debug, bluemira_print, bluemira_warn
+from bluemira.utilities.tools import get_module
+
+
+class ProminenceDownloader:
+    """
+    Object to import prominence's binary file to enable job data downloading.
+
+    the binary file has a lot of python code directly in it that is
+    not accessible from the prominence module.
```
hsaunders1904, line 38: If the code is not accessible from the module, is there any guarantee that its API is stable? I'm not familiar with `prominence` so there may not be, but is there no way to achieve this functionality through some public API? Be that through calling the executable in a subprocess or otherwise?

bluemira/utilities/prominence.py
```diff
@@ -0,0 +1,127 @@
...
+        """
+        Download the data from a run.
+
+        Temporarily changes directory so that saving happens where desired
+        and not in working directory.
+
+        """
+        with patch("builtins.print", new=self.captured_print):
+            with patch("builtins.open", new=self.captured_open):
+                self._prom_bin.command_download(self)
```
hsaunders1904, line 71: Using `mock.patch` within production code feels very dirty.

Would `contextlib.redirect_stdout` work for print? Not sure about `open`, I'll investigate when it's not Friday afternoon.

hsaunders1904 (changes requested):
I would move this into `codes`, as this is essentially an interface to an external library. Also,

bluemira/utilities/prominence.py
```diff
@@ -0,0 +1,127 @@
...
+
+    @staticmethod
+    def _get_binary():
```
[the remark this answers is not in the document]
je-cook: Sure, the idea before was to return early if we had a valid python file (see the which comment for context). I've turned it into a yield to preserve that.

bluemira/utilities/prominence.py
```diff
@@ -0,0 +1,127 @@
...
+        """
+        Import the prominence binary directly.
+
+        There are a lot of python functions
+        that do not exist in the main module
+        """
+        for path in sys.path:
+            filepath = Path(path, "prominence")
+            if filepath.is_file():
```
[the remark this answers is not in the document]
je-cook: It is but the first returned value by which is not necessarily a valid python file and I couldnt find a way to do the equivalent to `which -a prominence` within just the python path(s)

bluemira/utilities/prominence.py
je-cook: No I cant guarantee any of that. There are two cli programs that are in the organisation https://github.com/prominence-eosc but the pinned one (which hasnt been updated in a while) isnt v1 yet. The other cli system doesnt seem to have any download functionality.

prominence (as far as I'm aware) is used by the likes of JETTO to submit jobs but may in the future be useful when we start doing parameter scans.

We could call it through subprocess but I felt it was counter intuitive to go through bash back to python especially when you may have to reauthorise on each command when this is expanded into a full API. We probably should discuss further.

bluemira/utilities/prominence.py
je-cook: I agree.

I didnt know about `contextlib.redirect_stdout` but it looks like it will be a bit more complicated to redirect the stdout into a function as it is called (definitely possible..thinking some sort of queue type thing).

tests/bluemira/utilities/test_prominence.py
```diff
@@ -0,0 +1,85 @@
...
+        with patch("builtins.open", new=self.downloader.captured_open):
+            with open("TESTFILE", "w") as testfile:
+                testfile.write("test")
+
+        with open(os.sep.join([self.downloader._save_dir, "TESTFILE"]), "r") as tf:
+            assert tf.readlines() == ["test"]
+
+
+if __name__ == "__main__":
+    pytest.main([__file__])
```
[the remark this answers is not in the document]
je-cook: Nope, deleted, v old stale code. Also got me to do it everywhere see #913

hsaunders1904 (approved):
I'm not wild about monkey patching and using undocumented APIs, but if this works, and we've got tests, then potentially it's not worth spending much more time on it.

I do think a comment, or comments, justifying the use would be good, though. Then the next person coming along (and us in a few months) is aware of why we did things this way

PR #917, opened 2022-04-04
PROCESS <=> bluemira shield/+vacuum vessel thickness inconsistency

### LINKED ISSUE #904: Vacuum vessel thickness inconsistencies with PROCESS
## Describe the bug

This is more a note that we have to address a couple of little things surrounding the inboard and outboard vacuum vessel thickness `Parameters` `tk_vv_in` and `tk_vv_out`.

These are technically mapped to PROCESS, but PROCESS has a "shield" component which is not used in EU-DEMO, and for whatever reason is still seen in the outputs (or at least our mock JSON thereof).

We should remember to set the shield thicknesses in PROCESS to 0.0, such that our input VV thicknesses travel through PROCESS correctly and come out the other side with no knowledge of this shield component.

I don't know what the PROCESS variable names are anymore, off the top of my head it was `shldth` or something.

## Steps to reproduce

Run the EU-DEMO example file with PROCESS in mock and see that the vacuum vessel is too thin :)

## Expected behaviour
I would like to be able to run the build with the desired VV thicknesses (in both mock, read, and run). We might need to do the sums of the PROCESS vv and shield thicknesses in read if the shield thicknesses are not 0.0, or catch this somehow in the mapping.

CoronelBuendia:
## Linked Issues

Closes #904 

## Description
I'm hoping to get rid of the shld+VV agglomeration that PROCESS does (on the bluemira side only). I wanted to set the shld thickness to 0.0 in PROCESS but this raises an error in PROCESS. Setting this to 1e-6 means PROCESS can run seemingly OK. I'm sure we're screwing around with some of the neutronics and cryo calcs, but they'd be relatively unimportant I reckon.

## Interface Changes

## Checklist


je-cook (approved):

je-cook:
Passed some feedback up the chain, at some point in the future we wont have to do this ~hack

Issue #585, opened 2022-01-11
Simplify common build stage pattern


DanShort12:
## Description of issue / requirement to address

A lot of the build stages for EU-DEMO (and possibly more generally) have a common pattern of (print start) / configure / [register] / run / (print complete). This leads to some code duplication that could be removed by encapsulating that pattern - this would also make it easier for a user to see how to create their own build stage.

## Proposed solution

Refactor build stages to support a standard pattern.

PR #920, opened 2022-04-04
✨ Design stage decorator

je-cook:
## Description

This removes some repeat code and improves the user experience when creating new reactors. The decorator takes care of the logging of each stage and the stage name no longer needs to be passed around.

## Interface Changes

Each design stage should now be decorated with `@Reactor.design_stage(<stage name>)`

## Checklist

### DIFF
--- a/bluemira/base/design.py
+++ b/bluemira/base/design.py
@@ -56,7 +56,7 @@ class DesignABC(abc.ABC):
     _build_config: Dict[str, BuildConfig]
     _builders: Dict[str, Builder]
     _solvers: Dict[str, FileProgramInterface]
-    _stage: List = []
+    _stage: List[str] = []
 
     def __init__(
         self,


bluemira/builders/EUDEMO/reactor.py
```diff
@@ -290,109 +265,88 @@ def build_in_vessel_component_shapes(self, component_tree: Component):
...
             self._params.to_dict(),
             config,
             cryo_vv_xz=cryo_vv_xz,
         )
-        self.register_builder(builder, name)
-        component = super()._build_stage(name)
-
-        bluemira_print(f"Completed design stage: {name}")
+        self.register_builder(builder)
+        component = super()._build_stage()
```
je-cook, line 350: Considered just directly returning `super()._build_stage()` but I dont know whether we want to make the flow and naming conventions obvious so I left it here. 

If we want to do that let me know

hsaunders1904 (changes requested):
The decorator is a good addition.

Not 100% sure how I feel about the `stage` argument being extracted into a class member instead of being passed into the `_build_stage`, etc., functions. If/when I forget to decorate my design stage, I'm probably going to be confused about why one of my stages is building twice (because `self.stage` hasn't been updated). In my opinion, using a function argument is more explicit. State is generally more difficult to keep track of (and debug) than locally scoped variables, at least for me. Let me know what you think, completely up to you.

We should also document that when you write a `build_` function, you must decorate it with the new decorator.

DanShort12 (commented):
I like the idea of this decorator - it both removes a fair bit of duplication and makes it explicit that the intent of a function is to be used as a design stage. Perhaps there's some scope in the future for adding some more common functionality into the decorator and make the body of the function just do the setup of the builder (we could even get the the stage where things are self-registering and the `run` method doesn't (necessarily) need to be written for each `Reactor` - probably too early to do that though, as things that have sub-build methods like IVC would need to be handled.

CoronelBuendia:
Can we say that this would close #585? Or would it only partially address it?

CoronelBuendia:
Can we say that this would close #585? Or would it only partially address it?

bluemira/builders/EUDEMO/reactor.py
hsaunders1904: I see what you mean, but I don't think that `_build_stage` necessarily tells us any more about side effects than `_build_stage_component`. Maybe the function name being less specific gives us some idea that something else is going on...

I agree, it's probably best to keep it as one function, you don't want inconsistent state. Like I said, it doesn't _really_ matter on a private function like this. It was more a reply to James's note than a real gripe.

je-cook:
> Can we say that this would close #585? Or would it only partially address it?

I think it definitely starts this and we could always expand the role of the decorator

je-cook:
> The decorator is a good addition.
> 
> Not 100% sure how I feel about the `stage` argument being extracted into a class member instead of being passed into the `_build_stage`, etc., functions. If/when I forget to decorate my design stage, I'm probably going to be confused about why one of my stages is building twice (because `self.stage` hasn't been updated). In my opinion, using a function argument is more explicit. State is generally more difficult to keep track of (and debug) than locally scoped variables, at least for me. Let me know what you think, completely up to you.

I considered not using `self.stage` as I agree it could lead to exactly the situation you describe, possibly deleting the attribute after a stage (within the decorator) could be the way forward and then modifying `__getattribute__` to catch a missing stage and return a understandable error. I wanted to reduce the amount a new developer had to remember, but this could be too far

bluemira/builders/EUDEMO/reactor.py
je-cook: I agree on the code smell I will directly change to

`return super()._build_stage()`

bluemira/base/design.py
```diff
@@ -192,11 +193,40 @@ def _build_stage(self, name: str) -> Component:
...
             The resulting component from the build.
         """
-        component = self._builders[name]()
-        self._params.update_kw_parameters(self._builders[name].params.to_dict())
+        component = self._builders[self.stage]()
+        self._params.update_kw_parameters(self._builders[self.stage].params.to_dict())
 
         return component
 
+    def design_stage(name: str) -> Callable:  # noqa: N805
```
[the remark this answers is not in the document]
hsaunders1904: Not sure that it matters so much, I'd probably go for the static method - not requiring an extra import is a good enough reason for me.

je-cook:
This should now work with nested design stages

hsaunders1904 (approved):
Nice addition 👍

Issue #318, opened 2021-10-07
Equilibria file saving according to COCOS-11 convention


CoronelBuendia:
## Description of issue / requirement to address

Presently, when exporting equilibria to a file format (JSON or EQDSK), I'm not sure we actually follow the most common COCOS convention (https://crppwww.epfl.ch/~sauter/cocos/). 

We also treat loading of EQDSK files according to the header names, which for different codes (e.g. CREATE-*, EFIT++, etc.) are handled differently. It would be easiest to determine which COCOS convention such codes follow and use those to convert to our convention.

COCOS-11 is the most common standard it seems, so we should endeavour to implement that as the default setting. I believe it is what ITER uses.

## Proposed solution

Implement a strategy pattern for saving and loading EQDSKs, according to the COCOS conventions. I believe this can be done almost exclusively in the `file.py` of `equilibria`. There is one potential issue when treating combinations of negative/positive plasma current and negative/positive toroidal field. At the moment these are not actually treated in `bluemira`, but we will need to add this at some point.

## Alternative solutions

Figure out which COCOS convention we "follow" and claim that we export to a standard file format...

## Additional context

It is possible that COCOS-11 deviates from the standard global coordinate system which we use elsewhere in `bluemira`. We should only change how we treat equilibrium files, and not the general geometry coordinate convention.

je-cook:
see #3339 which will enable this

geograham:
ToDo: 

- [ ] Review equilibrium and profile classes - what cocos convention is assumed if any. Files to review:
      - [ ] equilibrium.py - J
      - [ ] profiles.py - G
      - [ ] physics.py - G
      - [ ] plasma.py - G
- [ ] Select a cocos convention to use within BM - should preserve BM coordinate conventions and make simpler changes where necessary (it is probably 3, like we currently assume).
- [ ] Add checks for expected cocos directions in all location where a particular cocos convention is assumed in bluemira.
- [ ] Confirm cocos conventions for all code we take input eqdsks from and document. 
- [ ] Modify to use the source cocos convention name (e.g., JETTO, FIESTA, bluemira, etc.) rather than a number when using the eqdsk package in BM - there may need to be some name additions to the eqdsk package. 
- [ ] Pick output convention for bluemira equilibrium saveing and plotting- best to keep the same as internal, unless have a very good reason. We may need to add some warning messages in addition to the eqdsk package conversion notification.  
[...]

PR #924, opened 2022-04-06
Hotfix: Equilibrium from a FIESTA eqdsk

### LINKED ISSUE #896: Equilibrium q profile loader/calculation issue
## Describe the bug

There is a discrpancy between the way `q` profile is loaded trhough the EQ object and the EQDSK reader reading FIESTA output files.

Observed discrepancy

image

## Steps to reproduce

The following script is run to make the above plot

``` python
from pathlib import Path
import numpy as np

from bluemira.equilibria.file import EQDSKInterface
from bluemira.equilibria import Equilibrium

# Define tested file
file_path = str(Path(Path.home(), "step-blueprint-reactor-design", "SPR-045", "4", "Inputs", "SPR45_ISD1_AH.geqdsk"))

# Load via eqdsk reader
reader = EQDSKInterface()
reader_dict = reader.read(file_path)

# Get reader qpsi and psi from 0 to 1
qpsi_reader = reader_dict["qpsi"]
qpsi_size = len(qpsi_reader)
psi = np.arange(qpsi_size) / float(qpsi_size)

# Plot qpsi reader
fig, ax = plt.subplots()
ax.plot(psi, qpsi_reader, '-r', label="Reader load")
ax.set(xlabel=r"$\psi_\mathrm{N}$", ylabel=r"$q_\psi$")

# Load via Equilibrium object
EQ = Equilibrium.from_eqdsk(file_path, load_large_file=False)
qpsi_EQ = EQ.q(psi)

# Plot qpsi Equilibrium
ax.plot(psi, qpsi_EQ, '-b', label="EQ load")
ax.set(xlabel=r"$\psi_\mathrm{N}$", ylabel=r"$q_\psi$")

# Add legend
ax.legend()

fig.canvas.draw()
plt.show()
```

The following STEP file will be provided privately if UKAEA user.

## Expected behaviour

Both profiles are expected to the exactly the same!

CoronelBuendia:
## Linked Issues

Closes #896

## Description

Adds a hotfix for loading FIESTA geqdsk files, in anticipation of future refactors addressing #318 properly.

As suspected, this is a simple case of different eqdsk conventions.

Note that the q profiles do not match up _exactly_, probably due to some previously mentioned discrepancies, and due to some "lacunes" of our q profile calculations on axis and at the separatrix.
image

## Interface Changes

N/A

## Checklist

### DIFF
--- a/bluemira/equilibria/equilibrium.py
+++ b/bluemira/equilibria/equilibrium.py
@@ -227,7 +227,7 @@ class MHDState:
             psi = e["psi"]
             e["dxc"] = e["dxc"] / 2
             e["dzc"] = e["dzc"] / 2
-        elif "Fiesta" in e["name"]:
+        elif "fiesta" in e["name"].lower():
             psi = e["psi"]
         else:  # CREATE
             psi = e["psi"] / (2 * np.pi)  # V.s as opposed to V.s/rad
--- a/bluemira/equilibria/equilibrium.py
+++ b/bluemira/equilibria/equilibrium.py
@@ -227,7 +227,7 @@ class MHDState:
             psi = e["psi"]
             e["dxc"] = e["dxc"] / 2
             e["dzc"] = e["dzc"] / 2
-        elif "FIESTA" in e["name"]:
+        elif "Fiesta" in e["name"]:
             psi = e["psi"]
         else:  # CREATE
             psi = e["psi"] / (2 * np.pi)  # V.s as opposed to V.s/rad
--- a/bluemira/equilibria/equilibrium.py
+++ b/bluemira/equilibria/equilibrium.py
@@ -227,6 +227,8 @@ class MHDState:
             psi = e["psi"]
             e["dxc"] = e["dxc"] / 2
             e["dzc"] = e["dzc"] / 2
+        elif "FIESTA" in e["name"]:
+            psi = e["psi"]
         else:  # CREATE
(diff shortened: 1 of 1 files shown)


bluemira/equilibria/equilibrium.py
```diff
@@ -227,11 +227,13 @@ def _get_eqdsk(cls, filename, force_symmetry=False):
             psi = e["psi"]
             e["dxc"] = e["dxc"] / 2
             e["dzc"] = e["dzc"] / 2
+        elif "fiesta" in e["name"].lower():
```
je-cook, line 230: Can we combine this with the first if statement as they do the same thing?

je-cook (changes requested):
Hopefully this kind of thing can be extracted into some sort of higher level interface around `EQDSKInterface` in the future when we start tackling #318

bluemira/equilibria/equilibrium.py
CoronelBuendia: I'm not sure that this is the only difference, or that SCENE and FIESTA have the same convention, so I've left room for treating different codes (with different conventions) differently.

je-cook (approved):

sebkahn:
Thanks for this quick bug fix!

This is apreciated.

It is worth opening an issue on the lacunes in the q profile caluclation, do we have a separate issue already? If not it is worth having it as I can already a person we both know complaining about it in STEP plasma.

PR #928, opened 2022-04-07
Add an abstract base class for solvers

### LINKED ISSUE #759: Add an abstract base class for solvers

We currently support `FileProgrammableInterface` for integration with some external codes (PROCESS and PLASMOD). It would be helpful to generalise this interface to other internal and external interfaces so that we can expand to other use cases.

## Proposed solution

Add a `SolverABC` that can support the general interfacing for the setup/run/teardown approach. We probably need a bit more thought on where that abstract class should live (codes vs base). Also need a bit of thought around what the minimum implementation of such an interface would look like.

## Alternative solutions

We could also support a soft interface by checking whether the class being used supports setup/run/teardown.

## Additional context

This came up as part of the solver registration work but there are a bunch of other things that could be considered solvers that we will want to register (balance of plant, equilibria?, fuel cycle, ...) so it would be nice to have an interface that supports running any of those from a design (or potentially builder).

hsaunders1904:
## Linked Issues

Closes #759

## Description

Adds a `SolverABC` class.

The solver is intended to be general enough that any type of problem can run, be that an external code like PROCESS, or a Bluemira algorithm. 

I've included an example on how I propose the class could/should be used.

It's largely modelled on the `FileProgramInterface` class (which will need to be rejigged), using `Task`s that can be run in different "modes" (e.g., `RUN`, `MOCK`, `READ`).

Looking for some feedback on the design to begin with, then we can start moving existing solvers onto the new interface.

Notes:

- Another options is to pass constructed `Task`s into the solver's constructor. This would separate the concerns of construction

- I was a little torn on use of `*args, **kwargs` in the base class. I've opted to use dictionaries for now, but fundamentally that's the same thing as `**kwargs`, but doesn't portray that the overriding class can set their own inputs...

- Are all solvers going to require a setup, run, and teardown? If not, then we need to offer some way of skipping those stages. One potential solutions is to have a `NoOpTask` which does nothing, or we could remove the requirement to define the `setup_cls` and `teardown_cls` attributes.

- In the example, the `Task` instance's "run mode" methods return their outputs, and these outputs are passed along to the next `Task`. Should we instead be setting some result attribute and using that? I did try this, but found myself forgetting to set the attribute when writing up the example. It feels more familiar returning the value.

## Interface Changes

No interface changes yet, but once we've got some feedback on the solver API, we can start updating the existing solvers.

## Checklist


je-cook (commented):
I like it, cleans up some of the haphazard stuff we have at the moment.

On your points:

I prefer the tasks are set for each solver like you have here. It enforces the only real interface as through the solver itself which I prefer because it seems simpler for usability. Also a solver is specific to each task and it will be still possible to modify its methods on the fly if really required. I think its an unlikely situation but we can come back to it at some later date if required

I think kwargs is more explicit.

Technically if there is no run method for a specific task it wont run anyway. Maybe it would be good to set a default task that does nothing to avoid that confusion.

Possibly we should store the results in the execute method. Part of being able to access multiple runs. Either way for now I think it is fine

hsaunders1904:
> I think kwargs is more explicit

I agree, I think it communicates the intent more clearly.

> Possibly we should store the results in the execute method. Part of being able to access multiple runs. Either way for now I think it is fine

I've left it returning the results for now, we'll see how well that works.

`NoOpTask` has been added as well.

@je-cook If you're happy, I think this is now ready to go in. Then I can start on moving over the existing solvers.

je-cook (approved):

Issue #892, opened 2022-03-30
Support multiple build stages for a single component


CoronelBuendia:
## Description of issue / requirement to address

At present a build stage in a reactor produces a component which is added to the component reactor tree.

For some components, the build sequence actually comes in multiple different build steps, scattered throughout the reactor build procedure. E.g. the thermal shield:

* build vacuum vessel
* build vacuum vessel thermal shield (VVTS)
* ...
* build PF coils
* build cryostat thermal shield (CTS)
* build ports (thermal shield ports)
* ...

IThe ThermalShield component should have the VVTS, CTS and ports as children.

At present this is not possible.

## Proposed solution
Have separate builders for each of the sub-components, and assemble them into the (e.g.) ThermalShield component.

For this we would need to have `Component.add_child` which can support `merge_trees=True`.

## Alternative solutions

We could also have a single builder that gets called multiple times with different build steps and signatures. This could get messy and is likely to be less modular.

CoronelBuendia:
This may be possible with existing functionality. We will trial it with the example thermal shield for the EU-DEMO reactor and see how it goes

PR #933, opened 2022-04-08
Add merge method to Component

hsaunders1904:
## Linked Issues

Partially addresses #892.

## Description

Adds a `merge_children` method to components so that we can easily merge component trees within builds/designs.

This can be particularly helpful when building subcomponents of another component in multiple dimensions. This new method can be used when the xy, xz, and xyz components of each subcomponent need to be merged.

## Interface Changes

Added `merge_children` method to `Component`.

## Checklist

### DIFF
--- a/bluemira/base/components.py
+++ b/bluemira/base/components.py
@@ -137,10 +137,9 @@ class Component(NodeMixin, Plottable, DisplayableCAD):
         self: Component
             This component.
         """
-        if child.name in [_child.name for _child in self.children]:
-            raise ComponentError(
-                f"Component with name '{child.name}' already exists on '{self}'."
-            )
+        # TODO: Support merge_trees here too.
+        if child in self.children:
+            raise ComponentError(f"Component {child} is already a child of {self}")
         self.children = list(self.children) + [child]
 
         return self
--- a/bluemira/base/components.py
+++ b/bluemira/base/components.py
@@ -146,7 +146,45 @@ class Component(NodeMixin, Plottable, DisplayableCAD):
 
     def merge_children(self, other: Component):
         """
-        Merge the children the given component into this component.
+        Merge the children of the given component into this component.
+
+        For example, if this object has structure:
+
+        .. code-block::
+
+            parent_1 (Component)
+            ├── x (Component)
+            │   └── leaf_1_x (Component)
+            └── y (Component)
+                └── leaf_1_y (Component)
+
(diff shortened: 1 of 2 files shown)


CoronelBuendia (approved):

bluemira/base/components.py
```diff
@@ -158,6 +207,12 @@ def add_children(self, children: List[Component], merge_trees=False):
...
             This component.
         """
+        if isinstance(children, list) and len(children) == 0:
+            # I'm not sure what the block of code below this is for, but
+            # I assume it does something useful. However, it will throw
+            # an IndexError on an empty list, avoid that here.
+            return self
+
         if not isinstance(children, list) or len(children) == 0:
             child = children[0] if isinstance(children, list) else children
```
je-cook, line 217: any chance of a comment here @CoronelBuendia git blame tells me its apparently yours :slightly_smiling_face:?

Probably should narrow it down so we don't say "I dont know what this does" in the code

je-cook (changes requested):
looks good to me. 

One question, how is the connection between leaves reconstructed eg if you want `xy` and `leaf_x_2` is the x that is required? Is it just implicit because of the order the merge is constructed? 

If that doesn't make sense let me know and I will reword better

bluemira/base/components.py
CoronelBuendia: :O That doesn't look like me... I don't normally write 1 list if/elses... but if git blame says so, then I guess I did? If @DanShort12 has no recollection and the tests pass without it, I'd say get rid of anything we don't understand

hsaunders1904:
> @je-cook  One question, how is the connection between leaves reconstructed eg if you want xy and leaf_x_2 is the x that is required? Is it just implicit because of the order the merge is constructed?

Hmm, yeah, it's implicit. I think the behaviour for this is probably wrong.  An error may be more consistent, seeing as we raise if there are common non-leaf components. What do you think?

hsaunders1904:
Reverted [920a32a](https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/933/commits/920a32a38ecd7b09d225fa0b92129d85eee3ff00) in favour of opening a separate ticket: https://github.com/Fusion-Power-Plant-Framework/bluemira/issues/955.

je-cook:
The last commit says TODO is this still waiting on some changes?

PR #935, opened 2022-04-08
Add naughty geometry decorator

### LINKED ISSUE #929: Proposal: `debug_naughty_geometry` decorator

I'm always frustrated when a geometry operation fails with a `FreeCADError` and sometimes it can be a pain to recreate that geometry in order to understand better why the operation failed.

I thought it would be nice to have a decorator that stashes the failed operation in such a way that it is possible to inspect the state of affairs after the fact. I think it's likely we're going to get a lot of issues raised on geometry operations as they can be temperamental. So it would be nice to facilitate bug-reporting and debugging.

## Proposed solution

I've just used the following because I needed a failing offset in order to write a test, and it looks like it could be made into a useful decorator for some badly behaved geometry operations.

```python
    try:
        return BluemiraWire(
            cadapi.offset_wire(wire._shape, thickness, join, open_wire), label=label
        )
    except Exception as e:
        import json
        import os
        from bluemira.base.file import get_bluemira_path
        from datetime import datetime
        wire_dict = serialize_shape(wire)
        fp = get_bluemira_path("naughty_geometry", subfolder="generated_data")
        name = "offset_wire"+str(datetime.now())
        fn = os.sep.join([fp, name])

        with open(fn, "w") as file:
            json.dump(wire_dict, file, indent=4)
        raise e
```
Obviously, you'd want to capture the input arguments in the JSON as well.

Thoughts @ivanmaione @je-cook @hsaunders1904 ?

CoronelBuendia:
## Linked Issues

Closes #929 

Going to wait for #925 before probably refactor a little to catch the failure and allow the fallback to take place on `offset_wire`.

## Description
Adds a `debug_naughty_geometry` decorator.

## Interface Changes
Adds `generated_data/naughty_geometry.README`.
Adds the `debug_naughty_geometry` decorator to some troublesome geometry operations.
Changes `serialize_shape` slightly to put Base.Vector into a list such that it can be saved as a JSON.

## Checklist


CoronelBuendia:
This is ready now, @je-cook I've done something pretty questionable (I think) in terms of catching an error, logging the failed operation, but falling back to a viable alternative. I'm sure there's a better way..

tests/geometry/test_tools.py
```diff
@@ -519,3 +528,70 @@ def test_ValueError_given_inputs_axis_0_size_not_2(self, size, vector_name):
...
+        with open(newest, "r") as file:
+            data = json.load(file)
+
+        assert "var" in data
+        assert data["var"] == 1
+        assert "var2" in data
+        assert data["var2"] == [1, 2, 3]
+        assert "random_kwarg" in data
+        assert np.isclose(data["random_kwarg"], np.pi)
+        saved_wire = deserialize_shape(data["wire"])
```
CoronelBuendia, line 578: @ivanmaione

tests/geometry/test_tools.py
```diff
@@ -519,3 +528,70 @@ def test_ValueError_given_inputs_axis_0_size_not_2(self, size, vector_name):
...
+
+    wires = [
+        make_polygon({"x": [0, 2, 2, 0], "y": [-1, -1, 1, 1]}, closed=True),
+        make_circle(),
+        make_ellipse(),
+        PrincetonD().create_shape(),
+    ]
+
+    @pytest.mark.parametrize("wire", wires)
+    def test_file_is_made(self, wire):
```
hsaunders1904, line 558: I'm not a massive fan of using I/O in tests. It's slow and also opens the test suite up to race conditions when tests are run in parallel, which as the repo grows, we'll probably want to do if we aren't already. In fact, running this with 2 threads, the test class failed 5/5 times for me.

The usual solution is to mock out `builtins.open`. Is that doable here?

bluemira/geometry/tools.py
```diff
@@ -58,6 +63,84 @@ def convert(apiobj, label=""):
...
+        Override the JSONEncoder default object handling behaviour for BluemiraGeo.
+        """
+        if isinstance(obj, BluemiraGeo):
+            return serialize_shape(obj)
+        if isinstance(obj, np.ndarray):
+            return obj.tolist()
+        return super().default(obj)
+
+
+def debug_naughty_geometry(func):
```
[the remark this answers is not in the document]
hsaunders1904: I feel like the name `debug_naughty_geometry` is really describing what the decorator is for, but it's not what the decorator does. Maybe something more like `log_geometry_on_failure`, which is obviously less fun, but I think more descriptive. 

What do you think?

hsaunders1904 (changes requested):

tests/geometry/test_tools.py
je-cook:
```
with patch("builtins.open") as file:
    with pytest.raises(cadapi.FreeCADError):
```

then file is a mocked file writer so you can check what it is called with to re extract you data. Possibly with https://docs.python.org/3.8/library/unittest.mock.html#unittest.mock.Mock.call_args_list

je-cook (approved):
I'm happy with this, even though it is slightly abusing the error handling functionality I think its a good compromise for something that will be helpful in debugging.

hsaunders1904 (approved):

PR #962, opened 2022-04-13
Add `fallback_to` decorator for geometry operations

### LINKED ISSUE #961: Split out a `fallback` decorator from the `log_failed_geometry` decorator and apply to `make_bspline`

In #935 a decorator came in but it is doing two things: logging a failed geometry operation and allowing a fallback method to take place afterwards.

## Proposed solution

Split out the fallback handling to a second decorator. We also need to apply a fallback for make_bspline as it is proving troublesome.

## Alternative solutions

Leaving things as is is less explicit and requires more modification whenever we want to apply a log decorator to a function that has a fallback

CoronelBuendia:
## Linked Issues

Closes #961

I'll pick up changes from #963 when it goes in.

## Description

I don't know if I got lucky but this didn't take me nearly as much time as I had feared... I think I probably need to find a failing operation and make sure it works on real functions before its ready.

## Interface Changes

Adds a fallback operation for `make_bspline` which now uses `make_polygon` if an error is raised.

## Checklist


bluemira/geometry/tools.py
```diff
@@ -280,10 +304,13 @@ def interpolate_bspline(
         a bluemira wire that contains the bspline
     """
     points = Coordinates(points).T
-    return BluemiraWire(
-        cadapi.interpolate_bspline(points, closed, start_tangent, end_tangent),
-        label=label,
-    )
+    try:
```
je-cook, line 307: I think the decorator would have a better UX if you passed the error into the decorator (in this case `FreeCADError`) then you wouldnt need to modify the function to use the decorator.

eg
```python
@fallback_to(_make_polygon_fallback, cadapi.FreeCADError)
...

je-cook (changes requested):
Generally ok, just one structural comment

bluemira/geometry/tools.py
CoronelBuendia: Right but could we still log upon failure and then fallback with this approach?

CoronelBuendia: Actually perhaps we could, let me have a play!

je-cook (approved):

PR #981, opened 2022-04-21
Add pulsed PF coil current optimisation for fixed coilset design

CoronelBuendia:
## Linked Issues

This is a first pass refactor of the PF coilset optimisation problems. I've decided to stop here to keep the size of the PR down. There are probably a few spin-off PRs that can take place after this to address some useful refactors, such as:

* The dissolution of the `eq.force_field` attribute in favour of building the response matrices for coilset field and force constraints outside of `ForceField` and certainly outside of `Equilibrium`
* The refactor of equilibrium constraints to subclass directly from `OptimisationConstraint` and build their response matrices upon instantiation (and rebuild as required upon update) rather than passing the response matrices into the args of the constraint function (which is then wrapped by the user in an `OptimisationConstraint`. see #1027 
* Address the API of `eq_constraints` and `opt_constraints` in `equilibria`; they are very similar philosophically, and should probably all be `opt_constraints`. For example, the user should be able to choose between lumping a field null constraint into an L2 norm of a set of magnetic targets, or using a field null constraint directly in an optimisation problem. see #1027
* Refactor iterators see #1027
* Add a new converger for equilibria after the iterators are refactored.

I really struggled to get all of the functionality that is in this PR into a configurable design problem. There are so many layers of different objects that could be specified in different combinations. To access this all via a configurable `build_config` for {"PF Coils": {"runmode": run, "build_config": {...}} seems like the way to go, but I couldn't find a neat way of doing it and it was slowing me down. I thought it best to get something working first, then perhaps @hsaunders1904  or @je-cook could help out with plugging it all back in so that it is configurable from the `Reactor`'s `build_config`?

## Description

Adds the skeleton of a new `run.py` interface to `equilibria`.

I've flipped the objective of the optimisation problem to being the minimisation of the squared currents, with a constraint on the L2 norm of the magnetic targets, which I believe is what CREATE does. This is still debatable, and it's hard to determine a value of the rms of the L2 norm below which a plasma is "acceptable". This is in particular because we don't constrain "gaps" but rather an isoflux, and at the moment the field null is lumped in the L2-norm of the magnetic targets, meaning that the value of the constraint is in mixed units. Some of the TODOs listed above could help solve this problem.

## Interface Changes

## Checklist


bluemira/equilibria/opt_problems.py
```diff
@@ -815,3 +819,270 @@ def get_state_figure_of_merit(
         sub_opt()
         fom = sub_opt.opt.optimum_value
         return fom
+
+
+class BreakdownZoneStrategy(abc.ABC):
+    """
+    Abstract base class for the definition of a breakdown zone strategy.
+    """
```
je-cook, line 827: Some more docstring here please for the parameters

bluemira/equilibria/opt_problems.py
```diff
@@ -815,3 +819,270 @@ def get_state_figure_of_merit(
...
+        coilset: CoilSet,
+        breakdown_strategy: BreakdownZoneStrategy,
+        B_stray_max,
+        B_stray_con_tol,
+        n_B_stray_points,
+        optimiser: Optimiser = None,
+        max_currents=None,
+        constraints: List[OptimisationConstraint] = None,
+    ):
+        self.scale = 1e6  # current_scale
```
je-cook, line 954: This seems not hugely robust, say if we no longer use MA for whatever reason this will break, can this be at least configurable. The opposite also should be defined here instead of 1e-6 in the optimise method

je-cook: possible robust alternative: 

```
def __init__(..., scale="MA"):
    self.scale = raw_uc(1, scale, "A")
    self.inv_scale = raw_uc(1, "A", scale)

File: bluemira/equilibria/run.py [hunk 3]
```diff
@@ -0,0 +1,460 @@
...
+        grid: Grid,
+        coil_constraints: List[OptimisationConstraint],
+        profiles: Profile,
+        magnetic_targets: MagneticConstraintSet,
+        breakdown_strategy_cls: Type[BreakdownZoneStrategy],
+        breakdown_problem_cls: Type[PremagnetisationCOP],
+        breakdown_optimiser: Optimiser = Optimiser(
+            "COBYLA", opt_conditions={"max_eval": 5000, "ftol_rel": 1e-10}
+        ),
+        breakdown_settings: dict = {"B_stray_con_tol": 1e-8, "n_B_stray_points": 20},
```
je-cook, line 218:
```suggestion
        breakdown_settings: Dict = {"B_stray_con_tol": 1e-8, "n_B_stray_points": 20},
```

bluemira/equilibria/run.py
```diff
@@ -0,0 +1,460 @@
...
+        equilibrium_problem_cls: Type[CoilsetOptimisationProblem] = MinimalCurrentCOP,
+        equilibrium_optimiser: Optimiser = Optimiser(
+            "SLSQP", opt_conditions={"max_eval": 1000, "ftol_rel": 1e-6}
+        ),
+        equilibrium_settings: Optional[dict] = None,
```
je-cook, line 223:
```suggestion
        equilibrium_settings: Optional[Dict] = None,
```

bluemira/equilibria/run.py
```diff
@@ -0,0 +1,460 @@
...
+        """
+        R_0 = self.params.R_0.value
+        strategy = self._bd_strat_cls(
+            R_0, self.params.A.value, self.params.tk_sol_ib.value
+        )
+        coilset = deepcopy(self.coilset)
+
+        relaxed = all([c.flag_sizefix for c in coilset.coils.values()])
+        i = 0
+        while i == 0 or not relaxed:
```
je-cook, line 261: I think this needs to have a break or raise an error for a configurable certain size of iterations ie `i> num`

bluemira/equilibria/run.py
```diff
@@ -0,0 +1,460 @@
...
+            constraints = deepcopy(self._coil_cons)
+            for constraint in constraints:
+                constraint._args["eq"] = eq
+            constraints.append(L2_target_constraint)
+            problem = self._eq_prob_cls(eq, max_currents, optimiser, constraints)
+            coilset = problem.optimise()
+            self.take_snapshot(snap, eq, coilset, optimiser, self.eq_targets, profiles)
+
+
+if __name__ == "__main__":
```
je-cook, line 338: Can this whole section be moved to an example or a test? I dont think we should merge it here either way

je-cook (changes requested):
From what I can glean mostly fine. I know current time pressures but there dont seem to be many tests on this...
An example from that if main snippet would be helpful more generally.
I will try and do some running this afternoon

bluemira/equilibria/opt_problems.py
CoronelBuendia: More on this in #1027 I reckon. Trying to get rid of scale as much as possible so that it is set in only one place.

je-cook (approved):
I'm happy now with the addition of the spin of PRs. I will merge before I finish today

PR #991, opened 2022-04-25
Refactor PictureFrame parameterisation

je-cook:
## Description

This creates a factory metaclass for picture frame and removes all extra pictureframe classes while keeping the same functionality. 

I have quite extensively used lambda functions assigned to variables which flake doesnt like but I couldnt think of a cleaner way to organise the code and loads of 1 lines functions seemed overkill. Better suggestions welcomed.

I still want to add a few more tests (for failure modes) but it is ready for a general structure review

@shaildesai96 you may be interested

## Interface Changes

For complex pictureframe structures you now specify the leg type when setting up the class eg full domed looks like

```python
PictureFrame(upper="CURVED", lower="CURVED")
```

## Checklist

### DIFF
--- a/bluemira/geometry/parameterisations.py
+++ b/bluemira/geometry/parameterisations.py
@@ -29,7 +29,7 @@ import abc
 import json
 from enum import Enum
 from functools import partial
-from typing import Dict, Optional, TextIO, Union
+from typing import Dict, Iterable, Optional, TextIO, Union
 
 import numpy as np
 from scipy.special import iv as bessel
@@ -998,14 +998,14 @@ class PictureFrameTools:
 
     @staticmethod
     def _make_domed_leg(
-        x_out,
-        x_curve_start,
-        x_mid,
-        z_top,
-        z_mid,
-        r_c,
-        axis=[0, -1, 0],
-        flip=False,
+        x_out: float,
+        x_curve_start: float,
+        x_mid: float,
+        z_top: float,
+        z_mid: float,
+        r_c: float,
+        axis: Iterable[float] = (0, -1, 0),
+        flip: bool = False,
     ):
         """
         Makes smooth dome for CP coils. This includes a initial straight section
@@ -1026,7 +1026,7 @@ class PictureFrameTools:
             Vertical position of flat section [m]
         r_c: float
             Radius of corner transition. Nominally 0 [m]
-        axis: np.ndarray
+        axis: Iterable[float]
             [x,y,z] vector normal to plane of parameterisation
         flip: bool
             True if limb is lower limb of section, False if upper
@@ -1145,7 +1145,15 @@ class PictureFrameTools:
         )
 
     @staticmethod
-    def _make_flat_leg(x_in, x_out, z, r_i, r_o, axis=[0, 1, 0], flip=False):
+    def _make_flat_leg(
+        x_in: float,
+        x_out: float,
+        z: float,
+        r_i: float,
+        r_o: float,
+        axis: Iterable[float] = (0, 1, 0),
+        flip: bool = False,
+    ):
         """
         Makes a flat leg (top/bottom limb) with the option of one end rounded.
 
@@ -1161,7 +1169,7 @@ class PictureFrameTools:
             Radius of inner corner [m]
         r_o: float
             Radius of outer corner [m]
-        axis: np.ndarray
+        axis: Iterable[float]
             [x,y,z] vector normal to plane of parameterisation
         flip: bool
             True if limb is lower limb of section, False if upper
@@ -1215,7 +1223,14 @@ class PictureFrameTools:
         return BluemiraWire(wires, label=f"{label}_limb")
 
     @staticmethod
-    def _make_tapered_inner_leg(x_in, x_mid, z_in, z1, z2, axis=[0, 1, 0]):
+    def _make_tapered_inner_leg(
+        x_in: float,
+        x_mid: float,
+        z_in: float,
+        z1: float,
+        z2: float,
+        axis: Iterable[float] = (0, 1, 0),
+    ):
         """
         Makes a tapered inboard leg using a circle arc taper, symmetric about the
         midplane with the tapering beginning at a certain height and reaching a
@@ -1233,7 +1248,7 @@ class PictureFrameTools:
             Vertical position of top of limb [m]
         z2: float
             Vertical position of bottom of limb [m]
-        axis: np.ndarray
+        axis: Iterable[float]
             [x,y,z] vector normal to plane of parameterisation
 
         Returns
--- a/bluemira/geometry/parameterisations.py
+++ b/bluemira/geometry/parameterisations.py
@@ -27,7 +27,9 @@ from __future__ import annotations
 
 import abc
 import json
-from typing import TextIO, Union
+from enum import Enum
+from functools import partial
+from typing import Dict, Optional, TextIO, Union
 
 import numpy as np
 from scipy.special import iv as bessel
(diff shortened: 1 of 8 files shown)


bluemira/geometry/parameterisations.py
```diff
@@ -1052,108 +1079,90 @@ def _make_domed_leg(
...
-            out_wire.append(leg_curve)
-            out_wire.append(joint_curve)
-            out_wire.append(make_polygon([p2, p1]))
-            out_wire.append(corner_in)
-        label = "bot_limb" if flip else "top_limb"
-        return BluemiraWire(out_wire, label=label)
 
     @staticmethod
-    def _make_flat_leg(axis, x_in, x_out, z, r_i, r_o, flip=False):
+    def _make_flat_leg(x_in, x_out, z, r_i, r_o, axis=[0, 1, 0], flip=False):
```
hsaunders1904, line 1148: Default argument here is mutable - could a tuple be used instead?

bluemira/geometry/parameterisations.py
```diff
@@ -1052,108 +1079,90 @@ def _make_domed_leg(
...
-        z: np.float
+        z: float
             Vertical position of limb [m]
-        r_i: np.float
+        r_i: float
             Radius of inner corner [m]
-        r_o: np.float
+        r_o: float
             Radius of outer corner [m]
+        axis: np.ndarray
```
hsaunders1904, line 1164: Does this typing need to be as specific as this? Also, the default value is not this type. Perhaps this should be changed to `Iterable[float]` or `Iterable[int]` or similar?

bluemira/geometry/parameterisations.py
```diff
@@ -1196,39 +1205,36 @@ def _make_flat_leg(axis, x_in, x_out, z, r_i, r_o, flip=False):
...
             wires.reverse()
 
-        label = "bot_limb" if flip else "top_limb"
-
-        return BluemiraWire(wires, label=label)
+        return BluemiraWire(wires, label=f"{label}_limb")
 
     @staticmethod
-    def _make_tapered_inner_leg(axis, x_in, x_mid, z_in, z_mid_up, z_mid_down):
+    def _make_tapered_inner_leg(x_in, x_mid, z_in, z1, z2, axis=[0, 1, 0]):
```
hsaunders1904, line 1218: Mutable default value again

bluemira/geometry/parameterisations.py
```diff
@@ -1000,13 +998,13 @@ class PictureFrameTools:
...
     @staticmethod
     def _make_domed_leg(
-        axis,
         x_out,
         x_curve_start,
         x_mid,
         z_top,
         z_mid,
         r_c,
+        axis=[0, -1, 0],
```
hsaunders1904, line 1007: Mutable default argument

hsaunders1904 (changes requested):
Looks good to me, I've never really seen metaclasses used in anger before, so cool to see one doing something. Just a couple of comments.

bluemira/geometry/parameterisations.py
```diff
@@ -1016,34 +1014,63 @@ def _make_domed_leg(
...
 
         Returns
         -------
         shape: BluemiraWire
             CAD Wire of the geometry
         """
+        # Labels
+        if flip:
+            label = "bottom"
+            z_top_r_leg = lambda: z_top + r_leg
```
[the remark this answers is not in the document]
je-cook: Most of them are collected because the variables in question are defined later in the code otherwise `if flip` is peppered throughout the function. I didnt like it before and I'm not really sure on it now but I didnt come up with a nicer way to collect all the variations together

tests/geometry/test_parameterisations.py
```diff
@@ -182,6 +187,38 @@ def test_ordering(self):
...
+        wire = p.create_shape()
+        assert np.isclose(wire.length, result, rtol=1e-4, atol=1e-5)
+
+    @pytest.mark.parametrize(
+        "upper, lower", [["CURVED", "CURVED"], ["CURVED", "FLAT"], ["FLAT", "CURVED"]]
+    )
+    def test_ordering(self, upper, lower):
+        p = PictureFrame(upper=upper, lower=lower)
+        wire = p.create_shape()
+        assert _wire_edges_tangent(wire._shape)
```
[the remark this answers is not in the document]
je-cook: Yeah it probably signifies a missing function from our geometry API (not that it should do much but it should be done). We should have a wider discussion about it tomorrow.

bluemira/geometry/parameterisations.py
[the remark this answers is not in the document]
hsaunders1904: Ah, I see now - some variables used in the `lambda`s do not exist until later on. This is fine with me, just unusual to see `lambda`s like this. Potentially worth a comment explaining why they're used ("_some variables used within these lambdas are undefined at this point, using functions saves having to use `if flip .. else ..` throughout the function_"), may save some head-scratching down-the-line.

Probably not worth the time, but some potential solutions are: to have two separate functions `_make_domed_leg_bottom` and `_make_domed_leg_top` with common code refactored into helper functions, then the only `if flip` would be deciding which of the two methods to call; or just always build a "bottom" then perform a rotation after the wire is built `if flip` (I'm not familiar enough with the code to be sure how feasible that is). I'll leave it up to you to decide what you think's best/worth doing, feel free to resolve this comment when you're happy.

hsaunders1904 (approved):

bluemira/geometry/parameterisations.py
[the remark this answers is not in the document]
je-cook: Done here (not it the meta class though)

bluemira/utilities/opt_variables.py
```diff
@@ -519,6 +519,30 @@ def __getitem__(self, name) -> BoundedVariable:
...
+        try:
+            return super().__getattribute__(attr)
+        except AttributeError:
+            try:
+                return self[attr].value
+            except KeyError:
+                raise AttributeError(
+                    f"'{self.__class__.__name__}' object has no attribute '{attr}'"
+                ) from None
+
```
CoronelBuendia, line 545: What's this doing as part of this PR? Was it necessary for the refactor? Does it change user-facing behaviour?

CoronelBuendia (commented):
If the tests pass... then fine by me. That said, just because one can do just about anything in Python, doesn't mean that we should! This quite complex packaging for quite simple behaviour feels slightly unnerving. I worry that it will be brittle and difficult to understand / modify. I'm happy with it if it works and we can move forwards, but it scares me!

Looks better with fewer lambdas. Just a comment on the modification to optvariables; not sure if it is necessary for this implementation, and if so it feels like some behaviour leaking out into another module (quite a lot of that already in here in all fairness!)

bluemira/utilities/opt_variables.py
je-cook: This allows for direct value access of a variable. I thought it was a good idea to have access to both the whole variable by dictionary access and an easy way to get the variables value directly. If you want I can remove it though.

je-cook:
> If the tests pass... then fine by me. That said, just because one can do just about anything in Python, doesn't mean that we should! This quite complex packaging for quite simple behaviour feels slightly unnerving. I worry that it will be brittle and difficult to understand / modify. I'm happy with it if it works and we can move forwards, but it scares me!

I would say that it is the same as a function that creates and modifies an instance of the class and calls it (thereby the function being the outward facing UX). I preferred this way as to the user they still create a class instance for all the parameterisations. I think there is probably still some room for improvement but I can't see it right now. 

I see complex shaped pictureframes as a mostly advanced feature either way and I think we need to be very careful adding more to it in future. At least now it is easier to read the build function and if we could consolidate those inputs to be the same it may be a good future improvement.

CoronelBuendia (approved):

PR #1008, opened 2022-05-03
PLASMOD interface tweaks

CoronelBuendia:
## Linked Issues

A couple of minor tweaks to the PLASMOD interface. In particular the loop voltage should only be an output. If we need to be able to run PLASMOD with a prescribed loop voltage, we should add a model for this. I am uncomfortable with values above/below a certain number dictating different behaviour in PLASMOD. Same issue with the plasma volume, although there we definitely do want to be able to run in both modes (prescribed and calculated).

## Description

## Interface Changes

## Checklist


bluemira/codes/plasmod/mapping.py
```diff
@@ -508,6 +516,8 @@ class Profiles(Model):
     # ############################
     # [-] plasma effective charge
     "Z_eff": ("Zeff", "amu"),  # TODO check dimensionless?
+    # [V] target loop voltage (if lower than -1e-3, ignored)-> plasma loop voltage
+    "v_burn": ("v_loop", "V"),
```
CoronelBuendia, line 520: We basically never want to prescribe the loop voltage to PLASMOD. I'm not entirely sure why this exists as a mode, but I'm certain we don't want to use it. Hence the move from an IN-OUT to an OUT only.

CoronelBuendia:
I have run PLASMOD alone (with and without V_loop bugfix) with the defaults settings, and I have run PLASMOD via our interface using the same default settings (specifying all of them that we can via problem settings, with only a few coming in via parameters).



I get almost the same results (this is without the V_loop bugfix):

abs rel diff P_fus: 8.698766251038394e-08
abs rel diff v_loop: 1.0899590211803107e-07
abs rel diff q_95: 1.1928431140169382e-08

I suspect this is down to different floating point sizes in the default inputs in PLASMOD, and how we will write any float to a given precision.

bluemira/codes/plasmod/mapping.py
```diff
@@ -100,6 +100,16 @@ class SOLModel(Model):
     SICCINIO = 1
 
 
+class PLHModel(Model):
+    """
+    L-H transition power scaling model
+
+    6 - Martin
+    """
```
je-cook, line 108: For consistency: 

```suggestion
    
    Plasmod variable name: "plh"
    """
```

I was thinking it may be possible to modify `Model` to add a KEY attribute that is only accessible as `Model.KEY` but for now I think this is a good compromise

je-cook (changes requested):
Just one minor otherwise good

je-cook (approved):

je-cook:
closing and opening to trigger sonarcloud

PR #1026, opened 2022-05-06
✨ script to update pip dependencies

je-cook:
## Description

A script to update the pip dependencies to whatever is on develop. Tries to fail as early as possible and tell the user. This does not touch conda

## Interface Changes

Use `bash scripts/update_dependencies.sh`

## Checklist

### DIFF
--- a/scripts/update_dependencies.sh
+++ b/scripts/update_dependencies.sh
@@ -11,24 +11,23 @@ fi
 BM_INSTALLED=$(pip list | grep bluemira | awk 'NR==1 {print $1}')
 [ "$BM_INSTALLED" = "bluemira" ] || { echo "bluemira not found in the current python environment, exiting" && exit 1; }
 
-# check we are in the bluemira git repo
-[ $(git rev-parse --is-inside-work-tree) = "true" ] ||  { echo "Not in bluemira git repo, exiting" && exit 1; }
+readonly BLUEMIRA_ROOT="$(realpath "$(dirname "$0")"/..)"
+echo 'Bluemira directory is '$BLUEMIRA_ROOT
 
-CURRENT_REPO=$(git remote get-url origin)
-REPO_GIT_URL="git@github.com:Fusion-Power-Plant-Framework/bluemira.git"
-REPO_HTTPS_URL="https://github.com/Fusion-Power-Plant-Framework/bluemira.git"
-[ "$CURRENT_REPO" = "$REPO_GIT_URL" ] || [ "$CURRENT_REPO" = "$REPO_HTTPS_URL" ] || { echo "Not in bluemira git repo, exiting" && exit 1; }
-
+OLD_DIR=$(pwd)
+cd $BLUEMIRA_ROOT
 # check that the requirements files havent changed
-git diff --exit-code requirements.txt || { echo "requirements.txt modified on this branch, exiting" && exit 1; }
-git diff --exit-code requirements-develop.txt || { echo "requirements-develop.txt modified on this branch, exiting" && exit 1; }
+git diff --exit-code $BLUEMIRA_ROOT"/requirements.txt" || { cd $OLD_DIR && echo "requirements.txt modified on this branch, exiting" && exit 1; }
+git diff --exit-code requirements-develop.txt || { cd $OLD_DIR && echo "requirements-develop.txt modified on this branch, exiting" && exit 1; }
 
 # checkout dependencies from develop
 git checkout $REQ_BRANCH requirements.txt requirements-develop.txt
 
 # install dependencies
-pip install -r requirements.txt || { echo "pip update failed" && exit 1; }
-pip install -r requirements-develop.txt || { echo "pip update failed" && exit 1; }
+pip install -r requirements.txt || { cd $OLD_DIR &&  echo "pip update failed" && exit 1; }
+pip install -r requirements-develop.txt || { cd $OLD_DIR &&  echo "pip update failed" && exit 1; }
 
 # revert checkout of dependencies files
 git checkout $(git rev-parse --abbrev-ref HEAD) requirements.txt requirements-develop.txt
+
+cd $OLD_DIR
--- a/scripts/update_dependencies.sh
+++ b/scripts/update_dependencies.sh
@@ -13,28 +13,29 @@ else
     REQ_BRANCH="develop"
 fi
 
+
 # make sure bluemira is in the current environment
 BM_INSTALLED=$(pip list | grep bluemira | awk 'NR==1 {print $1}')
 [ "$BM_INSTALLED" = "bluemira" ] || { echo "bluemira not found in the current python environment, exiting" && exit 1; }
 
 readonly BLUEMIRA_ROOT="$(realpath "$(dirname "$0")"/..)"
-echo 'Bluemira directory is '$BLUEMIRA_ROOT
+echo "Bluemira directory is $BLUEMIRA_ROOT"
 
-OLD_DIR=$(pwd)
-cd $BLUEMIRA_ROOT # git commands dont work when not in a git directory
+GIT_CMD="git -C ""$BLUEMIRA_ROOT"
 
 # check that the requirements files havent changed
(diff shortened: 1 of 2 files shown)


CoronelBuendia:
Could we also put something in the developer docs / FAQ , along with the 1-liner for updating the conda environment if need be?

scripts/update_dependencies.sh
```diff
@@ -0,0 +1,34 @@
...
+# make sure bluemira is in the current environment
+BM_INSTALLED=$(pip list | grep bluemira | awk 'NR==1 {print $1}')
+[ "$BM_INSTALLED" = "bluemira" ] || { echo "bluemira not found in the current python environment, exiting" && exit 1; }
+
+# check we are in the bluemira git repo
+[ $(git rev-parse --is-inside-work-tree) = "true" ] ||  { echo "Not in bluemira git repo, exiting" && exit 1; }
+
+CURRENT_REPO=$(git remote get-url origin)
+REPO_GIT_URL="git@github.com:Fusion-Power-Plant-Framework/bluemira.git"
+REPO_HTTPS_URL="https://github.com/Fusion-Power-Plant-Framework/bluemira.git"
```
hsaunders1904, line 19: I'm not convinced this is the best way to determine we're in the repo. For most open source projects, `origin` will refer to your _fork_ of the repository, and `upstream` would point to the main repository.

Do we actually need this check? I'd probably opt for getting the base repo directory using `dirname` and use absolute paths everywhere. Then it doesn't matter where in the file system we are, the script always acts the same.

Something like this, maybe?

```bash
readonly BLUEMIRA_ROOT="$(realpath "$(dirname "$0")"/..)"
...
requirements_file="${BLUEMIRA_ROOT}/requirements.txt"
pip install -r "${requirements_file}"
```

scripts/update_dependencies.sh
```diff
@@ -0,0 +1,34 @@
...
+[ "$CURRENT_REPO" = "$REPO_GIT_URL" ] || [ "$CURRENT_REPO" = "$REPO_HTTPS_URL" ] || { echo "Not in bluemira git repo, exiting" && exit 1; }
+
+# check that the requirements files havent changed
+git diff --exit-code requirements.txt || { echo "requirements.txt modified on this branch, exiting" && exit 1; }
```
hsaunders1904, line 23: TIL `git diff --exit-code`. Nice!

hsaunders1904 (changes requested):

scripts/update_dependencies.sh
```diff
@@ -0,0 +1,34 @@
+set -e
+
+if [ "$1" ]
+  then
+    REQ_BRANCH="$1"
+else
+    REQ_BRANCH="develop"
+fi
```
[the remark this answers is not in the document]
je-cook: I see your point, and I sort of agree. The reason I had it like this was so that everyone is then always on the most recent dependencies irrelevant of branch, essentially how it will be when we have docker images. 

I was thinking for versioned releases I would modify this to the version tag.

I do agree its a bit counter intuitive but I dont want people o be on different dependency versions...

scripts/update_dependencies.sh
je-cook: Fair point! I was just trying to make sure we had the right repo and the right requirements file. Good catch on the origin/upstream thing. Will modify because its probably badly done and overkill protection

scripts/update_dependencies.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# make sure bluemira is in the current environment
+BM_INSTALLED=$(pip list | grep bluemira | awk 'NR==1 {print $1}')
+[ "$BM_INSTALLED" = "bluemira" ] || { echo "bluemira not found in the current python environment, exiting" && exit 1; }
+
+readonly BLUEMIRA_ROOT="$(realpath "$(dirname "$0")"/..)"
+echo 'Bluemira directory is '$BLUEMIRA_ROOT
```
hsaunders1904, line 21: I would double quote all these variable accesses to guard against word splitting.

```suggestion
echo "Bluemira directory is $BLUEMIRA_ROOT"
```

scripts/update_dependencies.sh
```diff
@@ -0,0 +1,40 @@
...
+
+OLD_DIR=$(pwd)
+cd $BLUEMIRA_ROOT # git commands dont work when not in a git directory
```
hsaunders1904, line 24: You can use git in a separate directory using `git -C /path/to/git/repo <rest of git command>`.

I would probably prefer this, as then we don't need to worry about changing directories after errors/leaving the user in a separate directory on an unexpected error. 

You could always assign `GIT_CMD="git -C \"$BLUEMIRA_REPO\"`", and use, e.g., `$GIT_CMD diff --exit-code` (untested, but I think something like this would work).

scripts/update_dependencies.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# check that the requirements files havent changed
+git diff --exit-code $BLUEMIRA_ROOT"/requirements.txt" || { cd $OLD_DIR && echo "requirements.txt modified on this branch, exiting" && exit 1; }
```
hsaunders1904, line 27: These are long commands, we could split them onto multiple lines 
```suggestion
git diff --exit-code $BLUEMIRA_ROOT"/requirements.txt" \
  || { cd $OLD_DIR && echo "requirements.txt modified on this branch, exiting" && exit 1; }
```

scripts/update_dependencies.sh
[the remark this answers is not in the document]
hsaunders1904: I guess it depends on what the purpose of the script is. If it's just so that we don't need to remember the "update" commands, I would say we should use the current branch by default. If the purpose is to get the most up-to-date dependencies from the repo, then we can stick with this.

I prefer the former, but am happy to be overruled.

hsaunders1904 (commented):
Sorry, a few nitpicks. Also, an opinion on what the default behaviour should be,

scripts/update_dependencies.sh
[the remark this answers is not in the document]
je-cook: I dont really mind @CoronelBuendia which would you prefer :slightly_smiling_face:?

hsaunders1904 (approved):
The script looks good to me now, and runs fine. 

This is good to go once @CoronelBuendia casts his vote for the desired behaviour.

scripts/update_dependencies.sh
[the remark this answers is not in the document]
CoronelBuendia: Mmm.. to be discussed in our meeting later perhaps. Not sure either. Really depends on how often we're changing our dependencies. I'm on the fence at the moment

scripts/update_dependencies.sh
```diff
@@ -0,0 +1,41 @@
...
+     || { echo "requirements.txt modified on this branch, exiting" && exit 1; }
+$GIT_CMD diff --exit-code requirements-develop.txt \
+     || { echo "requirements-develop.txt modified on this branch, exiting" && exit 1; }
+
+# checkout dependencies from develop
+$GIT_CMD checkout $REQ_BRANCH requirements.txt requirements-develop.txt
+
+# install dependencies
+pip install -r "$BLUEMIRA_ROOT/requirements.txt" || { echo "pip update failed" && exit 1; }
+pip install -r "$BLUEMIRA_ROOT/requirements-develop.txt" || { echo "pip update failed" && exit 1; }
```
CoronelBuendia, line 37: Shouldn't this install them with something like `python -m pip install --no-cache-dir -e .'[dev]'` or similar? or -U or --force-reinstall or whatever other pip flags are needed to properly update the environment? If only it were this simple! It's the pip flags and their many embodiments I struggle with on top of remembering to do this.

je-cook: All of the above flags are useful for some things like updating specific dependencies or if something isnt playing ball (eg because of some conda weirdness). That install method will install the latest (dev) version of bluemira in editable mode, not its dependencies, unless there is a change to the setup.py file. That is because we keep the setup.py file as loose as possible to not restrict developer of other codes and their needed dependency versions (with no guarantee of anything, but will probably work). 

If you install with the requirements files you get the same dependencies as our CI, it doesnt however touch the bluemira install itself which has no effect because we currently recommend editable mode

PR #1033, opened 2022-05-09
✨ Polyscope CAD viewer

### LINKED ISSUE #997: CAD viewer crash
## Describe the bug

The CAD viewer crashes 

## Steps to reproduce

1. Run any example where the `show_cad` method is run
2. CAD viewer briefly appears and then crashes
3. There is no error or traceback it just looks like the program exited 

## Expected behaviour

Be able to view the CAD

## Screenshots / Tracebacks

If applicable, add screenshots / tracebacks to help explain your problem.

## Additional Context

Things I've tried:

- Remaking my env
- rolling back the last ~15 dependency updates
- running a debugger in the show_cad function:
   -  (unhelpful as its the app._exec() that crashes everything on the last line)
   - run without adding any components, skipping this for loop: https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/63572e0bba20a5bb97acfb04444a5e1c087a2f9f/bluemira/codes/_freecadapi.py#L1711
   - running without the coin root works (ie the viewer doesnt crash, but there is nothing to display..) https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/63572e0bba20a5bb97acfb04444a5e1c087a2f9f/bluemira/codes/_freecadapi.py#L1709
 
- OS: Ubuntu 22.04
- Branch (if developing) develop

### LINKED ISSUE #1859: CAD colours (and other display options) are not carried through in builders
## Describe the bug

When running eudemo the colours for the CAD are not set properly so it is always the next colour in the cycle.
Pretty sure its a general builder issue

## Steps to reproduce

For example with the plasma. Just run the eudemo script to after the plasma creation:
```python
+-------------------------------------------------------------------------+
| Mocking PROCESS systems code run                                        |
+-------------------------------------------------------------------------+

In [1]: reactor.plasma.component().get_component('LCFS').display_cad_options
Out[1]: 
DisplayCADOptions({'color': '#4a98c9', 'transparency': 0.0}
)

In [2]: from bluemira.display.palettes import BLUE_PALETTE

In [3]: BLUE_PALETTE['PL']
Out[3]: '#cc8acc'
```

## Expected behaviour

The plasma colour should be `#cc8acc` not `#4a98c9`

- OS:  Ubuntu 22.04
- Version:  HEAD
- Branch:  develop

je-cook:
UPDATE AS OF 3/01/23

This has been updated as a separate backed of which there are now 2. 
I am getting asked more often for 'pretty' CAD ie renderings which polyscope can provide.

Side benefit we are now completely disconnected from the view irrelevant of backend. It is all dynamic.

let me know any thoughts, personally I think its worth the minor maintenance load, and the default is still freecad

## Linked Issues

Closes #1859

## Description

This PR replaces the old CAD viewer with one made with polyscope ([polyscope.run](https://polyscope.run)) which is easily installed with `pip install polyscope` all its dependencies seem to be built into the binary removing many chances for crashes/errors especially prevalent in graphical things.

Upsides:
- Quick (full reactor view ~10s)
- Pretty (materials eg wax, candy give good looking/metallic effects)
- Configurable before and during viewing (you can slice on the fly)
- antialiasing can be enable to improve quality 
- screenshot button built in
- No longer tied to freecad's viewer
- Programmable flying around object within callbacks possible
- Custom UI elements can be developed

Downsides:
- All objects have to be tessellated into a mesh to be viewed (the reason for not being faster)

~Current implementation specific (fixable with some work):~
  ~Wires in the current implementation cannot be viewed because my naiive implementation only works the the object has a face~
    ~Objects with wires without faces attached but still with some faces will only show the faces not the wires~
    ~The following two tests will fail with `--plotting-on` until this is fixed:~
        ~`tests/display/test_displayer.py::TestComponentDisplayer::test_display`~
        ~`tests/display/test_displayer.py::TestGeometryDisplayer::test_display`~ Wires now work
- We no longer have infinite inertia :smiling_face_with_tear: 

Thanks to @smason for doing the initial legwork!

## Interface Changes

A few new optional options for both viewing objects and changing the behaviour of the viewer

## Checklist


bluemira/display/displayer.py
```diff
@@ -163,13 +194,113 @@ def show_cad(
...
+        parts = [parts]
+
+    # loop over every face adding their meshes to polyscope
+    for shape_i, (part, option) in enumerate(zip(parts, options)):
+        verts, faces = cadapi.collect_verts_faces(part._shape, option["tesselation"])
+
+        m = ps.register_surface_mesh(
+            clean_name(part.label, shape_i),
+            np.vstack(verts),
+            np.vstack(faces),
```
smason, line 271: `collect_verts_faces` has already stacked everything, so no need to do it again

smason:
wow, that was quick progress!

I could have a look at pulling wires out into polyscope if that would help.  have only cared about debugging issues with faces / edges before

DanShort12:
Thanks @je-cook - looks brilliant, and nice to a bunch of new features "for free". I've tried this out on WSL and it works (albeit a bit slow / jittery) but it needs LIBGL_ALWAYS_INDIRECT set to 0 (our old displayer needed it set to 1). I think that's something to do with how polyscope sets things up before displaying and it somehow can't find the indirect OpenGL library while it does that.

bluemira/display/displayer.py
je-cook: oops yeah, thanks!

je-cook:
> wow, that was quick progress!
> 
> I could have a look at pulling wires out into polyscope if that would help. have only cared about debugging issues with faces / edges before

If you have time that would be a great help!

je-cook:
> Thanks @je-cook - looks brilliant, and nice to a bunch of new features "for free". I've tried this out on WSL and it works (albeit a bit slow / jittery) but it needs LIBGL_ALWAYS_INDIRECT set to 0 (our old displayer needed it set to 1). I think that's something to do with how polyscope sets things up before displaying and it somehow can't find the indirect OpenGL library while it does that.

Thanks for testing! Ok good to know I will put that somewhere in our docs

je-cook:
@DanShort12 in one of the menus there is a transparency mode which defaults to 'pretty' if you set it to 'none' I think you should get some speedup. I can add it as an option if that works (might do anyway..)

CoronelBuendia:
This looks really good, and at least on native Ubuntu seems to be working less jittery at all (unless you maximise the window).

I like the professional-looking window and the fact that you can play around with the display after the fact.

I don't note the same speed gains as you. For the EU-DEMO example if I just do `sector.show_cad()` after everything is already built it takes 30 seconds. This does however include the reconstruction of the CAD Matryoshkas at present though.

A few comments:
* Let's keep the original colours that we had before, i.e. no additional surface effects etc. It's important to keep the same colours between the plots and the CAD.
* We should enforce more of the defaults ourselves upon instantiation:
    * transparency mode should be None if there are no transparent components
    * the "ground plane" mode should be None with no reflection required
    * Anti-aliasing should be 1, (I couldn't seem to get 0 to work)
    * Materials should be disabled
* Tesselation should be increased 

The only potential deal-breaker is the inability to display primitives. It looks as though polyscope is effectively designed for point clouds and surface meshes. @ivanmaione may not want to lose the possibility to view primitives. I know it's a real loss for me...

CoronelBuendia:
Also thanks @je-cook for getting this up and running so quickly. Should we add Sam to contributors if we end up using this and if appropriate?

je-cook:
>   * Materials should be disabled
you have to have a material, 'clay' is the default which seems fairly normal (possibly 'wax' too) https://polyscope.run/features/materials/ other than that I havent changed the colours. I'll leave it as option to set but change the default.

>   * Anti-aliasing should be 1, (I couldn't seem to get 0 to work)
1 is off, I put it in a docstring I think

the rest I will do. I could leave the old CAD viewer in the freecad api (possibly even with the ability to change the viewer in the the displayer function). I removed it because it was broken for me.

je-cook:
fyi upping the tesselation while having any transparency effects makes the renderer quite slow

hsaunders1904:
Gave this a go on WSL, and I needed to `export LIBGL_ALWAYS_INDIRECT=0` as Dan mentioned before, but it works 🎉.

However, it's painfully slow for me. The old CAD viewer is also slow, but this was really difficult to use. I was getting as low 1.6 FPS when viewing the full EUDEMO reactor, at this speed any interactivity was basically impossible. Interactivity with the full reactor on the old viewer wasn't fast, but it was usable.

It's also more memory hungry, WSL was using nearly 5 GBs whilst displaying the sliced reactor. Compared to ~4 GBs using the old viewer.

This is obviously a nicer, more featureful viewer, it's just very slow on my PC (16 GB, 4 core 10th Gen i5, laptop). Is there anything I can tweak to try to speed it up?

image

je-cook:
> This is obviously a nicer, more featureful viewer, it's just very slow on my PC (16 GB, 4 core 10th Gen i5, laptop). Is there anything I can tweak to try to speed it up?

Yep just about to push some tweaks that should speed it up. Ive reduced the default tessellation size also and it doesnt seem to effect the speed too much locally but I would be interested to see the result elsewhere

je-cook:
Profiling wise attached is an svg from py-spy produced with `py-spy record -r 300 python examples/design/EU-DEMO/EUDEMO_reactor.py`. Github mucks with the file so its no longer interactive, ask me for the proper copy. Memory use maxed out according to btop at around 1.2-3 GB

python-2022-05-11T13:38:16+01:00

hsaunders1904:
Still very slow for me 😞. Actual generation of the display doesn't feel any slower (though I haven't done any formal benchmarks), the only issue is the interaction with the GUI. It's almost impossible for me to rotate/translate the objects. I can just about deal with the plasma on its own, that's usable, but everything else is a struggle.

Hopefully others have better luck! It could potentially be a WSL bottleneck, and it works far better on native Ubuntu. Perhaps Windows 11's [WSL GUI support](https://docs.microsoft.com/en-us/windows/wsl/tutorials/gui-apps) could improve things. I'm also wondering about having to turn off OpenGL. Are we losing hardware acceleration there? That could explain the slow-down 🤷‍♂️

smason:
I've had a bit of a poke and found out why it crashes when `LIBGL_ALWAYS_INDIRECT` isn't forced on, but it's going to be using software rendering for the time being.

Have created https://github.com/nmwsharp/polyscope/issues/171 in the hope of some suggestions

bluemira/codes/_freecadapi.py
```diff
@@ -1796,31 +1797,120 @@ def _colourise(
         _colourise(child, options)
 
 
+def collect_verts_faces(
+    solid: Part.Shape, tesselation: float = 0.1
+) -> (np.ndarray, np.ndarray):
```
hsaunders1904, line 1802:
```suggestion
) -> Tuple[np.ndarray, np.ndarray]:
```

hsaunders1904 (approved):
Looks pretty good, nice to have the option of the two back-ends. The simple reactor works pretty well with Polyscope on WSL. We'll have to give it a go with EUDEMO when that's properly back up and running

je-cook:
> We'll have to give it a go with EUDEMO when that's properly back up and running

screenshot_000000

je-cook:
> OK, just looked again, and I'm OK with it being an optional viewer, and agree with the view that the maintenance costs are worth it. There are definitely a lot of advantages to this viewer.
> 
> I have a couple of comments:
> 
> * Needs a rebase (its painless, I've just done it locally)

Done
> * Why do solids when sliced appear hollow? Is that for speed, or are we doing something funny?

Because they are created as faces which were just what was easiest when we implemented it. There is a `register_volume_mesh` function but I havent explored what else is needed to produce a volume.

> * Are we sure we want to keep polyscope as an optional dependency?
>   
>   * If so, then we need a better error than the below
>   * Preferably something like: "Please run pip install polyscope=x.x.x to use this option"
>   * If not, then it should be added to the requirements

I'll add a better soft fallback. I think it should be optional at least for now, its not something we need to require even though it is an isolated dependency.

hsaunders1904 (approved):
Looks good to me. My laptop still has to work pretty hard to render the polyscope display, but as it's now an optional backend that's not a problem.

I see that we fall back to using the FreeCAD/coin viewer if polyscope is not installed. Should we also fall back to the FreeCAD viewer if an invalid backend is passed into `show_cad`? We currently get an index error.

oliverfunk:
👏🏼

PR #1062, opened 2022-05-16
`I_not_dI` half-way house

CoronelBuendia:
## Linked Issues

@alexanderianblair figured we only presently used dI with the TIkhonov current gradient optimiser and everything else is I... let me know what you think.

## Description

## Interface Changes

## Checklist


examples/equilibria/double_null_ST.py
```diff
@@ -315,7 +315,7 @@ def init_equilibrium(grid, coilset, constraint_set):
         li=None,
     )
     constraint_set(eq)
-    optimiser = UnconstrainedTikhonovCurrentCOP(eq, constraint_set, gamma=1e-7)
+    optimiser = UnconstrainedTikhonovCurrentGradientCOP(eq, constraint_set, gamma=1e-7)
```
alexanderianblair: Missing `coilset` argument in `UnconstrainedTikhonovCurrentGradientCOP` - `double_null_ST.py` is being removed anyway in #1027, so this is not a major issue

alexanderianblair (approved):
Happy with this - I think it's much cleaner now `I_not_dI` is not needed to be passed through from the iterator, and it's clearer to the user exactly what object is being optimised.

Minor note that `double_null_ST.py` is broken here, but this is removed in #1027 anyway.

PR #1067, opened 2022-05-17
FEM Fixed Boundary

### LINKED ISSUE #478: Add coupling to PLASMOD and fixed boundary equilibrium

### LINKED ISSUE #784: fixed-boundary-equilibrium: transport solver class

This feature is the first step to close issue #478.

## Proposed solution

Implement of a `transport_solver` class that interfaces with PLASMOD (or, in future, other transport codes) and provides (at least) `p'` and `FF'`.

## Alternative solutions

Implement alternative functions that return `p'` and `FF'`.

## Additional context

It is still not clear if, in case of coupling with an external transport solver, the `transport_solver` should run internally the transport code or just take its output as input.

### LINKED ISSUE #786: fixed-boundary-equilibrium: Grad-Shafranov solver

This feature is necessary to close issue #478.

## Proposed solution

Implement a GS solver in Fenics using the magnetostatic solver implemented in #785 

## Alternative solutions

-

## Additional context

-

### LINKED ISSUE #787: fixed-boundary-equilibrium: equilibrium solver

This feature is the first step to close issue #478.

## Proposed solution

Implement an equilibrium class that iterates between a transport solver and GS ensuring convergence on specific parameters

## Alternative solutions

-

## Additional context

-

ivanmaione:
## Linked Issues

Closes #478 
Closes #786 
Closes #787 
Closes #784 

## Description

## Interface Changes

## Checklist


bluemira/equilibria/fem_fixed_boundary/utilities.py
```diff
@@ -59,30 +67,192 @@ class ScalarSubFunc(dolfin.UserExpression):
...
+    kappa:
+        array of the kappa value of the isoflux curves
+    delta:
+        array of the delta value of the isoflux curves
+    """
+    r_geo = np.zeros(len(levels))
+    kappa = np.zeros(len(levels))
+    delta = np.zeros(len(levels))
+
+    tri = Triangulation(points[:, 0], points[:, 1])
```
CoronelBuendia, line 218: I factored out the direct calls to `matplotlib.pyplot.Axis` this is a similar trick we use in `equilibria/find.py::get_contours`

I only did this because I couldn't work out why it was still plotting with `plot=False` :D

CoronelBuendia (commented):
@ivanmaione I've just had a quick look for now, I haven't tried doing anything serious yet. I came across the plotting in the calculation of delta and kappa and factored it, which is probably a little more heavy-lifting than I was expecting, but for now that is the only thing of note.

I rebased and did a little cleaning, etc. I will pick up tomorrow afternoon probably.

bluemira/equilibria/fem_fixed_boundary/equilibrium.py
```diff
@@ -0,0 +1,266 @@
...
+        # calculate kappa_95 and delta_95
+        r_geo, kappa_95, delta_95 = calculate_plasma_shape_params(
+            points, x2d_data, [np.sqrt(0.95)]
+        )
+        r_geo, kappa_95, delta_95 = r_geo[0], kappa_95[0], delta_95[0]
+
+        # calculate the iteration error
+        err_delta = abs(delta_95 - delta95_t) / delta95_t
+        err_kappa = abs(kappa_95 - kappa95_t) / kappa95_t
+        iter_err = max(err_delta, err_kappa)
```
CoronelBuendia, line 225: Perhaps something like `np.hypot(err_delta, err_kappa)` would be more stable here.

bluemira/equilibria/fem_fixed_boundary/equilibrium.py
```diff
@@ -0,0 +1,266 @@
...
+
+        # calculate the new kappa_u and delta_u
+        kappa_u_0 = builder_plasma.params.get_param("kappa_u")
+        delta_u_0 = builder_plasma.params.get_param("delta_u")
+
+        kappa_u = theta * kappa_u_0 * (kappa95_t / kappa_95) + (1 - theta) * kappa_u_0
+        delta_u = theta * delta_u_0 * (delta95_t / delta_95) + (1 - theta) * delta_u_0
```
CoronelBuendia, line 232: Hmm.. maybe a Newton-Raphson iteration approach might be more robust? I can't really explain why I think that, but I don't think it would be too hard to do here. With the modification above it would also mean a derivative would make sense.

examples/equilibria/fem_fixed_boundary/equilibrium_example.py
```diff
@@ -0,0 +1,146 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with bluemira; if not, see <https://www.gnu.org/licenses/>.
+
+"""
+An example that shows how to set up the problem for the fixed boundary equilibrium.
+"""
+
```
CoronelBuendia, line 25: When running this example I see kappa upper reaching values of > 2.3 and then it finishes with an error.

bluemira/equilibria/fem_fixed_boundary/equilibrium.py
```diff
@@ -0,0 +1,266 @@
...
+        # build the plasma
+        # - xz - plasma toroidal cross section
+        # - xyz - to get the volume
+        plasma = builder_plasma.build_xz().get_component("xz").get_component("LCFS")
+        plasma_volume = (
+            builder_plasma.build_xyz()
+            .get_component("xyz")
+            .get_component("LCFS")
+            .shape.volume
+        )
```
CoronelBuendia, line 111: I can't think when we'd ever use this without cylindrical coordinates, so the CAD here feels a little gratuitous!

What about:
```python
plasma = builder_plasma.build_xz().get_component("xz").get_component("LCFS")
plasma_volume = 2 * np.pi * plasma.shape.center_of_mass[0] * plasma.shape.area
```

bluemira/equilibria/fem_fixed_boundary/fem_magnetostatic_2D.py
```diff
@@ -183,3 +196,232 @@ def calculate_b(self):
...
+            # curr_tot = dolfin.assemble(self.g * dx())
+            # print(f"post - curr_tot = {curr_tot} - curr_dens = {self.g}")
+            bluemira_print_flush(
+                f"iter = {i} eps = {eps:.3E} psi_ax : {self.psi_ax:.2f}"
+            )
+
+            new_psi = dolfin.Function(self.V)
+            new_psi.set_allow_extrapolation(True)
+
+            self.psi.vector()[:] = (1 - relaxation) * self.psi.vector()[
```
CoronelBuendia, line 421: OK, so I understand that dolfin sometimes works in mysterious ways... It's hard for me to read this "Python" and understand that this is updating the psi vector in place! I assume that is what it is supposed to be doing.

bluemira/equilibria/fem_fixed_boundary/fem_magnetostatic_2D.py
```diff
@@ -183,3 +196,232 @@ def calculate_b(self):
...
+            eps = np.linalg.norm(diff, ord=2) / np.linalg.norm(new, ord=2)
+
+            # curr_tot = dolfin.assemble(self.g * dx())
+            # print(f"post - curr_tot = {curr_tot} - curr_dens = {self.g}")
+            bluemira_print_flush(
+                f"iter = {i} eps = {eps:.3E} psi_ax : {self.psi_ax:.2f}"
+            )
+
+            new_psi = dolfin.Function(self.V)
+            new_psi.set_allow_extrapolation(True)
```
CoronelBuendia, line 419: Scary stuff! Does this somehow enable what happens below? It looks like dead code at a glance

CoronelBuendia (commented):
Hi @ivanmaione , would you be able to discuss this at some point next week? I have lots of questions, but I think I can do a thing or two to help - just want to make sure I have understood and run it by you first.

bluemira/equilibria/fem_fixed_boundary/fem_magnetostatic_2D.py
```diff
@@ -183,3 +196,234 @@ def calculate_b(self):
...
+            target current (also used to initialize the solution in case self.psi is
+            still 0 and pprime and ffprime are, then, not defined)
+
+        Returns
+        -------
+        g: Callable
+            source current to solve the magnetostatic problem
+        """
+        dx = dolfin.Measure("dx", domain=self.mesh)
+        area = dolfin.assemble(dolfin.Constant(1) * dx())
```
CoronelBuendia, line 269: I'm not sure if this is the case, but it looks like this is where the first initial guess of the current distribution is made (i.e. constant current across the cells (??)). If this is the case, this is probably where the lowest hanging fruit is: a good starting guess makes a big difference for convergence.

Here's a potential pseudo-code way of getting a good starting guess:

* estimate the current centroid (either via a 0-D Shafranov shift estimate, or by some other means - I have seen a couple of approaches that could be useful here)
* use the flux functions and some estimate of psi_norm to calculate an initial current distribution

I've probably not fully wrapped my head around this yet though. From looking at the behaviour of the G-S iterations, though, it looks like the starting guess is not so good.

bluemira/equilibria/fem_fixed_boundary/utilities.py
```diff
@@ -59,30 +67,192 @@ class ScalarSubFunc(dolfin.UserExpression):
...
+    for i, value in enumerate(levels):
+        vertices = tcg.create_contour(value)[0][0]
+        x = vertices.T[0]
+        y = x * 0
+        z = vertices.T[1]
+        vertices = Coordinates({"x": x, "y": y, "z": z})
+        wire = interpolate_bspline(vertices, f"psi_{value:.2f}", True)
+        interp_points = wire.discretize(1000)
+
+        ind_z_max = np.argmax(interp_points.z)
```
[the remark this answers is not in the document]
ivanmaione: If there are methods already tested that do the same, I have nothing against using them.

bluemira/equilibria/fem_fixed_boundary/fem_magnetostatic_2D.py
CoronelBuendia: OK, so I've played around with this a little bit by emulating a current distribution with strong central peaking:

```python
        grid_x, grid_z = np.array(self.mesh.coordinates()).T
        min_x = np.min(grid_x)
        max_x = np.max(grid_x)
        min_z = np.min(grid_z)
        max_z = np.max(grid_z)
        r_major = min_x + 0.5 * (max_x - min_x)+0.5
        z_major = 0.0
        max_distance = np.hypot(r_major-min_x, z_major-max(abs(min_z), max_z))

        def g_initial(x):
            r = x[0]
            z = x[1]
            distance = np.hypot((r_major-r), (z_major-z))
            return max(0, 2 - np.exp(distance / max_distance))

        def g(x):
            if self.psi_ax == 0:
                return g_initial(x)
            else:
                r = x[0]
                ...
```

I also tried `(return 1 - np.sqrt(distance/max_distance)`

Two things to note:
* With the function above on the first iteration in the example the psi on axis is much closer to the converged psi on axis (172.2 V.s on first iteration vs. 159.05 V.s using the constant current density)
* This has almost no effect on the number of iterations required to reach convergence.

I am quite surprised by this! One could play around with some better dolfin functions and try and get the integral of the first guess right (if it isn't somehow adjust in Fenics already), but it seems to me like this low-hanging fruit brings nothing!

bluemira/equilibria/fem_fixed_boundary/equilibrium.py
CoronelBuendia: It doesn't seem to make much different, but worth keeping in mind when trying to find a stable scheme

examples/equilibria/fem_fixed_boundary/equilibrium_example.py
CoronelBuendia: This was resolved by reinitialising the builder

bluemira/equilibria/fem_fixed_boundary/equilibrium.py
CoronelBuendia: Done

bluemira/equilibria/fem_fixed_boundary/equilibrium.py
```diff
@@ -0,0 +1,286 @@
...
+
+        points = mesh.coordinates()
+        x2d_data = np.array([gs_solver.psi_norm_2d(x) for x in points])
+
+        # calculate kappa_95 and delta_95
+        r_geo, kappa_95, delta_95 = calculate_plasma_shape_params(
+            points, x2d_data, [np.sqrt(0.95)]
+        )
+        r_geo, kappa_95, delta_95 = r_geo[0], kappa_95[0], delta_95[0]
```
CoronelBuendia, line 216: As discussed, there is an alternative way to get the shape parameters of the 95th percentile flux surface: by optimisation (grid-searching) of the smooth psi_norm2d function. 

I attempted an implementation of this (using scipy's minimize, which is useful for sketching out optimisation problems) and the results were rather strange.
First, it seems like I can recover the etrema of the 95th flux surface very well, and get very similar results to the mesh approach. I even believe that if you were to use finer and finer mesh size, the extrema would tend towards those of the optimisation approach which for comparably speaking is instantaneous.

However, this change completely breaks the PLASMOD <-> fixed boundary convergence, with kappa and delta values reaching non-sensical values.

On paper, the convergence is improved:

image

but does not fill me with joy as I had hoped, given the resulting values of kappa and delta:
* kappa_u_final = 2.14 (!!)
* delta_u_final = 0.878 (!!)

v.s for the mesh approach:
* kappa_u_final = 1.87
* delta_u_final=0.4772

One thing to note is that changing the p_order to 1 improves the discrepancy between the opt mesh approach, but still results in crazy results for the opt approach.

This requires further investigation, as I believe it has the potential to remove one meshing step, and improve convergence. At present I simply don't understand why the result is nonsense, sadly.

bluemira/equilibria/fem_fixed_boundary/fem_magnetostatic_2D.py
```diff
@@ -183,3 +201,221 @@ def calculate_b(self):
...
+
+            if verbose_plot:
+                self._plot_array(
+                    points, prev, f"Normalized magnetic coordinate at iteration {i}"
+                )
+
+            super().solve(
+                self.g, dirichlet_bc_function, dirichlet_marker, neumann_bc_function
+            )
+            self._update_curr(curr_target)
```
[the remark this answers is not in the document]
ivanmaione: I_p is normalized through the function self._update_curr(curr_target) that modifies the normalization factor self.k

bluemira/equilibria/fem_fixed_boundary/equilibrium.py
CoronelBuendia: OK, the problem was down to a simple `np.sqrt(0.95)` vs `0.95`..!

Now it works, and appears more stable and faster across the board:
image
with the results being approximately equal in terms of kappa and delta.

bluemira/equilibria/fem_fixed_boundary/utilities.py
```diff
@@ -59,30 +67,192 @@ class ScalarSubFunc(dolfin.UserExpression):
...
+    tcg = TriContourGenerator(tri.get_cpp_triangulation(), psi)
+
+    for i, value in enumerate(levels):
+        vertices = tcg.create_contour(value)[0][0]
+        x = vertices.T[0]
+        y = x * 0
+        z = vertices.T[1]
+        vertices = Coordinates({"x": x, "y": y, "z": z})
+        wire = interpolate_bspline(vertices, f"psi_{value:.2f}", True)
+        interp_points = wire.discretize(1000)
```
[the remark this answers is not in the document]
CoronelBuendia: Resolved with the extrema finding via optimisation

bluemira/equilibria/fem_fixed_boundary/fem_magnetostatic_2D.py
CoronelBuendia: Potential future work, but not necessary at present

bluemira/equilibria/fem_fixed_boundary/equilibrium.py
CoronelBuendia: OK, done now

bluemira/equilibria/fem_fixed_boundary/fem_magnetostatic_2D.py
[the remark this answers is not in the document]
CoronelBuendia: OK, this is true, and I have moved `self._update_curre(curr_target)` to after the extraction of psi values, as discussed. This does mean that the psi values (but not psi_norm) are technically not consistent with the plasma current within a given iteration (or at least this is my possibly flawed interpretation of things!). Either way, it does not seem to have any real-world effect on the convergence.

bluemira/equilibria/fem_fixed_boundary/equilibrium.py
```diff
@@ -0,0 +1,256 @@
...
+        if plot:
+            plasma.plot_options.show_faces = False
+            plasma.plot_2d(show=True)
+
+        kappa_u = builder_plasma.params.get_param("kappa_u")
+        kappa_l = builder_plasma.params.get_param("kappa_l")
+        delta_u = builder_plasma.params.get_param("delta_u")
+        delta_l = builder_plasma.params.get_param("delta_l")
+
+        kappa = (kappa_u + kappa_l) / 2
```
CoronelBuendia, line 109: Not sure where to go about this (probably in the input rather than in here), but one way of speeding up convergence would be to set the starting kappa_u and kappa_l such that they equal 1.12 * kappa_95, which is a rule of thumb used in PROCESS based on some FIESTA scans. Of course, there is the awkward decision of how to split between kappa_u and kappa_l, but that is not going to go away for SNs using the JohnerLCFS parameterisation.

bluemira/equilibria/fem_fixed_boundary/equilibrium.py
```diff
@@ -0,0 +1,256 @@
...
+        _, kappa_95, delta_95 = calculate_plasma_shape_params(
+            gs_solver.psi_norm_2d,
+            mesh,
+            np.sqrt(0.95),
+        )
+
+        # calculate the iteration error
+        err_delta = abs(delta_95 - delta95_t) / delta95_t
+        err_kappa = abs(kappa_95 - kappa95_t) / kappa95_t
+        iter_err = max(err_delta, err_kappa)
```
CoronelBuendia, line 191: In future, we could use a `ConvergenceCriterion` (as in equilibria/solve.py), but it is not necessary for this PR.

In future, we could use a more generic version of the `PicardIterator` (as in equilibria/solve.py) but right now this functionality only makes sense for free boundary equilibria with coils (as they are used to stabilise the problem, in crude terms) - none of which is necessary for this Picard iteration.

bluemira/equilibria/fem_fixed_boundary/equilibrium.py
```diff
@@ -0,0 +1,256 @@
...
+            print(f"bluemira delta95 = {delta_95}")
+            print(f"target delta95 = {delta95_t}")
+
+            print(f"|Target - bluemira|/Target = {err_delta}")
+
+            print(" ")
+            print(f"bluemira kappa95 = {kappa_95}")
+            print(f"target kappa95 = {kappa95_t}")
+
+            print(f"|Target - bluemira|/Target = {err_kappa}")
```
CoronelBuendia, line 221: Perhaps this is a little too verbose even for `verbose=True`? With the `bluemira_debug` this is largely already captured. There is a lot going on in terms of printing to the console at the moment, perhaps as this emerges from development we could look to condense the output a little more than I have already done?

bluemira/equilibria/fem_fixed_boundary/fem_magnetostatic_2D.py
```diff
@@ -183,3 +200,255 @@ def calculate_b(self):
...
+        plt.show()
+
+    def solve(
+        self,
+        pprime,
+        ffprime,
+        curr_target,
+        dirichlet_bc_function=None,
+        dirichlet_marker=None,
+        neumann_bc_function=None,
```
CoronelBuendia, line 359: I don't see anybody changing these defaults any time soon..! Obviously I've only been running the one case and never changing this, so it's as you see fit.

bluemira/equilibria/fem_fixed_boundary/fem_magnetostatic_2D.py
```diff
@@ -183,3 +200,255 @@ def calculate_b(self):
...
+        self.g = self._create_g(pprime, ffprime, curr_target)
+
+        super().solve(
+            self.g, dirichlet_bc_function, dirichlet_marker, neumann_bc_function
+        )
+        self._update_curr(curr_target)
+
+        eps = 1.0
+        i = 0  # iteration counter
+        while eps > iter_err_max and i < max_iter:
```
CoronelBuendia, line 416: As above:



In future, we could use a ConvergenceCriterion (as in equilibria/solve.py), but it is not necessary for this PR.

In future, we could use a more generic version of the PicardIterator (as in equilibria/solve.py) but right now this functionality only makes sense for free boundary equilibria with coils (as they are used to stabilise the problem, in crude terms) - none of which is necessary for a fixed boundary problem, or for this PR.

bluemira/equilibria/fem_fixed_boundary/fem_magnetostatic_2D.py
```diff
@@ -183,3 +200,255 @@ def calculate_b(self):
...
+
+    @property
+    def psi_ax(self):
+        """Poloidal flux on the magnetic axis"""
+        return np.max(self.psi.vector()[:])
+
+    @property
+    def psi_b(self):
+        """Poloidal flux on the boundary"""
+        return 0.0  # np.min(self.psi.vector()[:])
```
CoronelBuendia, line 233: Your call! I did this for sanity while playing around with stuff and forgot to remove... just realising that if somebody passes a different BC into the solve (see above comment) which isn't 0.0, this could cause a massive problem! Sorry for messing this up :face_with_head_bandage:

bluemira/equilibria/fem_fixed_boundary/utilities.py
```diff
@@ -59,30 +68,255 @@ class ScalarSubFunc(dolfin.UserExpression):
...
+        ax.plot(*pi, marker="o", color="r")
+        ax.plot(*pu, marker="o", color="r")
+        ax.plot(*pl, marker="o", color="r")
+
+        ax.plot(*po_opt, marker="o", color="b")
+        ax.plot(*pi_opt, marker="o", color="b")
+        ax.plot(*pu_opt, marker="o", color="b")
+        ax.plot(*pl_opt, marker="o", color="b")
+        ax.set_aspect("equal")
+        plt.show()
```
CoronelBuendia, line 300: This was mostly for debugging, I left it in in case you wanted to play around with it, but please feel free to remove.

CoronelBuendia (commented):
@ivanmaione OK I have finished my frantic playing around with things / tweaking stuff. I will move on to other stuff now for the rest of the week before I am off.

Really nice to see this in such a good state. I am looking forward to be able to plug it in to the EU-DEMO calculation chain.

Probably the most important thing is a few tests... You mentioned you came some way implementing a Solov'ev verification, and I'd be very happy with that as a test! (I just remembered I had to send you a reference for that - I will look again).

I think there are some aspects that we can improve on (speed, convergence, etc.), but hopefully it is sufficiently stable as is to start using it seriously.

bluemira/equilibria/fem_fixed_boundary/utilities.py
```diff
@@ -59,30 +68,255 @@ class ScalarSubFunc(dolfin.UserExpression):
...
+        """
+        Constraint function for points on the psi_norm surface.
+        """
+        return psi_norm_func(x) - psi_norm
+
+    def find_extremum(func, x0):
+        """
+        Extremum finding using constrained optimisation
+        """
+        # TODO: Replace scipy minimize with something a little more robust
```
CoronelBuendia, line 260: See #1253

ivanmaione:
> Some things I found on the dolfin -> dolfinx conversion for the PR after this:
> 
> ```
> dolfin.dot                         -> ufl.dot
> dolfin.dx                          -> ufl.dx
> dolfin.Expression                  -> dolfinx.fem.Expression
> dolfin.Function                    -> dolfinx.fem.Function
> dolfin.FunctionSpace               -> dolfinx.fem.FunctionSpace
> dolfin.grad                        -> ufl.grad
> dolfin.mesh                        -> dolfinx.mesh
> dolfin.Mesh                        -> dolfinx.mesh.Mesh
> dolfin.MeshFunction
> dolfin.pi                          -> np.pi
> dolfin.TestFunction                -> ufl.TestFunction
> dolfin.TrialFunction               -> ufl.TrialFunction
> dolfin.ds                          -> ufl.ds
> dolfin.DirichletBC                 -> dolfinx.fem.dirichletbc
> dolfin.solve  
> dolfin.VectorFunctionSpace         -> dolfinx.fem.VectorFunctionSpace
> dolfin.project
> dolfin.as_vector                   -> ufl.as_vector
> dolfin.Constant                    -> dolfinx.fem.Constant
> dolfin.Measure                     -> ufl.Measure
> dolfin.assemble                    -> dolfinx.fem.assemble_matrix, dolfinx.fem.assemble_scalar, dolfinx.fem.assemble_vector 
> dolfin.cpp.mesh.MeshFunctionSizet
> dolfin.MeshValueCollection         -> Possibly dolfinx.mesh.meshtags
> dolfin.XDMFFile                    -> dolfinx.io.XDMFFile
> dolfin.plot                        -> dolfinx.plot
> dolfin.UserExpression              -> Possibly dolfinx.fem.Expression
> ```

Thanks, James. It seems a useful conversion table.

CoronelBuendia (approved):

bluemira/equilibria/fem_fixed_boundary/fem_magnetostatic_2D.py
CoronelBuendia: A fix is in #1404

CoronelBuendia:
OK, @hsaunders1904, all yours! We split one issue out (#1448) because we're not sure how best to treat and what changes might be coming.

There are a couple of other technical changes we're planning on making in future PRs, but these will be "under the hood".

bluemira/equilibria/fem_fixed_boundary/transport_solver.py
```diff
@@ -0,0 +1,186 @@
...
+    # This abstract class specifies all the properties and methods that
+    # should be provided by a generic transport solver. For the moment
+    # they are limited to pprime and ffprime. It is only a draft version.
+    # TODO: see #1448
+
+    def __init__(self, *args, **kwargs):
+        self.solver = None
+
+    @property
+    @abstractmethod
```
hsaunders1904, line 46: Could this be an `@abstractproperty`?

bluemira/equilibria/fem_fixed_boundary/transport_solver.py
```diff
@@ -0,0 +1,186 @@
...
+    def ffprime(self):
+        """
+        Get ffprime as function of psi_norm
+        """
+        pass
+
+
+class NoneTransportSolver(TransportSolver):
```
hsaunders1904, line 62: Is this used anywhere (do we need it?)?

bluemira/equilibria/fem_fixed_boundary/transport_solver.py
```diff
@@ -0,0 +1,186 @@
...
+        psi_ax))
+        """
+        profile_data = self.solver.get_profile(profile)
+        interp_data = interp1d(self.__x_psi, profile_data, kind="linear")
+        return interp_data(self.__x_phi)
+
+    @property
+    def x(self):
+        """Get the magnetic coordinate"""
+        # This variable is "reduntant" since it is equale to the renormalized x_phi.
```
hsaunders1904, line 126:
```suggestion
        # This variable is "redundant" since it is equal to the renormalized x_phi.
```

hsaunders1904: Is this comment worth putting in the docstring? I can't see another way to publicly access `x_phi` though, so this doesn't appear to be redundant? Or am I missing something?

bluemira/equilibria/fem_fixed_boundary/utilities.py
```diff
@@ -59,30 +68,345 @@ class ScalarSubFunc(dolfin.UserExpression):
...
+
+def plot_profile(x, prof, var_name, var_unit):
+    """
+    Plot profile
+    """
+    fig, ax = plt.subplots()
+    ax.plot(x, prof)
+    ax.set(xlabel="x (-)", ylabel=var_name + " (" + var_unit + ")")
+    ax.grid()
+    plt.show()
```
hsaunders1904, line 177: We may want to consider making this `show` optional, as we usually do for plotting.

bluemira/equilibria/fem_fixed_boundary/utilities.py
```diff
@@ -59,30 +68,345 @@ class ScalarSubFunc(dolfin.UserExpression):
...
+
+
+def calculate_plasma_shape_params(psi_norm_func, mesh, psi_norm, plot=False):
+    """
+    Calculate the plasma parameters (r_geo, kappa, delta) for a given magnetic
+    isoflux using optimisation.
+
+    Parameters
+    ----------
+    psi_norm_func: callable
```
hsaunders1904, line 219:
```suggestion
    psi_norm_func: Callable
```

examples/equilibria/fem_fixed_boundary/equilibrium_example.py
```diff
@@ -0,0 +1,150 @@
...
+    "kappa_95": 1.652,
+    "delta_95": 0.333,
+    "delta": (main_params["delta_l"] + main_params["delta_u"]) / 2,
+    "kappa": (main_params["kappa_l"] + main_params["kappa_u"]) / 2,
+    "q_95": 3.25,
+    "f_ni": 0,
+}
+
+# plasmod options
+PLASMOD_PATH = "/home/ivan/Desktop/bluemira_project/plasmod/bin/"
```
hsaunders1904, line 94: Absolute path should be removed

examples/equilibria/fem_fixed_boundary/equilibrium_example.py
```diff
@@ -0,0 +1,150 @@
...
+PLASMOD_PATH = PLASMOD_PATH = os.path.join(
```
hsaunders1904, line 95:
```suggestion
PLASMOD_PATH = os.path.join(
```

examples/equilibria/fem_fixed_boundary/transport_solver_tutorial.py
```diff
@@ -0,0 +1,97 @@
...
+
+# # A simple example to run Plasmod
+
+# Define plasmod binary path
+
+# %%
+
+set_log_level("DEBUG")
+
+PLASMOD_PATH = PLASMOD_PATH = os.path.join(
```
hsaunders1904, line 51:
```suggestion
PLASMOD_PATH = os.path.join(
```

hsaunders1904 (changes requested):
I'm a little concerned about the reliance of the solver on the plasma builder; it feels to me that the equilibria module shouldn't care about builders. Would passing in an `Equilibrium` give us the same functionality?

Given time constraints, for the upcoming development release this is probably fine as currently structured, but I feel strongly that we should remove the dependency on builders for the 0.1.0 release. The plasma builder is changing quite a bit as well, so I think this will break given the new builder/designer architecture; so it will need looking at again regardless.

Otherwise, just a few minor things I spotted.

CoronelBuendia:
Thanks, @hsaunders1904. I will pick up tomorrow and spin anything I can't handle into a separate issue for future releases.

bluemira/equilibria/fem_fixed_boundary/transport_solver.py
CoronelBuendia: @ivanmaione , we don't. Any thoughts?

bluemira/equilibria/fem_fixed_boundary/equilibrium.py
```diff
@@ -0,0 +1,256 @@
...
+    verbose=False,
+):
+    """
+    Solve the plasma fixed boundary problem using delta95 and kappa95 as target
+    values and iterating on PLASMOD to have consistency with pprime and ffprime.
+
+    Parameters
+    ----------
+    builder_plasma: Builder
+        Plasma poloidal cross section builder object
```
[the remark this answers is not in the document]
CoronelBuendia: See #1454

hsaunders1904 (approved):
With https://github.com/Fusion-Power-Plant-Framework/bluemira/issues/1454 open, I'm happy to ✔️

Issue #1049, opened 2022-05-13
Inconsistent results in EUDEMO reactor example


hsaunders1904:
## Describe the bug

Since #980, the `EUDEMO_reactor.py` example's results have been inconsistent. 

I've seen three outcomes:

1. The build runs fine, and everything is as expected.
2.  I get a `bluemira.geometry.error.DisjointedFace` error. [\#1045](#1045) seems relevant.
		<details>
		  <summary>Stack trace</summary>

		Traceback (most recent call last):
		  File "examples/design/EU-DEMO/EUDEMO_reactor.py", line 319, in <module>
			component = reactor.run()
		  File "/home/bf2936/bluemira/code/bluemira/bluemira/builders/EUDEMO/reactor.py", line 84, in run
			component.add_child(self.build_VV_thermal_shield(component))
		  File "/home/bf2936/bluemira/code/bluemira/bluemira/base/design.py", line 240, in wrapper
			ret = func(self, *args, **kwargs)
		  File "/home/bf2936/bluemira/code/bluemira/bluemira/builders/EUDEMO/reactor.py", line 243, in build_VV_thermal_shield
			vvts = super()._build_stage()
		  File "/home/bf2936/bluemira/code/bluemira/bluemira/base/design.py", line 213, in _build_stage
			component = self._builders[self.stage]()
		  File "/home/bf2936/bluemira/code/bluemira/bluemira/base/builder.py", line 113, in __call__
			return self.build()
		  File "/home/bf2936/bluemira/code/bluemira/bluemira/builders/thermal_shield.py", line 89, in build
			component.add_child(self.build_xyz())
		  File "/home/bf2936/bluemira/code/bluemira/bluemira/builders/thermal_shield.py", line 148, in build_xyz
			vvts_face = self._vvts_face.deepcopy()
		  File "/home/bf2936/bluemira/code/bluemira/bluemira/geometry/face.py", line 73, in deepcopy
			return BluemiraFace(boundary, self.label)
		  File "/home/bf2936/bluemira/code/bluemira/bluemira/geometry/face.py", line 50, in __init__
			self._create_face()
		  File "/home/bf2936/bluemira/code/bluemira/bluemira/geometry/face.py", line 102, in _create_face
			raise DisjointedFace("Any or more than one face has been created.")
		bluemira.geometry.error.DisjointedFace: Any or more than one face has been created.

3. No error, but the build does not look correct.
    <details>
[...]

PR #1079, opened 2022-05-20
Increase robustness of EU-DEMO build and fix closure problem

### LINKED ISSUE #1078: Some closure inconsistencies may be causing bugs or cause bugs in future.
## Describe the bug

When a wire is closed in the primitive form, I assume this means that its start and end points are coincident to within some tolerance. I do not know if FreeCAD automatically "snaps" them to be mathematically identical, but I suspect not.

I've noticed that when discretising a closed wire with `byedges=False` the resulting numpy array (under the hood) is not closed in such a way that when passed into `Coordinates` it is considered open (i.e. that `np.allclose(points[0], points[-1], rtol=0, atol=EPS) == False`

If I pass this "open" set of `Coordinates` back into `make_polygon`, the resulting wire is indeed closed (even with `make_polygon(coordinates, closed=False)`.

## Steps to reproduce

```python
from bluemira.geometry.parameterisations import TripleArc
from bluemira.builders.EUDEMO.tools import varied_offset

wire = TripleArc().create_shape()
coordinates = wire.discretize(ndiscr=100, byedges=False)

offset_wire = varied_offset(wire, 0.2, 0.6, 60, 160, num_points=200)

print(offset_wire.is_closed())

offset_coordinates = offset_wire.discretize(ndiscr=100, byedges=False)
print(offset_coordinates.closed)
```

## Expected behaviour

I expect the discretisation of closed wire to be closed within machine precision. Alternatively I expect that our "machine precision for geometry" be the same across FreeCAD and discretised representations.

## Additional context
I'm investigating the various issues we are seeing in the EU-DEMO build where occasionally we get open wires, disjointed faces, reconstruction errors, etc. I'm not sure that this is related, but I'm trying to bomb-proof stuff as much as possible!

@ivanmaione FYI

CoronelBuendia:
## Linked Issues

closes #1078 

## Description

Switches offset method in thermal shield to "intersect" which seems to be more robust in this case.

Adds some geometry closed checks.

## Interface Changes

N/A

## Checklist

### DIFF
--- a/bluemira/codes/_freecadapi.py
+++ b/bluemira/codes/_freecadapi.py
@@ -660,6 +660,9 @@ def discretize(w: Part.Wire, ndiscr: int = 10, dl: float = None):
     # discretization points array
     output = w.discretize(ndiscr)
     output = vector_to_numpy(output)
+
+    if w.isClosed():
+        output[-1] = output[0]
     return output
 
 
--- a/bluemira/geometry/tools.py
+++ b/bluemira/geometry/tools.py
@@ -547,6 +547,9 @@ def offset_wire(
     wire: BluemiraWire
         Offset wire
     """
+    if not open_wire:
+        wire = wire.deepcopy()
+        wire.close()
     return BluemiraWire(
         cadapi.offset_wire(wire._shape, thickness, join, open_wire), label=label
     )
--- a/bluemira/geometry/tools.py
+++ b/bluemira/geometry/tools.py
@@ -481,15 +481,22 @@ def _offset_wire_discretised(
     """
     Fallback function for discretised offsetting
 
-    Notes
-    -----
-    Can only handle closed wires
+    Raises
+    ------
+    GeometryError
+        If the wire is not closed. This function cannot handle the offet of an open
+        wire.
     """
     from bluemira.geometry._deprecated_offset import offset_clipper
 
-    if not open_wire:
+    if not wire.is_closed() and not open_wire:
         wire = wire.deepcopy()
         wire.close()
+
+    if not wire.is_closed() and open_wire:
+        raise GeometryError(
+            "Fallback function _offset_wire_discretised cannot handle open wires."
+        )
 
     coordinates = wire.discretize(byedges=byedges, ndiscr=ndiscr)
 
--- a/bluemira/geometry/tools.py
+++ b/bluemira/geometry/tools.py
@@ -487,6 +487,10 @@ def _offset_wire_discretised(
     """
     from bluemira.geometry._deprecated_offset import offset_clipper
 
+    if not open_wire:
+        wire = wire.deepcopy()
+        wire.close()
+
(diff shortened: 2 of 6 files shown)


bluemira/codes/_freecadapi.py
```diff
@@ -660,6 +660,9 @@ def discretize(w: Part.Wire, ndiscr: int = 10, dl: float = None):
     # discretization points array
     output = w.discretize(ndiscr)
     output = vector_to_numpy(output)
+
+    if w.isClosed():
```
ivanmaione, line 664: I think this "trick" should be moved into the `discretize` method of BluemiraWire. `_freecadapi` methods should return, if possible, what is expected doing the same operation in FreeCAD.

bluemira/geometry/tools.py
```diff
@@ -487,6 +487,10 @@ def _offset_wire_discretised(
     """
     from bluemira.geometry._deprecated_offset import offset_clipper
 
+    if not open_wire:
```
ivanmaione, line 490: Not sure how the `open_wire` parameter combines with the note "Can only handle closed wires". A little bit confusing for me. Probably renaming it as "force_closure" (or similar) would be more explanatory? Similar comment for the signature of `offset_wire`.
Moreover, the method should raise an error if the wire is not closed and, in this case, `open_wire` is `True`.

ivanmaione (changes requested):

bluemira/codes/_freecadapi.py
CoronelBuendia: Yeah, that makes sense - will do

bluemira/geometry/tools.py
CoronelBuendia: Agree to clean this up... I probably need to update my environment as I ran this 10 times without any failure on my machine at work. It fails first time in the CI with an error I've never seen :/

hsaunders1904:
Sorry, new error: `bluemira.geometry.error.NotClosedWire: The open boundary has not been closed.`. [Error log](https://github.com/Fusion-Power-Plant-Framework/bluemira/files/8741705/error-log2.txt).

I haven't been able to reproduce outcome 3 in https://github.com/Fusion-Power-Plant-Framework/bluemira/issues/1049 though - so that's good.

bluemira/codes/_freecadapi.py
CoronelBuendia: I note that the same trick is already used in `discretize_by_edges` in `_freecadapi.py`

ivanmaione: _youthful mistakes_ :) I still support my first comment idea, so I would propose to change also `discretize_by_edges`.  The problem probably resides in the fact that there is no counterpart in `tools.py`.

CoronelBuendia: Farmed this off to #1085 for now. Probably needs a broader treatment rather than a piecemeal approach.

ivanmaione (approved):

PR #1086, opened 2022-05-23
Loop voltage estimate, spitzer and coulomb logarithm

CoronelBuendia:
## Linked Issues

Adds some 0-D plasma physics mostly to estimate loop voltage following some cute derivations from Hartmut Zohm, with help from William Morris.

I'm not really sure yet where this should live, and its possible it won't be directly used for some time. There will be increasing amounts of the 0-D plasma physics things though, and I'm wondering whether we should put all these little tools in one place for clarity and convenience.

e.g. Closes #{ID}

## Description

## Interface Changes

## Checklist


tests/equilibria/test_physics.py
```diff
@@ -28,3 +32,30 @@ def test_psi_bd(self):
...
+        df.loc[8] = ["Hartmut's case", 10.0**20, 1000.0, 16.5]
+        cls.data = df
+
+    def test_coulomb_logarithm_values(self):
+        for i in range(9):
+            temp_in_k = self.data.loc[i, "T [eV]"] * EV_TO_J / K_BOLTZMANN
+            density = self.data.loc[i, "n [1/m^3]"]
+            value = round(coulomb_logarithm(temp_in_k, density), 1)
+            reference_value = self.data.loc[i, "ln Lambda (Goldston)"]
+            np.testing.assert_allclose(value, reference_value, rtol=0.054)
```
hsaunders1904, line 61: I would avoid asserting within a loop when possible, it can be a pain when a case fails, but you don't know which case it was. Also, is the `DataFrame` necessary here? Would it be simpler for the cases to be in a list? Suggestion below:

```python
     CASES = {
        # Case: [ n [1/m^3], T [eV], ln Lambda (Goldston) ]
        "Solar wind": [10.0**7, 10.0, 26],
        "Van Allen belts": [10.0**9, 10.0**2, 26],
        "Earth's ionosphere": [10.0**11, 10.0**-1, 14],
        "Solar corona": [10.0**13, 10.0**2, 21],
        "Gas discharge": [10.0**16, 10.0**0, 12],
        "Process plasma": [10.0**18, 10.0**2, 15],
        "Fusion experiment": [10.0**19, 10.0**3, 17],
        "Fusion reactor": [10.0**20, 10.0**4, 18],
        "Hartmut's case": [10.0**20, 1000.0, 16.5],
    }

    @pytest.mark.parametrize("case", CASES.values(), ids=CASES.keys())
    def test_coulomb_logarithm_values(self, case):
        density, temp, reference_value = case
        temp_in_k = temp * EV_TO_J / K_BOLTZMANN

        value = round(coulomb_logarithm(temp_in_k, density), 1)

        np.testing.assert_allclose(value, reference_value, rtol=0.054)
```

bluemira/equilibria/physics.py
```diff
@@ -455,3 +463,195 @@ def beta_N_to_beta(beta_N, a, Btor, Ip):  # noqa :N802
...
+    """
+    A 0-D estimate of the loop voltage during burn
+
+    Parameters
+    ----------
+    R_0: float
+        Major radius [m]
+    B_t: float
+        Toroidal field on axis [T]
+    Z_eff: float
```
hsaunders1904, line 623: We have a unit for `Z_eff` in `spitzer_conductivity()` but not here

bluemira/equilibria/physics.py
```diff
@@ -455,3 +463,195 @@ def beta_N_to_beta(beta_N, a, Btor, Ip):  # noqa :N802
...
+    v = thermal_velocity(temperature, ELECTRON_MASS)
+    lambda_de_broglie = de_broglie_length(v, mu_12)
+    b_perp = impact_parameter_perp(v, mu_12)
+    b_min = max(lambda_de_broglie, b_perp)
+    return np.log(np.sqrt(1 + (lambda_debye / b_min) ** 2))
+
+
+def spitzer_conductivity(Z_eff, T_e, ln_lambda=17):
+    """
+    Formula for conductivity as per L. Spitzer
```
hsaunders1904, line 588: Is there more detail we could give here, for those not familiar with Spitzer (me)? The conductivity through what?

From a quick google, is the below correct?

```suggestion
    Formula for electrical conductivity in a plasma as per L. Spitzer
```

hsaunders1904 (commented):
Noted that this is a WIP, but thought I'd get a few comments in early (mostly very small things)

bluemira/base/constants.py
```diff
@@ -223,6 +232,13 @@ def _transform(
     ureg.Quantity("m_e").to("g") * ureg.Quantity("avogadro_constant").to_base_units()
 ).magnitude
 
+
+# electron mass [kg]
+ELECTRON_MASS = ureg.Quantity("m_e").to("g").to_base_units().magnitude
+
+# proton mass [kg]
+PROTON_MASS = ureg.Quantity("m_p").to("g").to_base_units().magnitude
```
je-cook, line 240: On a similar vein to @hsaunders1904 just passing through. You could directly do `.to('kg')` or just `to_base_units()`.

```suggestion
# electron mass [kg]
ELECTRON_MASS = ureg.Quantity("m_e").to_base_units().magnitude

# proton mass [kg]
PROTON_MASS = ureg.Quantity("m_p").to_base_units().magnitude
```

tests/equilibria/test_physics.py
```diff
@@ -28,3 +32,30 @@ def test_psi_bd(self):
 
         assert round(abs(psib - 143), 0) == 0  # 142.66 ~ 143 V.s
         # CREATE DEMO 2015 test case
+
+
+class TestCoulombLogarithm:
+    @classmethod
+    def setup_class(cls):
+        df = pd.DataFrame(
+            columns=["Case", "n [1/m^3]", "T [eV]", "ln Lambda (Goldston)"]
```
je-cook, line 41: Possibly could use `https://github.com/hgrecco/pint-pandas` if you want units in pandas Dataframes

hsaunders1904 (approved):
LGTM 👍

CoronelBuendia:
Already a conflict in this somehow. Probably best to wait for #1112 to go in first and then reshuffle here thereafter @hsaunders1904 @je-cook

bluemira/plasma_physics/collisions.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+    return (mass_1 * mass_2) / (mass_1 + mass_2)
+
+
+def thermal_velocity(temperature, mass):
+    """
+    Parameters
+    ----------
+    temperature: float
+        Temperature [K]
```
hsaunders1904, line 81: Just realised, temperatures in public functions [should be in Celsius](https://bluemira.readthedocs.io/en/develop/conventions.html#unit-convention).

bluemira/plasma_physics/collisions.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def coulomb_logarithm(temperature, density):
+    """
+    Calculate the value of the Coulomb logarithm for an electron hitting a proton.
+
+    Parameters
+    ----------
+    temperature: float
+        Temperature [K]
```
hsaunders1904, line 138: Another Kelvin

hsaunders1904 (changes requested):
Looks fine after the rebase, just spotted some temperature units that potentially need changing.

bluemira/plasma_physics/collisions.py
CoronelBuendia: Uff.. yeah I knew I would come to regret that decision :D. It feels extremely unnatural to use Celsius in plasma physics functions like this. It's already weird we're not using [eV] or [keV]. To be honest, [K] is SI... and [C] is like an accepted derivate. Perhaps we should be simple and use [K] everywhere. It does seem a natural middle ground between plasma physics [eV] and engineering [C].

Thoughts @hsaunders1904 @je-cook @ivanmaione ?

(Incidentally I know that we could use parameters here or pint to get the conversions more naturally, but I felt that would be overkill here, and certainly inconsistent with the rest of the stuff we presently have).

hsaunders1904: Kelvin, to me, does seem the natural choice, seeing as we use SI for almost everything else. However, we'd obviously have to consider how difficult it would be to update the existing temperatures throughout the code (it could well be trivial, but may well not be).

In any case, this is a question that's out of the scope of this PR. My thinking is that we should try to be consistent in the first instance. I say get this PR merged using Celsius here for now, and we can revisit the question in a weekly meeting or at the workshop.

je-cook: I prefer Kelvin too and I think we should just stomach the changes that may need to be made. We shouldnt be hardcoding many temperatures anyway and if we are we can just use the unit converter function.

Maybe for now just stick in a 
```
raw_uc(temperature, 'celsius', 'kelvin')
```
and we can remove it later

hsaunders1904 (approved):

PR #1110, opened 2022-05-30
`Equilibrium` API improvements: move I_p, R_0, B_0 to live exclusively in `Profile`

CoronelBuendia:
## Linked Issues
Moves the float arguments `I_p` `R_0` and `B_0` to live exclusively in `Profile` simplifying the API somewhat, and avoiding repetition of input parameters which could lead to bugs and confusion. The downside is that now an Equilibrium must be specified with a profile, which makes sense anyway.

Removes `profiles` from the iterators.

@je-cook

@alexanderianblair if curious only!
## Description

## Interface Changes

## Checklist


je-cook (commented):
I think not public for `._profiles`. The original profiles object can be modified/used if required just as is done here.

I think the downside of having to specify a profile is not so much of an issue because of `CustomProfile` and examples of that are shown in the tests and examples.

If there isn't a good different reason to leave `R_0` on the breakdown I would remove it too.

Would there be an advantage of leaving `profiles` on the iterators?

CoronelBuendia:
> 



> I think not public for `._profiles`. The original profiles object can be modified/used if required just as is done here.

> I think the downside of having to specify a profile is not so much of an issue because of `CustomProfile` and examples of that are shown in the tests and examples.

OK, fair enough

 
> If there isn't a good different reason to leave `R_0` on the breakdown I would remove it too.

Yeah, there's no good reason - have removed
 
> Would there be an advantage of leaving `profiles` on the iterators?

Honestly, not right now no. In future, (iterating profiles) possibly yes. The this is that we have profiles now living more prominently on `Equilibrium` so it's a pretty grey area.

je-cook (approved):

PR #1113, opened 2022-06-03
WIP: Add PF coil position optimisation problem.

### LINKED ISSUE #529: Add PF coil position optimisation problem

The present EU-DEMO build for PF coils (see #507) only implements runmodes `mock` and `read`. We need a default `run` mode implementation, which:

* Optimises the positions and sizes of the PF coils, subject to a myriad of equilibrium and premagnetisation magnetic constraints, coil constraints.
* Generates the reference premagnetisation, start of flat-top and end-of-flatop `MHDState`s.
* With a bunch of configuration of equilibrium solver tweaks, grid size / resolution, optimiser tweaks etc.

## Proposed solution
Implement the above optimisation problem, for premagnetisation and SOF, for now.

## Alternative solutions

This was possible in BLUEPRINT, and should be possible again here. Setting things up from scratch will hopefully allow for a cleaner interface, but it would in theory be possible to resurrect the BLUEPRINT approach. This is not desirable, as it will be muckier and much less configurable.

CoronelBuendia:
## Linked Issues

Closes #529 
## Description

## Interface Changes

## Checklist


examples/equilibria/single_null.py
```diff
@@ -20,24 +20,35 @@
 # License along with bluemira; if not, see <https://www.gnu.org/licenses/>.
 
 """
-Attempt at recreating the EU-DEMO 2017 reference equilibria from a known coilset.
+Equilibrium and coilset optimisation - developer tutorial
```
CoronelBuendia, line 23: @sebkahn @markj997  - FYI example as discussed

CoronelBuendia:
I'm not over the moon with this... It works but it's messy, fiddly, and brittle. I'm starting to form the opinion that any specification of this type of problem is necessarily opinionated, and thereby truly arbitrary selection of objectives, constraints, parameterisations and the like is going to be an impossible goal that I don't want to waste too much time chasing.

It's not completely finished (still need to add the keep-out zones for EU-DEMO) but I thought it would be best to get this reviewed before making it any bigger...

@je-cook as lead, but @hsaunders1904 @ivanmaione @alexanderianblair if interested!

`examples/single_null.py` is effectively a breakdown of what is going on in the `run.py` and EU-DEMO build. That is the best point of entry for anybody interested solely in the functionality.

je-cook (changes requested):
A few comments, I'll have another look tomorrow

bluemira/equilibria/opt_constraints.py
```diff
@@ -220,12 +219,36 @@ def __init__(self, coilset, B_max, tolerance=1.0e-6):
...
 
+        x, z = self._get_constraint_points(coilset)
+
+        super().__init__(x, z, B_max, tolerance=tolerance, constraint_type="inequality")
+
+    @staticmethod
+    def _get_constraint_points(coilset):
         x, z = np.zeros(coilset.n_coils), np.zeros(coilset.n_coils)
         for i, coil in enumerate(coilset.coils.values()):
             x[i] = coil.x - coil.dx
```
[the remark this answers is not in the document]
CoronelBuendia: Not difficult, but the peak fields are invariably in the CS and they peak on the outside edge.

This is a discretised constraint with a very low resolution for speed; we just pick the point where we think the peak fields are going to be.

I'd like to write a more general zonal constraint, which takes the peak field at any point in the coil cross-section, but for this an analytical Jacobian would not be possible (I think!).

I've updated the docstring and added a note in the docstring to capture this.

bluemira/builders/EUDEMO/pf_coils.py
```diff
@@ -145,13 +212,149 @@ def reinitialise(self, params, **kwargs) -> None:
 
         self._reset_params(params)
         self._coilset = None
+        self._tf_coil_boundary = tf_coil_boundary
+        self._keep_out_zones = keep_out_zones
 
-    def run(self, *args):
+    def run(self):
```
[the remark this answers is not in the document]
CoronelBuendia: Aye, me neither! It works for our specific case right now. I think it's probably best to wait for the architecture refactor before fiddling with this to make it more configurable.

bluemira/equilibria/run.py
```diff
@@ -254,108 +417,218 @@ def __init__(
...
+        current_opt_constraints: Optional[List[OptimisationConstraint]],
+        coil_constraints: Optional[List[OptimisationConstraint]],
+        equilibrium_constraints: List[OptimisationConstraint],
+        profiles: Profile,
+        breakdown_strategy_cls: Type[BreakdownZoneStrategy],
+        breakdown_problem_cls: Type[BreakdownCOP],
+        breakdown_optimiser: Optimiser = Optimiser(
+            "COBYLA", opt_conditions={"max_eval": 5000, "ftol_rel": 1e-10}
+        ),
+        breakdown_settings: Dict = {"B_stray_con_tol": 1e-8, "n_B_stray_points": 20},
```
je-cook, line 512: This will run into the mutable defaults issue. Possibly we should have some defaults baked into each problem class and set this to None. The dictionaries are fine in the `breakdown_optimser` because they are wrapped in the `Optimiser` object

bluemira/equilibria/run.py
```diff
@@ -254,108 +417,218 @@ def __init__(
...
-                eq.coilset,
-                eq,
-                optimiser,
-                max_currents=max_currents,
-                constraints=constraints,
+        # Re-run breakdown
+        psi_bd_orig = self._psi_premag
+        self.coilset = optimised_coilset
+        self.run_premagnetisation()
+        if not self._psi_premag >= psi_bd_orig - 2.0:
```
je-cook, line 598: less confusing?
```suggestion
        if self._psi_premag < psi_bd_orig - 2.0:
```

je-cook (changes requested):
Last couple then I think I'm done

CoronelBuendia:
Uff the rebase on this one is complicated with the move of the EU-DEMO folder... at least for me! @je-cook would you mind giving it a shot? I'll pick up your comments thereafter.

je-cook:
Done, let me know if there are problems. I'll look at test failures if there are any but diff didn't look too huge. I think it all came from an extra couple of imports from some moved builder modules.

CoronelBuendia:
> Done, let me know if there are problems. I'll look at test failures if there are any but diff didn't look too huge. I think it all came from an extra couple of imports from some moved builder modules.

Brilliant thanks! Sorry, I took one look at it and baulked :D

je-cook (approved):

PR #1128, opened 2022-06-09
Spherical Harmonics

markj997:
## Description

Implementation of necessary functions, constraints, examples, etc, enabling use of spherical harmonics for optimisation and solving of equilibria.

### DIFF
--- a/bluemira/equilibria/flux_surfaces.py
+++ b/bluemira/equilibria/flux_surfaces.py
@@ -32,7 +32,9 @@ from typing import Iterable
 import matplotlib.pyplot as plt
 import numba as nb
 import numpy as np
+import scipy.constants as const
 from scipy.integrate import solve_ivp
+from scipy.special import lpmv
 
 from bluemira.base.look_and_feel import bluemira_warn
 from bluemira.equilibria.constants import PSI_NORM_TOL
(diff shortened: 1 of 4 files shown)


bluemira/equilibria/flux_surfaces.py
```diff
@@ -871,3 +873,102 @@ def poloidal_angle(Bp_strike, Bt_strike, gamma):
...
+            * const.mu_0
+            * (r_t / r_f) ** degree
+            * np.sin(theta_f)
+            * lpmv(1, degree, np.cos(theta_f))
+            / np.sqrt(degree * (degree + 1))
+        )
+
+    amplitudes = np.matmul(currents2harmonics, I_f)
+
+    return amplitudes
```
je-cook, line 919: Just fyi the '@' symbol I think accomplished the same thing (uses matmul under the hood) also you could just directly return the calculation eg 
```suggestion
    return currents2harmonics @ I_f
```

bluemira/equilibria/opt_constraints.py
```diff
@@ -61,6 +62,44 @@ def _get_dummy_equilibrium(equilibrium):
...
+    max_degree: float
+        Maximum degree of spherical harmonics desired to constrain.
+    """
+    # Does the eq get updated at each iteration, requiring I_f as input or can
+    # we just extract I_f from the eq object within the function?
+
+    # Could look at a way of adding harmonics until a desired precision in core
+    # plasma is reached.
+    for coil in eq.coilset:
+        x_f, z_f = coil.x, coil.z
```
je-cook, line 89: Is this not just getting the x and z of the last coil in the coilset?
```python
coil = eq.coilset[-1]
x_f, z_f = coil.x, coil.z
```

je-cook (changes requested):
Looks good! Just some code styling things mostly, also one about where the loop should start

bluemira/equilibria/opt_constraints.py
```diff
@@ -61,6 +62,44 @@ def _get_dummy_equilibrium(equilibrium):
...
+        Equilibrium used to for coilset.
+    r_t: float
+        Typical length scale of the problem (e.g. radius at outer midplane)
+    initial_harmonics: np.ndarry
+        Initial harmonic amplitudes obtained from desired core plasma
+    max_degree: float
+        Maximum degree of spherical harmonics desired to constrain.
+    """
+    # Does the eq get updated at each iteration, requiring I_f as input or can
+    # we just extract I_f from the eq object within the function?
```
sebkahn, line 84: The equilibrium gets update at each grad-Shafranov iteration by definition.

As far as I understand the GS does not get updated with the PF optimization, however the coilset by definition should. Although I am not fully aware of the implentation details @alexanderianblair ?

sebkahn (changes requested):

bluemira/equilibria/flux_surfaces.py
```diff
@@ -32,7 +32,9 @@
 import matplotlib.pyplot as plt
 import numba as nb
 import numpy as np
+import scipy.constants as const
```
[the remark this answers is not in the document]
je-cook: bluemira ones please. They all come from the same place which is the 2018 codata database. We get that via Pint just so we get them all from the same package

bluemira/equilibria/opt_constraints.py
je-cook: Cant remember, but I can make it so in the refactor. For now your suggestion is probably the right way forward

bluemira/equilibria/flux_surfaces.py
```diff
@@ -871,3 +873,102 @@ def poloidal_angle(Bp_strike, Bt_strike, gamma):
     sin_theta = num / Bp_strike
 
     return np.rad2deg(np.arcsin(sin_theta))
+
+
+def harmonic_amplitude(x_f, z_f, I_f, max_degree, r_t, cylindrical=False):
+    """
+
```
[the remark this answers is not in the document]
je-cook: https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-math :slightly_smiling_face:

sebkahn:
Good first pass, we should put the constraints on `opt_constraitns_funcs.py`

CoronelBuendia:
Closing in favour of #1963

PR #1145, opened 2022-06-13
✨ Save cad of a component, reactor or list of components

je-cook:
## Description

This allows for easy saving of CAD geometries. Its mostly recycled code and needs a test or two

## Interface Changes

None

## Checklist


bluemira/base/components.py
```diff
@@ -190,6 +190,23 @@ def prune_child(self, name: str):
...
+        options = []
+        if self.is_leaf and isinstance(self, PhysicalComponent):
+            shapes.append(self.shape)
+            options.append(self.display_cad_options)
+        else:
+            for child in self.children:
+                shape, option = child._get_shapes()
+                shapes.extend(shape)
+                options.extend(option)
+
```
sebkahn, line 207: 1. The logic of getting a flat list of all the `PhysicalComponents` down the tree is a useful one for potentially other purposes. Can we put this in a separated function?

2. Is it possible to have a component without `PhysicalComponent` down the line ? If yes this function may error out or be stuck in a infinite loop?

je-cook:
Draft because awaiting #1157

bluemira/geometry/tools.py
```diff
@@ -871,6 +872,22 @@ def save_as_STEP(shapes, filename="test", scale=1):
     cadapi.save_as_STEP(freecad_shapes, filename, scale)
 
 
+def save_cad(components, filename="test", scale=1):
```
hsaunders1904, line 875: Is there a requirement to have a default file name here? Usually, '`save`' functions require a path to be specified. I foresee some accidental file overwrites with a default.

hsaunders1904 (commented):
I see a lot of this is in #1157. Just one comment for the other code

bluemira/geometry/tools.py
je-cook: Year fair enough, it was really just copied from `save_as_STEP` i'll remove the default from both

hsaunders1904:
> If you want me to add a `save_as_STEP` test let me know. I think thats the only possibly needed test

Yes, please. A regression test from a STEP file should hopefully be simple, but would add a lot of value.

je-cook:
That test was more annoying than expected. I've also excluded `test_data` folders from some of our precommit stuff

bluemira/geometry/tools.py
```diff
@@ -859,21 +860,29 @@ def circular_pattern(
...
-    if not filename.endswith(".STP"):
-        filename += ".STP"
+    if not filename.endswith(
+        tuple(
+            getattr(fe, func)()
+            for func in ("upper", "lower")
+            for fe in (".stp", ".step")
+        )
+    ):
+        filename += ".stp"
```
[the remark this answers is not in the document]
je-cook: Done

bluemira/codes/_freecadapi.py
```diff
@@ -963,8 +963,14 @@ def save_as_STEP(shapes, filename="test", scale=1):
...
     """
-    if not filename.endswith(".STP"):
-        filename += ".STP"
+    if not filename.endswith(
+        tuple(
+            getattr(fe, func)()
+            for func in ("upper", "lower")
+            for fe in (".stp", ".step")
+        )
+    ):
```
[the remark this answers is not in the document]
je-cook: Used your below suggestion

je-cook:
@CoronelBuendia can you take a look at this for the geometry reviewers? Its not very geometry focused either way

hsaunders1904 (approved):

CoronelBuendia (approved):
LGTM. Unrelated side note that we might want to rename `save_as_STEP` to `save_as_STP` or similar in future, just to avoid any confusion.

je-cook:
> LGTM. Unrelated side note that we might want to rename `save_as_STEP` to `save_as_STP` or similar in future, just to avoid any confusion.

Done

PR #1157, opened 2022-06-20
Generalise searching through a component

je-cook:
## Linked Issues

Spurred by comments in #1145 

## Description

Generalise component search instead of homebrewed version. This really is a split out from #1145 because it was a bit out of scope to add a property search for a cad saver addition

## Interface Changes

New search for property function

## Checklist

### DIFF
--- a/bluemira/base/components.py
+++ b/bluemira/base/components.py
@@ -28,7 +28,6 @@ from __future__ import annotations
 from typing import TYPE_CHECKING, Any, Callable, Iterable, List, Optional, Union
 
 import anytree
-import numpy as np
 from anytree import NodeMixin, RenderTree
 
 from bluemira.base.error import ComponentError
@@ -156,16 +155,11 @@ class Component(NodeMixin, Plottable, DisplayableCAD):
                 return getattr(found_nodes, properties[0])
             return [getattr(found_nodes, prop) for prop in properties]
         else:
-            return tuple(
-                obj.tolist()
-                for obj in np.array(
-                    [
-                        [getattr(node, prop) for prop in properties]
-                        for node in found_nodes
-                    ],
-                    dtype=object,
-                ).T
-            )
+            # Collect values by property instead of by node
+            node_properties = [
+                [getattr(node, prop) for prop in properties] for node in found_nodes
+            ]
+            return tuple(map(list, zip(*node_properties)))
 
     def _get_thing(self, filter_: Callable, first: bool, full_tree: bool):
         found_nodes = anytree.search.findall(
--- a/bluemira/base/components.py
+++ b/bluemira/base/components.py
@@ -312,7 +312,7 @@ def get_properties_from_components(
     comps: Union[Component, Iterable[Component]], properties: Union[str, Sized[str]]
 ):
     """
-    Get internal shapes and display options from Components
+    Get properties from Components
 
     Parameters
     ----------
--- a/bluemira/base/components.py
+++ b/bluemira/base/components.py
@@ -108,15 +108,15 @@ class Component(NodeMixin, Plottable, DisplayableCAD):
 
         Notes
         -----
-            This function is just a wrapper of the anytree.search.findall or find
-            functions.
+            This function is just a wrapper of the anytree.search.findall
+            function.
         """
         return self._get_thing(
             lambda n: anytree.search._filter_by_name(n, "name", name), first, full_tree
         )
 
     def get_component_properties(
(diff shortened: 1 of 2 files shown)


bluemira/base/components.py
```diff
@@ -107,21 +108,74 @@ def get_component(
...
-            This function is just a wrapper of the anytree.search.findall_by_attr
+            This function is just a wrapper of the anytree.search.findall
             function.
         """
-        if full_tree:
-            found_components = anytree.search.findall_by_attr(self.root, name)
-        else:
-            found_components = anytree.search.findall_by_attr(self, name)
+        return self._get_thing(
+            lambda n: anytree.search._filter_by_name(n, "name", name), first, full_tree
```
je-cook, line 115: I know this is using a private function but it seems a bit ott to rewrite what is a 3 line function that is used as the filter in `findall_by_attr` anyway

bluemira/base/components.py
```diff
@@ -107,21 +108,74 @@ def get_component(
...
+            return tuple(
+                obj.tolist()
+                for obj in np.array(
+                    [
+                        [getattr(node, prop) for prop in properties]
+                        for node in found_nodes
+                    ],
+                    dtype=object,
+                ).T
+            )
```
je-cook, line 168: If you have a nicer way to 'rotate' a list of lists I'm open to it :)

bluemira/base/components.py
```diff
@@ -107,21 +108,74 @@ def get_component(
...
+        found_nodes = self._get_thing(lambda n: filter_(n, properties), first, full_tree)
+
+        if found_nodes is None:
+            return tuple([] for _ in properties)
 
-        if first:
-            return found_components[0]
+        if not isinstance(found_nodes, Iterable):
+            if len(properties) == 1:
+                return getattr(found_nodes, properties[0])
```
hsaunders1904, line 156: Is this returning a list of properties as described in the docstring?

bluemira/base/components.py
```diff
@@ -107,21 +108,74 @@ def get_component(
...
+            The name of the component to search for.
+        first: bool
+            If True, only the first element is returned, by default True.
+        full_tree: bool
+            If True, searches the tree from the root, else searches from this node, by
+            default False.
+
+        Returns
+        -------
+        Lists of properties of components
```
hsaunders1904, line 136: We could be clearer here that this returns a list if `first`, but a tuple of lists if `not first`. Or am I misunderstanding?

bluemira/base/components.py
```diff
@@ -107,21 +108,74 @@ def get_component(
...
-        else:
-            found_components = anytree.search.findall_by_attr(self, name)
+        return self._get_thing(
+            lambda n: anytree.search._filter_by_name(n, "name", name), first, full_tree
+        )
 
-        if len(found_components) == 0:
-            return None
+    def get_component_properties(
+        self, properties: Union[Iterable[str], str], first=True, full_tree=False
```
hsaunders1904, line 119: Not all `Iterable`s have `__len__`. I think we want `Sized` here

bluemira/base/components.py
hsaunders1904: How about:

```python
node_properties = [
    [getattr(node, prop) for prop in properties] for node in found_nodes
]
return tuple(map(list, zip(*node_properties)))
```

Either way, I think this is a bit cryptic. Probably warrants a comment noting that we're collecting similar _property_ values, instead of ordering by _node_ property values

bluemira/base/components.py
```diff
@@ -258,3 +312,46 @@ def conductor(self):
     @conductor.setter
     def conductor(self, value):
         self._conductor = value
+
+
+def get_properties_from_components(
```
hsaunders1904, line 317: This looks very useful

hsaunders1904 (changes requested):
These should be useful changes. Just a couple of things

bluemira/base/components.py
hsaunders1904: I guess I'm OK with this, `get_component` is pretty well tested, so we should get an early warning if this API changes or disappears.

bluemira/base/components.py
je-cook: Thats nicer :) Sure I'll add both

bluemira/base/components.py
je-cook: Update, hopefully thats clearer

PR #1179, opened 2022-06-24
Refactor `PlasmaCoil` for speed and logic

### LINKED ISSUE #1149: `Equilibrium` `PlasmaCoil` speed and logic improvements

When optimising PF coil positions we tend to "freeze" the plasma contributions which is presently done via a `PlasmaCoil` object that behaves like a `Coil` with a fixed current (albeit with a variable current density).

This is quite slow to instantiate, and what's more is quite slow to call (or rather slower than it could be).

## Proposed solution

Cache plasma coil (coming in anyway with #1113) but also for points on the grid make use of psi, Bx, and Bz interpolation rather than convolving Greens' functions. For points off the grid, convolve greens functions (if at all necessary) with stashed response matrices.

CoronelBuendia:
## Linked Issues

Closes #1149 

## Description
Refactor of `PlasmaCoil` to only use green's functions for points outside a `Grid`.

Previously, these were being used for all points in order to reconstruct the plasma psi and Bx, Bz contributions from the toroidal current density. This is not necessary for points on the grid as we technically have direct access to them.

Not sure why I had it like this!

## Interface Changes

`PlasmaCoil` now imported from `equilibria/plasma.py`

`Equilibrium` simplified so that anything related to the plasma lives in its `.plasma` attribute.

Still need to address the `plasma_coil` method on `Equilibrium`, but wasn't sure if this would go in first or #1113.

## Checklist


je-cook:
Looks good, I'll try and get this and #1113 reviewed asap

bluemira/equilibria/plasma.py
```diff
@@ -0,0 +1,257 @@
...
+
+        x = np.array(x)
+        z = np.array(z)
+        if x.shape != z.shape:
+            raise EquilibriaError("x and z arrays of different dimension.")
+
+        values = np.zeros(x.shape)
+        values = values.flatten()
+
+        for i, (xx, zz) in enumerate(zip(x.flatten(), z.flatten())):
```
je-cook, line 54: https://stackoverflow.com/questions/30523123/when-to-use-flat-flatiter-or-flatten also ravel is an option (there are defintiely too many options...)
```suggestion
        for i, (xx, zz) in enumerate(zip(x.flat, z.flat)):
```

je-cook (approved):
2 minors but otherwise looks good. Conceptually I prefer the separation from CoilGroup related logic

DarioV86 (approved):

PR #1180, opened 2022-06-24
Add warnings to CI and run more tests for the develop dependencies PRs

### LINKED ISSUE #1174: Add warnings and run more tests for the develop_dependencies merge PR

It would be nice to both run the longrun tests and print any warnings in a comment when merging in `develop_dependencies` so we don't missing warnings in the CI

## Proposed solution

Add `pytest-json-report` to our pytest plugins and run:

`pytest --json-report --json-report-indent=3 --reactor --longrun --private`

the following couple of commands will filter the warnings and print the number we have from the generated report:
`jq '.warnings | length' .report.json`
`jq '.warnings | unique' .report.json`

There is possibly a nice way to format that in a comment but code quotes would also be fine.

I think it would be possible to just modify our existing `ci.yml` eg naiively with an extra environment variable:
```
env:
     PYEST_LONGRUN: ${{ github.ref_name == "develop_dependencies" && "--longrun" || "" }}
```

## Alternative solutions

Just add the `--longrun` tests for the dependencies PR

hsaunders1904:
## Linked Issues

Closes #1174 

## Description

Add steps to the CI to notify us if there are warnings in our PRs. These notifications take the form of PR comments that list the warnings that pytest has collected.

It also adds a step when the CI is run on "push", i.e. it's run on the develop/main branch, to generate a 'baseline' report. On PRs, we can then use the baseline report and only notify about _new_ warnings. Hopefully this makes managing new warnings more straight-forward.

This PR also adds the `--longrun` flag to the pytest call if `develop_dependencies` is the _head_ branch of PR. 

## Checklist

### DIFF
--- a/scripts/format_warning_report.py
+++ b/scripts/format_warning_report.py
@@ -112,7 +112,7 @@ def format_warning_report(sys_args: List[str]) -> int:
     if inputs.compare is not None:
         ref_warnings = load_warnings(inputs.compare)
         new, fixed = compare_warnings(warnings, ref_warnings)
-        tada = " 🎉" if len(new) == 0 and len(fixed) >= 0 else ""
+        tada = " 🎉" if len(new) == 0 else ""
         report.append(
             f"Found {len(new)} new warning{'' if len(new) == 1 else 's'}, "
             f"{len(fixed)} fixed warning{'' if len(fixed) == 1 else 's'}.{tada}\n"
--- a/scripts/format_warning_report.py
+++ b/scripts/format_warning_report.py
@@ -1,4 +1,15 @@
 #!/usr/bin/env python3
+"""
+Take a '.report.json' output from pytest-json-report, and generate a
+markdown formatted report of any warnings generated.
+
+The script can also compare one JSON report to another, 'baseline',
+report and generate a markdown report that lists new warnings.
+
+In compare mode, the exit code is the number of new warnings. In normal
+mode, the exit code is the number of warnings. A negative exit code
+indicates an error.
+"""
 
 import argparse
 import json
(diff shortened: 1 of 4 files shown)


scripts/format_warning_report.py
```diff
@@ -0,0 +1,149 @@
...
+            whens[warning.when].append(warning)
+        except KeyError:
+            whens[warning.when] = [warning]
+    lines = []
+    for when, warns in whens.items():
+        lines.append(f"#### On {when}\n")
+        for warn in warns:
+            lines.append(f"- `{warn}`")
+        lines.append("")
+    return lines[:-1]  # we do not need a trailing new line
```
je-cook, line 76: This is probably a bit more robust? Don't want to remove too much
```suggestion
    return lines.strip("\n")  # we do not need a trailing new line
```

je-cook (approved):
Looks good to me, and I've seen it working on your fork. Just one very minor

scripts/format_warning_report.py
hsaunders1904: I'm building up the lines as a list, so we can't use `strip` here. I think the slicing is robust in this case, as the last line to be appended in the loop is always empty.

je-cook:
This might be a way around. You can list all artefacts of a repository with the GH cli:

https://docs.github.com/en/rest/actions/artifacts#list-artifacts-for-a-repository=

hsaunders1904:
```sh
artifact_url="$(gh api repos/Fusion-Power-Plant-Framework/bluemira/actions/artifacts --jq '.artifacts[] | select(.name == "report-json") | select(.workflow_run.head_sha == "${{ github.event.pull_request.base.ref }}") | .archive_download_url')"
gh api ${artifact_url} > report-json.zip
```

PR #1194, opened 2022-06-30
ADR for `ParameterFrame`

CoronelBuendia:
## Linked Issues

Does this PR close or fix any Issues? Remember to create an Issue before starting work so that your fix / proposal can be addressed by the team.

e.g. Closes #{ID}

## Description

What is your PR trying to achieve? How did you go about achieving it?

## Interface Changes

If you've had to update an interface or introduce a new interface as part of your change then let us know here.

## Checklist

I confirm that I have completed the following checks:

- [ ] Tests run locally and pass `pytest tests --reactor`
- [ ] Code quality checks run locally and pass `flake8` and `black .`
- [ ] Documentation built locally and checked `sphinx-build -W documentation/source documentation/build`

PR #1192, opened 2022-06-30
ADR for `DesignStage`

CoronelBuendia:
## Linked Issues

Does this PR close or fix any Issues? Remember to create an Issue before starting work so that your fix / proposal can be addressed by the team.

e.g. Closes #{ID}

## Description

What is your PR trying to achieve? How did you go about achieving it?

## Interface Changes

If you've had to update an interface or introduce a new interface as part of your change then let us know here.

## Checklist

I confirm that I have completed the following checks:

- [ ] Tests run locally and pass `pytest tests --reactor`
- [ ] Code quality checks run locally and pass `flake8` and `black .`
- [ ] Documentation built locally and checked `sphinx-build -W documentation/source documentation/build`

PR #1230, opened 2022-07-08
Move parameters

je-cook:
## Linked Issues

Part of #1192 and #1194 

## Description

- Moves non global parameters out of the global config
- Creates an unsed_config just for storage of unused parameters
- Allows a reactor to be made with its own global config
- Creates EUDEMO Configuration
- Fixes some missing params, config schema, and oddly defined variables.

I havent dealt with fuelcycle yet but I think that could wait until later in this process. I have also alphabetised the configs so I would start looking at the diff after the first 2 (I think) commits.

This is very much a stepping stone and not a final state

## Interface Changes

New `configuration` class attr for reading a specific config other than the default

## Checklist

### DIFF
--- a/bluemira/base/design.py
+++ b/bluemira/base/design.py
@@ -74,6 +74,11 @@ class DesignABC(abc.ABC):
     def configuration(self) -> Type[Configuration]:
         """
         Configuration
+
+        Reactor global configuration
+
+        currently all parameters of a reactor will be stripped down to just the global
+        ones
         """
         pass
 


bluemira/base/design.py
```diff
@@ -285,6 +292,7 @@ class Design(DesignABC):
     _params: Configuration
     _build_config: Dict[str, BuildConfig]
     _builders: Dict[str, Builder]
+    configuration: Configuration = Configuration
```
hsaunders1904, line 295: Should this be one of the following?

```python
configuration: Configuration = Configuration()
```
```python
configuration: Type[Configuration] = Configuration
```

bluemira/base/design.py
```diff
@@ -67,9 +67,16 @@ def __init__(
...
         self._extract_build_config(params)
         self._validate_params(params)
-        self._params = Configuration.from_template(self._required_params)
+        self._params = self.configuration.from_template(self._required_params)
         self._params.update_kw_parameters(params)
 
+    @abc.abstractproperty
+    def configuration(self) -> type(Configuration):
+        """
+        Configuration
```
hsaunders1904, line 76: I realise this may only be here temporarily, but it may help in refactoring later if we spelled the purpose of this configuration out a little. It's not immediately obvious to me.

Is it all parameters for the `Design`? Or the local? Or global?

bluemira/base/ununsed_config.py
```diff
@@ -0,0 +1,323 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with bluemira; if not, see <https://www.gnu.org/licenses/>.
+
+"""
+Unused params
+"""
+Parameter = None
```
hsaunders1904, line 25: Is there something preventing us importing `Parameter` here? Using `None` as a type feels a bit odd

examples/design/EU-DEMO/EUDEMO_reactor.py
```diff
@@ -154,11 +154,11 @@
...
     "tk_bb_ib": 0.755,
     "tk_bb_ob": 1.275,
-    "g_tf_pf": 0.05,
+    # "g_tf_pf": 0.05,
     "C_Ejima": 0.3,
     "eta_nb": 0.4,
-    "LPangle": -15,
-    "w_g_support": 1.5,
+    # "LPangle": -15,
+    # "w_g_support": 1.5,
```
hsaunders1904, line 161: Can these be removed? Feels like we're adding unnecessary tech debt in needing to remember to uncomment/remove these later

hsaunders1904 (changes requested):
Looks a good job on a tedious task! A couple of comments below.

bluemira/base/design.py
hsaunders1904: Also, I think this should be `Type[Configuration]`

bluemira/base/design.py
hsaunders1904: Actually, since this is an abstract property, should we be assigning it a value?

bluemira/base/ununsed_config.py
je-cook: That was me being lazy :smile:

bluemira/base/design.py
je-cook: It should be the second. 

In Design possibly not, but I was try to preserve existing functionality

CoronelBuendia:
@je-cook is this still relevant?

je-cook:
Other than for reference no and we can always go back and look

Issue #1247, opened 2022-07-12
Develop Power Cycle production sub-module


tplobo:
## Description of issue / requirement to address
The Power Cycle module should include a sub-routine to compute the power loads associated with the Balance-of-Plant system of the power plant, to be passed to a Power Cycle manager class.

## Proposed solution
Implement a sub-module named Balance-of-Plant, that creates all associated power loads.

je-cook:
Completed in #2873

Issue #1246, opened 2022-07-12
Develop Power Cycle consumption sub-module


tplobo:
## Description of issue / requirement to address
The Power Cycle module should include a sub-routine to sum all power consumption contributions in the power plant, to be passed to a Power Cycle manager class.

## Proposed solution
Implement a sub-module named Recirculating Power Loads, that creates all power consumption power loads.

je-cook:
Completed in #2873

PR #1250, opened 2022-07-12
Tplobo/powercycle base

### LINKED ISSUE #1243: Develop classes to build a pulse timeline for the Power Cycle
The Power Cycle aims at simulating a complete reactor pulse. Classes should be created to represent a generic scenario, that can used to organize the Power Cycle simulation in different pulses, each composed of different phases.

## Proposed solution
Create 3 classes to organize :
- `PowerCyclePhase` (shortest temporal period with unified characteristics for Power Cycle simulations)
- `PowerCyclePulse` (collection of phases)
- `PowerCycleScenario` (collection of pulses)
- `ScenarioBuilder` (creates a `PowerCycleScenario` instance from a `json` input file)

## Additional context
Phases will be classified as 'steady-state' or 'transient' in a subclass (namely, `BOPPhase`), for later implementation of a time-dependent Balance-of-Plant model for the Power Cycle.

### LINKED ISSUE #1244: Develop classes to represent power loads produced/consumed in the Power Cycle
The main calculations performed by the Power Cycle should be done in terms of time-dependent power loads. These power loads should be constructed from simple power requirement information and expected time-dependent behavior (e.g. 'ramp' or 'step').

## Proposed solution
Development of 4 classes to build power loads to be used in the Power Cycle module:
- `LoadData` (building blocks for `PowerLoad`, validates basic inputs before building them)
- `PowerLoad` (computes a time-dependent power curve based on expected behavior -- an interpolation model)
- `PhaseLoad` (set of power loads that represent the power consumption/production during a given phase, which can be normalized in time or not)
- `PulseLoad` (set of phase loads that are shifted in time to build a power curve during a pulse)

## Additional context
Phase and Pulse specifications dependent on timeline classes (#1243).

### LINKED ISSUE #1245: Develop base classes and tools for Power Cycle module
Many classes in the Power Cycle module present some common characteristics that can be shared by Abstract Base Classes (ABC).
Functions developed for convenience that are used by more than one class should be stand-alone for later re-location in bluemira.

## Proposed solution
Implement a `base` script that includes ABC classes for the Power Cycle module.
Implement a `tools` script with stand-alone functions to be used as convenience tools.

### LINKED ISSUE #1248: Develop a Power Cycle manager class to compute net electricity
To compute net electricity, the Power Cycle module should have a high-level manager class that creates power loads based on run inputs defined by `json` files. This manager class should taken into consideration 3 types of plant loads: active (consumption) loads, reactive (consumption) loads and production loads.
The class should build a net pulse load by subtracting active load from production loads.

## Proposed solution
Create a `PowerCycleManager` class.

tplobo:
# Linked Issues

Classes in the `time` file close #1243 
Classes in the `net/loads` file close #1244 
Classes in files `tools` and `base` close #1245 
Classes in the `net/importers` file mock classes to import data computed by other BLUEMIRA modules (to structure a solution for #1246 & #1247)
Classes in the `net/manager` file close #1248

---
# Description
This pool request is the first contribution for a time-dependent Power Cycle module in BLUEMIRA. It adds the necessary classes to represent time-dependent power loads in the power plant.
A simple example is included in `examples/power_cycle/simple_example` to introduce users to its functionalities.
Revision of files is advised in the order presented below.

## Tool functions & classes (`tools` file, 295 LOC)
Useful functions are divided in 3 major categories, but have not been added to custom classes to enable easier reintegration in other BLUEMIRA modules. Tool classes are a 4th category of tools, and are used for input standardization/validation. These categories are listed below:

1. Validation functions, to ensure objects of certain classes have a specific format for usage in the Power Cycle module.

2. Manipulation functions, to create, copy or modify instances of base classes, such as `list`, `str` or `dict`.

3. Plotting functions, to call or modify methods of the `matplotlib` package.

4. Classes that inherit from `dict`, to organize data in pre-determined formats.

## Base classes (`base` file, 212 LOC)
All abstract base classes for definitions in the Power Cycle module:

1. `PowerCycleABC`, almost all classes in the Power Cycle module inherit from this ABC, which defines the need of instances for a 'name' and 'label'. This should be later compared and possibly standardized with the 'name' and 'long_name' parameters in other BLUEMIRA modules. Equality checks ('__eq__') between instances of children of this class ignore name and label in the comparison. Other comparisons could be implemented in the future.

2. `PowerCycleTimeABC`, ABC class for classes defined in `time`.

3. `PowerCycleLoadABC`, ABC class for classes defined in `net/loads`.
 
5. `PowerCycleImporterABC`, ABC class for classes defined in `net/importers`.

## Time classes (`time` file, 280 LOC)
In the Power Cycle module, time is structured in three main classifications: phases, pulses and scenarios. Thus, the file `time` contains the following classes:

1. `PowerCyclePhase`, instances of which are the minimal representation of time. An instance is built with a `duration_breakdown`, a dictionary in which each item identifies a time period (keys are `str` that describe the item, values are numerical). The total phase duration is the sum of breakdown elements.

2. `PowerCyclePulse`, instances of which are the minimal time representation of a scenario. An instance is built with the `phase_set` parameter, a list of PowerCyclePhase` objects in the desired order of occurrence.

3. `PowerCycleScenario`, instances of which are the largest-scale representation of plant time. An instance is built with the `pulse_set` parameter, a list of `PowerCyclePulse` objects in the desired order of occurrence. Currently in simplified implementation; in the future a `repetition` parameter should be added to specify the number of repetitions for each pulse, so that `pulse_set` does not repeat information.

4. `ScenarioBuilder`, which reads a JSON file with inputs for building a scenario, and builds a `PowerCycleScenario` object with those inputs. This class does not require naming, so it does not inherit from `PowerCycleABC`.

## Load classes (`net/loads` file, 548 LOC)
In the Power Cycle module, creation of and operations between time-dependent power curves are performed with instances of the following classes, all of which have their own particular `plot` implementation:

1. `LoadData`,  instances of which are the minimal information necessary to represent a time-dependent power signal. Each value in a `time` list is associated with a value in a `data` list. The time-characterization may be modified by "stretching" (with `_normalize_time`) or "displacement" (with `_shift_time`). No dependence is defined for values between elements of `data`.

2. `LoadModel`,  a child of the `Enum` class. Instances of this class specify which type of interpolation, computed by the `scipy.interpolate.interp1d` method, should be applied to compute values between elements of the `data` attribute of a `LoadData` object. Instances of this class are not applied directly to `LoadData`, but instead as parameters of the `PowerLoad` class.

3. `PowerLoad`, instances of which build a simple power signal with lists of `LoadData` and `LoadModel` objects. An instance can sum the contributions of each `LoadData` by interpolating values using its respective `LoadModel` specification, and can plot the total curve in time. Instances can be added to one another, and multiplied/divided by scalar values.

4. `PhaseLoad`, instances of which specify in which time period a load should occur, built with lists of `PowerLoad` and `bool` objects. The boolean list acts as a series of "normalization" flags and a `PowerCyclePhase` object passed as parameter is used to specify the total load duration; `True` flags indicate that the total time length of a `PowerLoad` should be "stretched/compressed" to fit the phase duration. Instances can be added to one another, only if representing the same phase.

6. `PulseLoad`, instances of which build the total load during a pulse using, using a list of `PhaseLoad` objects. This list acts as a library of loads for each phase of a `PowerCyclePulse` object parameter. Multiple `PhaseLoad` objects for the same phase are added. Each of these loads is displaced in time, in order to occur after the end of the last phase. Plotting includes delimiter lines at the time that each phase ends.

7. `ScenarioLoad`, instances of which build the total load during a scenario. Currently not implemented.

## Importer classes (`net/importer` file, 65 LOC)
Static classes used to represent the future implementation of data transfer from other BLUEMIRA modules into the Power Cycle module. These classes does not require naming, so they do not inherit from `PowerCycleABC`.

1. `EquilibriaImporter`, static class that will import data from the `equilibria` module. 

2. `PumpingImporter`, static class that will import data from the module that will compute "tokamak pumping" data.

## Manager classes (`net/manager` file, 320 LOC)
Classes used to build a time evolution of the power load produced and consumed in the power plant, using data from other modules and a series of input files.

1. `PowerCycleSystem`, instances of which build `PhaseLoad` objects to represent active and reactive power being produced/consumed by the power plant.

2. `PowerCycleGroup`, instances of which are used to group plant systems according to their systemic classification.

3. `PowerCycleManager`, which calls `ScenarioBuilder` and reads a JSON file with inputs for all the power loads associated with that scenario. It builds Power Cycle groups and systems, and merges their PhaseLoads in one `PulseLoad` per load type (one for "active", one for "reactive"). This class does not require naming, so it does not inherit from `PowerCycleABC`.

---
# Interface Changes
None.

---
# Checklist
I confirm that I have completed the following general checks:

- [x] Code quality checks run locally and pass `flake8` and `black` .

I confirm that I have completed the following checks particular to this module:
- [X] Tests run locally and pass `pytest tests/power_cycle/`


bluemira/power_cycle/base.py
```diff
@@ -0,0 +1,468 @@
...
+    def __init__(self, name: str):
+
+        # Store name
+        self.name = self._validate_name(name)
+
+        # Dynamically add validation classmethod
+        # setattr(a, 'func', classmethod(func))
+
+    @classmethod
+    def _validate_name(cls, name):
```
je-cook, line 73: This is not a classmethod you can get what you want from `self.__class__.__name__`.

More generally this seems like overkill on the validation it seems odd that you are separating the errors from the location they need to be raised. I would remove this

bluemira/power_cycle/base.py
```diff
@@ -0,0 +1,468 @@
...
+                f"""
+                The 'name' used to create an instance of the
+                {class_name} class must be a `str`.
+                """
+            )
+
+        return name
+
+    @classmethod
+    def _issue_error(cls, label):
```
je-cook, line 89: Same here on both counts

bluemira/power_cycle/base.py
```diff
@@ -0,0 +1,468 @@
...
+            error_type = the_error.err_type
+            error_msg = the_error.err_msg
+
+            # Build raising function
+            raise_function = error_type + "Error"
+            raise_function = f"""
+                raise {raise_function}('''{error_msg}''')
+                """
+
+            exec(raise_function)  # Issue error
```
je-cook, line 114: Never use `exec` it is usually a security issue. This would be avoided when the errors are called where they are needed

bluemira/power_cycle/base.py
```diff
@@ -0,0 +1,468 @@
...
+
+            # Issue error
+            raise ValueError(
+                f"""
+                Unknown error label for error in class {class_name}.
+                """
+            )
+
+    @staticmethod
+    def _validate_list(input) -> list:
```
je-cook, line 128: For the few places this could be needed just add it to the functions that do. You shouldnt try and type protect too vigorously.

bluemira/power_cycle/base.py
```diff
@@ -0,0 +1,468 @@
...
+            raise TypeError(
+                """
+                The argument 'ax' used to create a plot is not
+                an instance of the `Axes` class.
+                """
+            )
+        return ax
+
+    @classmethod
+    def adjust_2d_graph_ranges(cls, x_frac=0.1, y_frac=0.1, ax=None):
```
je-cook, line 385: not a classmethod

bluemira/power_cycle/timeline.py
```diff
@@ -0,0 +1,306 @@
+"""
+Classes to define the timeline for Power Cycle simulations.
+"""
+
+# Import
+from typing import Union  # , List
+
+from bluemira.power_cycle.base import PowerCycleABC as imported_abc
```
je-cook, line 8: dont rename imports like this. It makes everything harder to follow

bluemira/power_cycle/timeline.py
```diff
@@ -0,0 +1,306 @@
...
+    # ------------------------------------------------------------------
+
+    _valid_dependencies = {
+        "ss": "steady-state",
+        "tt": "transient",
+    }
+
+    # Error messages
+    @classproperty
+    def _errors(cls):
```
je-cook, line 47: I doubt this is needed, call the errors where they're needed, later maybe we can do something like this if it is needed.

je-cook (commented):
I just had a very quick run through and thought i'd put some comments before this gets too far along. I've only gone through a couple of files but the comments probably apply everywhere

- A classmethod is very rarely needed and only needed if you modify the class so that every instance of that class now acts differently.

- Errors should be raised at their location, dont store the error messages in a different location to twhere it is being raise, same for the raising of the exception. We can deduplicate some error messages possibly towards the end. Not at the start
- Generally don't change the import names unless really needed.

_TIAGO_FILES_/SCRIPTS/EUDEMO_powercycle_builder.py
```diff
@@ -0,0 +1,33 @@
+"""
```
je-cook, line 1: This whole folder should be removed from the PR, you shouldn't commit it

je-cook: there also looks like there are some changes on tplobo/powercycle_fix that you might have meant to be here

bluemira/power_cycle/base.py
tplobo: Got it, will remove it in the new commit.

bluemira/power_cycle/timeline.py
tplobo: I usually do this when my classes are still frequently changing name, while I still haven't decided on a particular nomenclature for the architecture. I did not intend to leave them, but good that you mention it: I will already remove it in the new commit.

tplobo:
> I just had a very quick run through and thought i'd put some comments before this gets too far along. I've only gone through a couple of files but the comments probably apply everywhere

@je-cook, after I change things, should I resolve your comments or do you prefer to come here and do it yourself?

bluemira/power_cycle/base.py
tplobo: @je-cook,
I have defined this as a classmethod because I call another static method from the same class inside the routine.
Using a classmethod ensures the implementation is not dependent on the name I give to the class. Is this wrong?

bluemira/power_cycle/base.py
tplobo: I understood your comment and explanation during our meeting. I will forgo of the `PowerCycleError` class and just implement errors as dict properties of each class. This way only the error message is stored in the `_errors` attribute and the `raise` command will not be dynamically called.

_TIAGO_FILES_/SCRIPTS/EUDEMO_powercycle_builder.py
je-cook: I really meant on _your_ github account not our organisation

bluemira/power_cycle/base.py
je-cook: point 1 as we talked about is something we possibly can deal with later and for now just duplicate.

For point 2 it is then not a `staticmethod` because it requires information from the instance of the class (what methods exist). You should nearly always have to initialise classes as a `classmethod` should modify the class otherwise there is no need for it. 

Modifying a class rather than an instance of the class is dangerous because if a new instance of that class is created that will also have been modified.

bluemira/power_cycle/base.py
je-cook: That is true but is not the whole story. As I said above I would only really use a `classmethod` to modify the class or possibly to modify initialisation of the class neither are the case here. 

In fact I would suggest that this function should be a `staticmethod` or possibly a pure function outside the class as all it is doing is creating an axis. I would probably call the validation before I create some axes (in the method that calls this function).

This function will always (if designed in the expected way) be called from a class instance. If you want to both keep the method in the class and internally call the validate method (the only use of `cls`) then this is just a normal method because it requires access to the accessible methods within the class.

hsaunders1904 (changes requested):
This looks like a great start, the code looks well laid out and logical - I'm not too familiar with the physics, so can only comment on the code. There's a fair bit here, as well as being your first PR, so there are plenty of comments, sorry! 😄 

A general point on code comments: comments are good, but they should say how and why, not what. There's a fair bit of commenting that's just describing what your code's doing, where your code is perfectly clear as it is. I would go through and remove all those kinds of comments. 

I've never really seen `classmethod`s used in this way before, I must admit I'm still not entirely clear what the benefit is. At least for the validators, I tend to agree with @je-cook that we should be defining errors where they're raised, not as class properties. I think then a lot of the `classmethod` complications go away.

There's been quite a bit of previous debate about whether plotting functions belong on the classes, or separately as module-level functions. I can't really remember what decisions we made if any, but the geometry module, at least, has plotting functions that are separate from the objects. This may not necessarily need to be the case for this module, but we should probably discuss this at some point.

bluemira/power_cycle/base.py
```diff
@@ -0,0 +1,492 @@
...
+            default, this fraction is set to 10%.
+        x_frac: float
+            Fractional number by which y-scale will be enlarged. By
+            default, this fraction is set to 10%.
+        ax: Axes
+            Instance of the `matplotlib.axes.Axes` class. By default,
+            the currently selected axes are used.
+        """
+
+        # Validate axes
```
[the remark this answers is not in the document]
tplobo: Really? No comments at all? To me it feels so uninviting and difficult to read when there are only lines of code.

bluemira/power_cycle/display.py
```diff
@@ -0,0 +1,66 @@
...
+
+    Parameters
+    ----------
+    default_display_units: str
+        Path to the default `display_units.JSON` file
+    desired_display_units: str
+        Path to the desired `display_units.JSON` file
+
+    Other Parameters
+    ----------------
```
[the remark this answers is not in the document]
tplobo: This class is not completed yet, I foresee additional (secondary) parameters. But if there are none, I'll remove the docstring section, of course.

bluemira/power_cycle/loads.py
```diff
@@ -0,0 +1,1065 @@
...
+
+    # ------------------------------------------------------------------
+    # CLASS ATTRIBUTES
+    # ------------------------------------------------------------------
+
+    # Default number of points in each curve segment
+    _n_points = 100
+
+    # Implemented models (add model name here after implementation)
+    _valid_models = ["ramp", "step"]
```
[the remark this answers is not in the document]
tplobo: What did you have in mind? A small 'model' class with a list of models, that can be expanded, each item associated to a function/method for computing the interpolation?

bluemira/power_cycle/base.py
```diff
@@ -0,0 +1,492 @@
...
+                keywords.append(word)
+
+        # Output keywords
+        return keywords
+
+
+# ######################################################################
+# POWER CYCLE UTILITIES
+# ######################################################################
+class PowerCycleUtilities:
```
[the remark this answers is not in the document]
tplobo: Ah, good to know. I suggest we keep track of these "patterns" to write them down in the developer documentation. It will make things much easier to new programmers in the project.
I'm in the process of implementing this now.

_TIAGO_FILES_/SCRIPTS/EUDEMO_powercycle_builder.py
tplobo: Done!

bluemira/power_cycle/base.py
tplobo: I have resolved this problem in this particular case by doing what @hsaunders1904 suggested and placing both methods as stand-alone functions in a separate file 'tools.py'. I will check if this issue comes up anywhere else.

bluemira/power_cycle/base.py
tplobo: I had a conversation with @ivanmaione, and I will probably create a main `PowerCycleError` class, and create subclasses for each class in the module. This way the classes do not need each an `_errors` attribute. What do you think @je-cook?

bluemira/power_cycle/base.py
```diff
@@ -0,0 +1,492 @@
...
+    """
+    Useful functions for multiple classes in the Power Cycle module.
+    """
+
+    # ------------------------------------------------------------------
+    # DATA VISUALIZATION
+    # ------------------------------------------------------------------
+    @staticmethod
+    def print_header(header=None):
```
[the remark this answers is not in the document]
je-cook: It says outdated because the code that @hsaunders1904 has commented on has changed. The functions he's talking about still exist

bluemira/power_cycle/base.py
je-cook: Yep that sounds fine

bluemira/power_cycle/loads.py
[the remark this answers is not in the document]
je-cook: It would be quite simple:
```python
from enum import Enum, auto

class PowerCycleModel(Enum):
    RAMP = auto()
    STEP = auto()
```

Its usually convention for `Enum` variable names to be captialised . An `Enum` will automatically add existence checking with no extra work from the developer. Have a read through @hsaunders1904 's link.

bluemira/power_cycle/loads.py
```diff
@@ -0,0 +1,1065 @@
...
+
+        # Retrieve `model` attributes
+        this_model = self.model
+        other_model = other.model
+
+        # Create and output `another`
+        another_set = this_set + other_set
+        another_model = this_model + other_model
+        another_name = "Resulting PowerLoad"
+        another = PowerLoad(another_name, another_set, another_model)
```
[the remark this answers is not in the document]
tplobo: PowerLoad will not be directly used by the users. What I did in my MIRA implementations was to have a Power(Load?)Manager class that adds them, depending on inputs that characterize all power loads.

The idea is to have a pre-allocated PowerLoad instance (e.g. "coil loads") with empty vectors, and when specifying that it is the sum of other instances (e.g. "coil 1 load" and "coil 2 load") the command to re-write it only takes their relevant attributes, not their names. As in:
```
all_coils = PowerLoad("All Coils",Null,Null)
coil_1 = PowerLoad("Coil 1",coil_1_set,coil_model)
coil_2 = PowerLoad("Coil 2",coil_2_set,coil_model)

all_coils = coil_1 + coil_2
```
The last line does not overwrite `all_coils`, but only takes the 'set' and 'model' parameters of the (temporary) instance created by the sum (the `another` object, with `name` "Resulting PowerLoad").

Let me know what you think.

bluemira/power_cycle/loads.py
```diff
@@ -0,0 +1,1065 @@
...
+
+            # Refine current time vector
+            current_time = self._refine_vector(current_time, n_points)
+
+            # Append current time in time vector for plotting
+            time = time + current_time
+
+        # Sort and unique of complete time vector
+        time = list(set(time))
+        time.sort()
```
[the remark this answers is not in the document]
tplobo: I wanted to be clear to anyone reading that the list should be sorted. Is this a case in which I should make a comment like we discussed?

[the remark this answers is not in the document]
hsaunders1904: You could use a new variable `sorted_times` - then it's obvious that the times are sorted everywhere the variable is used, not just here.

```python
sorted_times = list(set(time))
```

bluemira/power_cycle/loads.py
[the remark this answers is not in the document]
hsaunders1904: It's not clear to me how that would work. 

The final line will discard the `PowerLoad` object you defined on the first line. Variables are just pointers in Python. After the first line `all_coils` points to the `PowerLoad` instance you've created, but after the last line, `all_coils` now points at the `PowerLoad` instance you created in the `__add__` function. So the name will be `'Resulting PowerLoad'`, not `'All Coils'`.

Or am I misunderstanding you?

bluemira/power_cycle/base.py
```diff
@@ -0,0 +1,492 @@
...
+
+        # Create string message
+        values_msg = "', '".join(string_values)
+        values_msg = "'" + values_msg + "'"
+
+        # Output message
+        return values_msg
+
+    @staticmethod
+    def add_dict_entries(dictionary, new_entries):
```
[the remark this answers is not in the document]
tplobo: @hsaunders1904, I have re-written the function in the `tools` file with some new syntaxes acceptable in the latest versions of python. Should be more readable now.

bluemira/power_cycle/loads.py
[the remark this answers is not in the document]
tplobo: I am sorry, I was not clear in my last comment. The code I wrote will definitely overwrite `all_coils`.

What I meant was that the action of creating and attributing values to PowerLoads is done by the PowerLoadManager class. It will sum power loads (which creates an instance of name "Resulting..."), but extract its `set` and `mod` arguments to apply them to a pre-defined PowerLoad instance that was initially created with some default/null arguments.
This way all PowerLoad instances are already defined when the PowerLoadManager class starts computing things, including the loads that will be sum of others. I believe this is preferable because I create all power loads when reading from an input file, and only compute them at a later stage.

Something like this, but applied using some method in the class (in my previous comment I meant to override the `setter` for power loads in the manager class):
```
all_coils = PowerLoad(name = "All Coils", set = Null, mod = Null)
coil_1 = PowerLoad("Coil 1",coil_1_set,coil_model)
coil_2 = PowerLoad("Coil 2",coil_2_set,coil_model)

temp_powerload = coil_1 + coil_2
all_coils.set = temp_powerload.set
all_coils.mod = temp_powerload.mod
```

bluemira/power_cycle/timeline.py
```diff
@@ -0,0 +1,309 @@
...
+class PowerCyclePhase(PowerCycleABC):
+    """
+    Class to define phases for a Power Cycle pulse.
+
+    Parameters
+    ----------
+    name: str
+        Description of the `PowerCyclePhase` instance.
+    label: str
+        Shorthand label for addressing the `PowerCyclePhase` instance.
```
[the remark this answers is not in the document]
tplobo: This is something we discussed in the Power Cycle module design meetings, that we decided to debate later whether it is necessary... The `name` attribute is the component/fluid/... actual descriptive name, while `label` is a 2 or 3 letter string that I use to always address/map the objects in dictionaries and the like.

tests/power_cycle/test_powercycletools.py
```diff
@@ -0,0 +1,96 @@
...
+)
+
+# Simple vectors for sample plotting
+sample_x = [1, 2]
+sample_y = [3, 4]
+
+bluemira_print("Test Power Cycle Tools")
+
+
+def test_add_dict_entries(visualize=None):
```
je-cook, line 21: `visualize` will never be set when the test is run it will always be none. 

```suggestion
def test_add_dict_entries():
```

je-cook (commented):
Mostly the test file is fine. I would go look at eg `tests/display/test_plotter.py` which apart from the variable defined at the top of the test file is a good representation of what we're looking for.

You don't have to collect tests into classes if they're unrelated or dont use any of the same variables. If they are related I would put them in the same test class.

tests/power_cycle/test_powercycletools.py
```diff
@@ -0,0 +1,96 @@
...
+    new_entries = {
+        "value_1": 1,
+        "value_4": 4,
+    }
+    new_dict = _add_dict_entries(old_dict, new_entries)
+    if visualize:
+        bluemira_print("Test Power Cycle Tools ('_add_dict_entries')")
+        pprint(old_dict)
+        pprint(new_entries)
+        pprint(new_dict)
```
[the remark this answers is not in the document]
tplobo: > otherwise it will not be in our logging system.

Ah, I did not know that. I have added any `print` call that might be useful for debugging the classes in calls of the `bluemira_debug` method.

tests/power_cycle/test_powercycle_tools.py
```diff
@@ -0,0 +1,153 @@
...
+        string_from_argument = _join_delimited_values(test_argument)
+        bluemira_debug(
+            f"""
+            {script_title()} (_join_delimited_values)
+
+            Generated string:
+            {string_from_argument}
+            """
+        )
+    assert isinstance(string_from_argument, str)
```
je-cook, line 64: This test is just testing builtin python functionality both of the tests here can be accomplished by 
`' ,'.join(all_arguments[0])` or `' ,'.join(all_arguments[1])`. Remove both the function and the test.

```suggestion
def test_join_delimited_values():
    all_arguments = [
        ["value 1", "value 2", "value 3"],
        {"value_1": 1, "value_2": 2, "value_3": 3},
    ]
    for test_argument in all_arguments:
        string_from_argument = _join_delimited_values(test_argument)
        bluemira_debug(
            f"""
            {script_title()} (_join_delimited_values)

            Generated string:
            {string_from_argument}
            """
        )
    assert isinstance(string_from_argument, str)
```

tests/power_cycle/test_powercycle_time.py
```diff
@@ -0,0 +1,287 @@
...
+)
+
+# from bluemira.power_cycle.tools import (
+#    adjust_2d_graph_ranges,
+#    validate_axes,
+# )
+
+
+def script_title():
+    return "Test Power Cycle 'time'"
```
je-cook, line 25:
```suggestion
```

tests/power_cycle/test_powercycle_time.py
```diff
@@ -0,0 +1,287 @@
...
+        sample_name = "Transition between dwell and flat-top"
+        sample_breakdown = {
+            "pump-down": constants.raw_uc(10, "minute", "second"),
+            "ramp-up": constants.raw_uc(5.2, "minute", "second"),
+            "heating": constants.raw_uc(1.4, "minute", "second"),
+        }
+        self.sample_name = sample_name
+        self.sample_breakdown = sample_breakdown
+
+    def test_validate_breakdown(self):
```
je-cook, line 112: Redo this test like above

tests/power_cycle/test_powercycle_net.py
```diff
@@ -0,0 +1,509 @@
...
+    PowerDataError,
+    PowerLoad,
+    PowerLoadError,
+    PowerLoadModel,
+)
+from bluemira.power_cycle.tools import adjust_2d_graph_ranges, validate_axes
+
+
+def script_title():
+    return "Test Power Cycle 'net'"
```
je-cook, line 21:
```suggestion
```

tests/power_cycle/test_powercycle_net.py
```diff
@@ -0,0 +1,509 @@
...
+
+
+def test_NetPowerABCError():
+    with pytest.raises(NetPowerABCError):
+        raise NetPowerABCError(
+            None,
+            "Some error in the 'NetPowerABC' class.",
+        )
```
je-cook, line 29: test the function the raises this

tests/power_cycle/test_powercycle_net.py
```diff
@@ -0,0 +1,509 @@
+from pprint import pformat
```
je-cook, line 1: Similar pattern of changes needed in this file

tests/power_cycle/test_powercycle_base.py
```diff
@@ -0,0 +1,114 @@
+from pprint import pformat
```
je-cook, line 1: And here

bluemira/power_cycle/tools.py
```diff
@@ -0,0 +1,179 @@
...
+        # Create string message with delimiters
+        values_msg = "', '".join(string_values)
+        values_msg = "'" + values_msg + "'"
+
+    except (TypeError):
+        TypeError(
+            "The argument to be transformed into a delimited string of "
+            "values is not a `list` or `dict`."
+        )
+    return values_msg
```
je-cook, line 82:
```suggestion
```

bluemira/power_cycle/base.py
```diff
@@ -0,0 +1,136 @@
...
+            all_errors = self._errors()
+            if case in all_errors:
+                extra_msg = str(all_errors[case])
+                message = message + " " + extra_msg
+            else:
+                raise ValueError(
+                    f"The requested error case '{case}' has not been "
+                    f"defined for the error class '{self.__class__}'."
+                )
+        super().__init__(message)
```
je-cook, line 31:
```suggestion
        try:
            msg += f" {self._errors()[case]}"
        except KeyError:
            raise ValueError(
                    f"The requested error case '{case}' has not been "
                    f"defined for the error class '{self.__class__}'."
                ) from None
        super().__init__(msg)
```

bluemira/power_cycle/base.py
```diff
@@ -0,0 +1,136 @@
...
+    Exception class for 'PowerCycleABC' class of the Power Cycle module.
+    """
+
+    def _errors(self):
+        errors = {
+            "name": [
+                "Invalid 'name' parameter. The 'name' attribute of ' "
+                f"instances of the {self._source} class must be of "
+                "the 'str' class.",
+            ],
```
je-cook, line 55: strings auto join in a list and a tuple if you remove the last command put it in a tuple it will just be a string
```suggestion
            "name": (
                "Invalid 'name' parameter. The 'name' attribute of ' "
                f"instances of the {self._source} class must be of "
                "the 'str' class."),
```

je-cook (changes requested):
Ok I've done enough for now. There are other things but I think this is enough to start with

tplobo:
About the quality control issues:

1. I changed the line to:
```
out = get_primary_screen_size(10)
```
but I still get a failed test. Since you said this is not important, we don't need to keep going, but technically `pytest tests --reactor` does not pass for me.

2. The folder `documentation/source/autoapi` does not exist for me. I don't know if `sphinx` is not working because my module folder includes two sub-folders (see original screenshot).

3. I ran `python -m pip install --no-cache-dir -e .'[dev]'` in my `bluemira` environment and it returns the following for `black`:
```
Requirement already satisfied: black in /Users/.../miniforge3/envs/bluemira/lib/python3.8/site-packages (from bluemira==1.2.0+146.g3aa1a92e) (22.10.0)
```
If I ask for the version directly:
```
... % black --version
black, 22.10.0 (compiled: yes)
Python (CPython) 3.8.14
```
which I believe is not the latest version, because in `requirements-develop.txt` there is `black==23.1.0`. Should I re-install bluemira?

Also, what are your VSCode settings for python? I have:
Screen Shot 2023-03-21 at 14 46 43 PM

Maybe VSCode is taking precedence over bluemira environment and that is my issue.

bluemira/power_cycle/net/loads.py
```diff
@@ -0,0 +1,1649 @@
...
+                plot_object = ax.plot(
+                    computed_time,
+                    current_curve,
+                    **kwargs,
+                )
+                current_plot_list.append(plot_object)
+
+                list_of_plot_objects.append(current_plot_list)
+
+        return ax, list_of_plot_objects
```
je-cook, line 615:
```suggestion
        if detailed:
            for e, (curr_ld, curr_lm) in enumerate(
                zip(self.loaddata_set, self.loadmodel_set)
            ):
                # Plot current LoadData with seconday kwargs
                curr_ld._make_secondary_in_plot()

                # Plot current curve as line with secondary kwargs
                kwargs.update(curr_ld._plot_kwargs)
                plot_object = ax.plot(
                    computed_time,
                    self._single_curve(curr_ld, curr_lm, computed_time),
                    **kwargs,
                )

                list_of_plot_objects.extend((curr_ld.plot(ax=ax)[1], plot_object))

        return ax, list_of_plot_objects
```

je-cook (commented):
I've gone through about half of net/loads.py and the general theme is you've over allocated variables. If you could try and follow some of the themes of my suggestions I think the nLOCs will be reduced drastically and improve readability.

I havent done any overall archtecture and logic reviewing yet, just code QA

je-cook:
>Since you said this is not important, we don't need to keep going, but technically `pytest tests --reactor` does not pass for me.

Ok no problem maybe we can narrow it down in the future.

> 2. The folder `documentation/source/autoapi` does not exist for me. I don't know if `sphinx` is not working because my module folder includes two sub-folders (see original screenshot).

I'm not sure on this I'd need to see I think. Nuclear option is to remove the folder documentation/build but I doubt that will work. I will look into it but I think possibly some references need to be added somewhere. You can run without the `-W` flag for now and it should at least produce something.

> 3. I ran `python -m pip install --no-cache-dir -e .'[dev]'` in my `bluemira` environment and it returns the following for `black`:
> 

You just need to update your dependencies. The above command only checks you have some version installed, use our script `bash scripts/update_dependencies.sh`

> Should I wait until we fix these, before pressing "Ready for review"? Or should I do it only when you create the feature branch?

I've marked as ready for review and will create a feature branch now

je-cook (commented):
Another go at loads.py

bluemira/power_cycle/tools.py
```diff
@@ -0,0 +1,334 @@
...
+    Returns the contents of a 'json' file in 'dict' format.
+    """
+    try:
+        with open(file_path) as json_file:
+            contents_dict = json.load(json_file)
+    except json.decoder.JSONDecodeError:
+        raise TypeError(
+            "The file could not be read as a 'json' file.",
+        )
+    return contents_dict
```
je-cook, line 230: I think a lot of things will be simplified if you do a bit more processing of the json files here. List of numbers shouldnt be stored as strings in the json eg this

```
                        "time": [
                            "[0, 1]",
                            "[0, 1]",
                            "[0, 1]"
                        ],
```

should just be this
```
                        "time": [
                            [0, 1],
                            [0, 1],
                            [0, 1]
                        ],
```
I think in all cases in your current json files this will suffice as the reader. We also want the json module to raise whatever error is appropriate for the file given.
```suggestion
    with open(file_path) as json_file:
        contents_dict = json.load(json_file)

    for k, value in contents_dict.items():
        if isinstance(value, list) and k in ('time', 'data'):
            contents_dict[k] = np.asarray(value)
    return contents_dict
```

je-cook:
Included/ superceeded by #2873

PR #1286, opened 2022-07-27
⚡️ Timeout display check

### LINKED ISSUE #264: Avoid hangs if DISPLAY is set up no x-server is running

Various runs of bluemira hang in WSL if DISPLAY is set but no x-server is running on the Windows side. Since some runs do not need images to be displayed and DISPLAY will likely be set in the .bashrc, this can sometimes lead to the x-server being needed to be switched on to avoid these hangs.

## Proposed solution

Implement a function to attempt a socket connection (with a sensible timeout) to detect if the x-server is running. If the x-server isn't running then warn the user and carry on. This may need some other updates such as setting to a display-less backend (`Agg` in matplotlib).

je-cook:
## Linked Issues

Closes #264

## Description

By using multiprocess with 1 core we can timeout the display checking function and produce a warning on timeout

## Interface Changes

None

## Checklist

### DIFF
--- a/bluemira/display/auto_config.py
+++ b/bluemira/display/auto_config.py
@@ -32,16 +32,21 @@ import numpy as np
 import seaborn as sns
 from PySide2 import QtWidgets
 
-from bluemira.base.look_and_feel import bluemira_warn
+from bluemira.base.look_and_feel import bluemira_debug, bluemira_warn
 from bluemira.display.palettes import BLUEMIRA_PALETTE
 
 
 @functools.lru_cache(1)
-def get_primary_screen_size():
+def get_primary_screen_size(timeout: float = 3):
     """
     Get the size in pixels of the primary screen.
 
     Used for sizing figures to the screen for small screens.
+
+    Parameters
+    ----------
+    timeout: float
+        timeout value in seconds
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


bluemira/display/auto_config.py
```diff
@@ -48,6 +50,27 @@ def get_primary_screen_size():
     height: Union[int, None]
         height of the primary screen in pixels. If there is no screen returns None
     """
+    pool = Pool(processes=1)
```
hsaunders1904, line 53: Could you use `with Pool(processes=1)` to avoid having to manually clean up?

tests/display/test_auto_config.py
```diff
@@ -0,0 +1,57 @@
...
+    time.sleep(t)
+    return True
+
+
+class TestGetScreenSize:
+    def setup(self):
+        get_primary_screen_size.cache_clear()
+
+    @patch("bluemira.display.auto_config._get_primary_screen_size", new=timeout)
+    def test_timeout(self, caplog):
```
hsaunders1904, line 46: Having these 2 tests take 4 seconds in total feels unnecessary. Could we make the timeout duration an argument to the function and reduce it to something like 0.05s in the tests?

hsaunders1904 (changes requested):
This is a very good warning to have - I'm always wondering what I've broken when I forget to turn my display forwarding on!

bluemira/display/auto_config.py
```diff
@@ -48,6 +50,27 @@ def get_primary_screen_size():
...
+        pool.join()
+        return val
+
+
+def _get_primary_screen_size():
+    """
+    Direct run of screen size check without subprocess
+    """
     if sys.platform.startswith("linux") and os.getenv("DISPLAY") is None:
         return None, None
```
hsaunders1904, line 75: Should we also be warning the user in this case?

je-cook: Hmm, I sort of think if you don't have a DISPLAY set up in any way it probably means that you don't have or want anything visual to pop up. Maybe a debug message would be enough?

hsaunders1904 (approved):

PR #458, opened 2021-11-23
shail_desai/cp-coils-documentation

shaildesai96:
Added CP coils documentation

PR #1287, opened 2022-07-27
TF coil parameterisation documentation

### LINKED ISSUE #607: Images showing geometry parameterisations

We currently only show the variables for a geometry parameterisation as text in the code. It would be good if we had a way to generate annotated plots of the parameterisation so that it is easy to see what the different variables correspond to.

## Proposed solution

Implement a function on the `GeometryParameterisation` class that can generate an annotated plot of the parameterisation with the variables shown. Include a custom sphinx directive to call this when we generate the documentation.

## Alternative solutions

We could manually make equivalent plots, but they could be less easy to maintain.

### LINKED ISSUE #608: Describe geometry parameterisation variables in docstring

Currently it is not possible to see the variables used in a geometry parameterisation without either generating a dummy parameterisation or digging into the code. It would be nice if the information about the variables was easier to access for a user.

## Proposed solution

Include the variables and their descriptions in the class docstring for `GeometryParameterisation` classes.

## Alternative solutions

We could also look to auto-generate the variable list and default values by including a custom sphinx directive in the class docstrings that generates an instance of the class and prints out the summary table. That would be less effective for e.g. hover text in IDEs though.

je-cook:
## Linked issues

Closes #607 
Closes #608

## Description

More development building on #458  including generation of images.

Ive added figures to the api docs for the parameterisations. I havent added all of the annotations but the shapes are plotted. There are TODO's that I will add them in after the next release. 

I've also rearranged the geometry docs a bit.

## Interface Changes

None

## Checklist


documentation/source/geometry/overview.rst
```diff
@@ -0,0 +1,282 @@
...
+        # shell creation
+        shell = geo.shell.BluemiraShell(faces, "shell")
+
+
+* a `BluemiraSolid`
+
+    A BluemiraSolid object is defined by its boundary that must be a closed
+    BluemiraShell object.
+
+    .. code-block:: pycon
```
hsaunders1904, line 86: I think most of these can just be 'python' code blocks. We don't get the syntax highlighting with `pycon` as I guess it just assumes it's all console output

documentation/source/geometry/overview.rst
```diff
@@ -0,0 +1,282 @@
...
+
+        bluemira.geometry.tools.save_as_STP(bmface, "face.step")
+        bluemira.geometry.tools.save_as_STP([bmface1, bmface2], "compound.step")
+
+
+FreeCAD objects
+---------------
+
+Below a list of typical geometry object properties of FreeCAD. Only that labelled as
```
hsaunders1904, line 126:
```suggestion
Below a list of typical geometry object properties of FreeCAD. Only those labelled as
```

documentation/source/geometry/parameterisation.rst
```diff
@@ -0,0 +1,188 @@
...
+please click on the above classes to be taken to our API documentation of each class.
+
+PrincetonD
+^^^^^^^^^^
+
+.. plot::
+    :caption: Princeton D
+
+    from bluemira.geometry.parameterisations import PrincetonD
+    PrincetonD().plot()
```
hsaunders1904, line 25: These look great!

hsaunders1904 (changes requested):

bluemira/geometry/parameterisations.py
```diff
@@ -767,6 +950,22 @@ def create_shape(self, label=""):
 
         return BluemiraWire(wires, label=label)
 
+    def _label_function(self, ax, shape):
+        """
+        Adds labels to parameterisation plots
+
+        TODO add labels for r1-5 a1-5
```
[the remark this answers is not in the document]
je-cook: I was hoping not because it will be quite a bit more effort to do it properly... I havent done angles at all yet

bluemira/geometry/parameterisations.py
```diff
@@ -232,6 +234,111 @@ def from_json(cls, file: Union[str, TextIO]) -> GeometryParameterisation:
...
+            label of annotation
+        xy1: Tuple
+            Tuple for first arrow point
+        xy2: Tuple
+            Tuple for second arrow point
+        xy3: Tuple
+            Tuple for arrow label location
+
+        """
+        key = key.replace("_", "")  # TODO proper latex
```
hsaunders1904, line 255: Another TODO here. If it's important to do, then we should probably open an issue, or we'll forget about it.

bluemira/geometry/parameterisations.py
```diff
@@ -232,6 +234,111 @@ def from_json(cls, file: Union[str, TextIO]) -> GeometryParameterisation:
...
+                ax.plot([shape.center_of_mass[0], xcor], [0, 0], color="k")
+                ax.plot(
+                    [shape.center_of_mass[0], xcor],
+                    [var["_value"], var["_value"]],
+                    color="k",
+                )
+                offset_ar_z += 1.5
+        return offset_ar_x, offset_ar_z
+
+    def plot(self, ax=None, labels=True, **kwargs):
```
hsaunders1904, line 322: `kwargs` here are unused. Should they be passed into the `plot_2d` call?

hsaunders1904: Should we be plotting labels by default? Feels to me like it's something users would use as a one-off, but wouldn't be using in the general case

je-cook: Done on both counts

bluemira/geometry/parameterisations.py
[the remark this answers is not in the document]
hsaunders1904: Do we want an issue to do this, then?

hsaunders1904 (approved):
Really nice addition ✔️

bluemira/geometry/parameterisations.py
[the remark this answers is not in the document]
je-cook: Yeah just opened #1747

PR #1288, opened 2022-07-27
Correction of the Ax_b_constraint function to make it compatible with list of constrained flux points

### LINKED ISSUE #1285: PsiConstraint currently incompatible with x and z np.arrays
## Describe the bug

When arrays are used to define the `x` and the `z` of the flux points used in `PsiConstraint` a matrix operation dimension error appears.

## Steps to reproduce

Use an array of points for the `PsiConstraint` and run an equilibrium optimization.

## Expected behaviour

No errors and an optimization that happends (but may fail for other reasons :D)

## Screenshots / Tracebacks

error traceback:

``` bash
  File "equilibrium_standalone.py", line 141, in <module>
    equilibrium, FoM_constraints = FBE_solver(
  File "/home/skahn/codes/step/step-blueprint-reactor-design/step_reactor/fbe_func.py", line 175, in FBE_solver
    iterator()
  File "/home/skahn/codes/bluemira/bluemira/equilibria/solve.py", line 495, in __call__
    iterator = iter(self)
  File "/home/skahn/codes/bluemira/bluemira/equilibria/solve.py", line 519, in __iter__
    self._setup()
  File "/home/skahn/codes/bluemira/bluemira/equilibria/solve.py", line 612, in _setup
    self._initial_optimise_coilset()
  File "/home/skahn/codes/bluemira/bluemira/equilibria/solve.py", line 606, in _initial_optimise_coilset
    self._optimise_coilset(**kwargs)
  File "/home/skahn/codes/bluemira/bluemira/equilibria/solve.py", line 461, in _optimise_coilset
    coilset = self.opt_prob.optimise(self.fixed_coils)
  File "/home/skahn/codes/bluemira/bluemira/equilibria/opt_problems.py", line 809, in optimise
    currents = self.opt.optimise(initial_currents)
  File "/home/skahn/codes/bluemira/bluemira/utilities/optimiser.py", line 133, in optimise
    x_star = super().optimise(x0)
  File "/home/skahn/codes/bluemira/bluemira/codes/_nlopt_api.py", line 430, in optimise
    x_star = self._opt.optimize(x0)
  File "/home/skahn/miniforge3/envs/bluemira/lib/python3.8/site-packages/nlopt/nlopt.py", line 334, in optimize
    return _nlopt.opt_optimize(self, *args)
  File "/home/skahn/codes/bluemira/bluemira/utilities/opt_problems.py", line 71, in __call__
    return self._f_constraint(constraint, vector, grad, **self._args)
  File "/home/skahn/codes/bluemira/bluemira/equilibria/opt_constraint_funcs.py", line 98, in Ax_b_constraint
    constraint[:] = np.dot(a_mat, scale * vector) - b_vec - value
  File "<__array_function__ internals>", line 5, in dot
ValueError: shapes (7,2) and (7,) not aligned: 2 (dim 1) != 7 (dim 0)
```

## Propose solution 

See PR xxx

sebkahn:
… an array of flux point constrained.

## Linked Issues

Closes #1285

## Description

## Interface Changes

## Checklist

### DIFF
--- a/bluemira/equilibria/opt_constraint_funcs.py
+++ b/bluemira/equilibria/opt_constraint_funcs.py
@@ -95,9 +95,9 @@ def Ax_b_constraint(constraint, vector, grad, a_mat, b_vec, value, scale):  # no
     scale: float
         Current scale with which to calculate the constraints
     """
-    constraint[:] = np.dot(a_mat, scale * vector) - b_vec - value
+    constraint[:] = a_mat.T @ vector * scale - b_vec - value
     if grad.size > 0:
-        grad[:] = scale * a_mat
+        grad[:] = scale * a_mat.T
     return constraint
 
 
--- a/bluemira/equilibria/opt_constraint_funcs.py
+++ b/bluemira/equilibria/opt_constraint_funcs.py
@@ -97,7 +97,7 @@ def Ax_b_constraint(constraint, vector, grad, a_mat, b_vec, value, scale):  # no
     """
     constraint[:] = a_mat @ vector * scale - b_vec - value
     if grad.size > 0:
-        grad[:] = scale * a_mat.T
+        grad[:] = scale * a_mat
     return constraint
 
 
--- a/bluemira/equilibria/opt_constraint_funcs.py
+++ b/bluemira/equilibria/opt_constraint_funcs.py
@@ -95,7 +95,7 @@ def Ax_b_constraint(constraint, vector, grad, a_mat, b_vec, value, scale):  # no
     scale: float
         Current scale with which to calculate the constraints
     """
-    constraint[:] = a_mat.T @ vector * scale - b_vec - value
+    constraint[:] = a_mat @ vector * scale - b_vec - value
     if grad.size > 0:
         grad[:] = scale * a_mat.T
     return constraint
--- a/bluemira/equilibria/opt_constraints.py
+++ b/bluemira/equilibria/opt_constraints.py
@@ -557,7 +557,7 @@ class PsiConstraint(AbsoluteMagneticConstraint):
         """
         Calculate control response of a CoilSet to the constraint.
         """
-        return np.array(coilset.control_psi(self.x, self.z))
+        return np.array(coilset.control_psi(self.x, self.z)).T
 
     def evaluate(self, eq):
         """
--- a/tests/equilibria/test_constraints.py
+++ b/tests/equilibria/test_constraints.py
@@ -32,6 +32,7 @@ from bluemira.equilibria.opt_constraints import (
     PsiConstraint,
 )
 from bluemira.equilibria.opt_problems import TikhonovCurrentCOP
+from bluemira.utilities.optimiser import Optimiser
 
 
 # @pytest.mark.longrun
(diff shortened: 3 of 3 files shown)


sebkahn:
One or two tests seems to fail

je-cook:
Thanks! That failure looks like one of our geometry bugs.

Could you add a test that runs to cover the former and this behaviour?

sebkahn:
> Thanks! That failure looks like one of our geometry bugs. I've rerun the CI, it can be temperamental.
> 
> Could you add a test that runs to cover the former and this behaviour?

Glad it is not something on my end

Found a slicker way to make the correction though !
Putting WIP wayting for the slicker bug fix :)

je-cook:
could you rebase on develop please to remove the extra commits?

je-cook (commented):
Can we have a test to check this functionality?

sebkahn:
> Can we have a test to check this functionality?

Of course, the constraints testing is now covering `PsiConstraints` too and will error out if the wrong vector stricture is provided to the response function.

bluemira/equilibria/opt_constraint_funcs.py
```diff
@@ -95,9 +95,9 @@ def Ax_b_constraint(constraint, vector, grad, a_mat, b_vec, value, scale):  # no
     scale: float
         Current scale with which to calculate the constraints
     """
-    constraint[:] = np.dot(a_mat, scale * vector) - b_vec - value
+    constraint[:] = a_mat @ vector * scale - b_vec - value
     if grad.size > 0:
-        grad[:] = scale * a_mat
+        grad[:] = scale * a_mat.T
```
sebkahn, line 100: Neet to test the optimization whith a gradient based method to see if this is a valid correction

alexanderianblair: Agreed - I think it's worth adding a test to check this constraint (and future ones) for gradient based methods (eg. SLSQP) explicitly in `test_constraints`, given how easy it would be for a newly written constraint to get the shape of grad wrong! 

FWIW - on a closer look, I don't think the transpose should be on L100 - I think it's already included in `a_mat` from L560 in `opt_constraints.py`. Easy to confirm with a quick test, though!

bluemira/equilibria/opt_constraints.py
```diff
@@ -557,7 +557,7 @@ def control_response(self, coilset):
         """
         Calculate control response of a CoilSet to the constraint.
         """
-        return np.array(coilset.control_psi(self.x, self.z))
+        return np.array(coilset.control_psi(self.x, self.z)).T
```
alexanderianblair, line 560: Agreed - this makes it consistent in shape with the other constraints.

tests/equilibria/test_constraints.py
```diff
@@ -85,9 +103,9 @@ def test_constraint_weights(self):
             # Test that weights have been applied
 
             problem = TikhonovCurrentCOP(
```
alexanderianblair, line 105: Could use a gradient based algorithm here to check gradient shapes - then new constraints could be added to this test much more easily when necessary.

alexanderianblair (commented):
Looks good - but just needs a test for the gradient part of the constraint, to avoid possible unexpected behaviour for gradient-based optimisation. Otherwise happy.

tests/equilibria/test_constraints.py
sebkahn: > Could use a gradient based algorithm here to check gradient shapes - then new constraints could be added to this test much more easily when necessary.

The `ThikhonovCurrentCOP` problem is already using `SLSQP` (line 779 if `opt_problem.py` defining the default optimiser used here)  that is a gradient based, so the gradient based is already covered ?

bluemira/equilibria/opt_constraint_funcs.py
sebkahn: Reverted to initial value

sebkahn: After test on the STEP equilibrium designeer the transpose should be not there.
As for the the reason the test does not catch it I do not understand it.

je-cook:
Superceeded and covered in #1963

PR #1294, opened 2022-08-01
Run plotting tests with non-interactive backend

### LINKED ISSUE #1173: Run plotting tests with non-interactive backend

We currently have a suite of plotting tests, however these are not run in the CI, so it's always too late before we see that they're broken. It's also not easy for runners of the tests to immediately see if something is wrong, unless they are very familiar with what the figures should look like. 

We want to run these tests in the CI to ensure the tests at least run without error. In an ideal world, we would also programmatically make some assertions about the resulting figures. Typically, we might save a reference figure and compare it to the figure produced by the test using [`matplotlib.testing.compare`](https://matplotlib.org/stable/api/testing_api.html#module-matplotlib.testing.compare). You can also test attributes of the figure, Matplotlib exposes `Artist` objects (lines, patches, etc.) on `Axes`, that we can check attributes of (e.g., coordinates, colo[u]r).

## Proposed solution

We turn off Matplotlib's interactive backend when running the tests without `--plotting-on`; i.e., `--plotting-on` turns on the interactive backend. This way the plot objects will be created, but will not open blocking displays.

The above will work for many of our tests, but we also need to consider the CAD viewer. I'm open to ideas on how we can stop that blocking the running of the tests. A quick and dirty solution is to just mock it out everywhere.

## Outputs

- The ability to run the plotting tests without GUIs blocking the execution of the test suite.
- The plotting tests run on the CI.

hsaunders1904:
## Linked Issues

Closes #1173 

## Description

Plotting tests are now enabled by default and run with a non-interactive backend. This means we run all the plotting tests on the CI etc., but no interactive windows prevent the execution of subsequent tests. The `--plotting-on` flag still remains, and is used to switch to an interactive backend. The `exec_` method on `QApplication` is mocked before the test suite is run, so no blocking window is opened to display CAD; again this behaviour can be overridden using the `--plotting-on` flag.

Sorry for so many review requests, but this touches almost everything!

## Checklist

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -23,9 +23,10 @@
 Used by pytest for configuration like adding command line options.
 """
 
+from unittest import mock
+
 import matplotlib as mpl
 
-import tests
 from bluemira.base.file import try_get_bluemira_private_data_root
 
 
@@ -37,7 +38,7 @@ def pytest_addoption(parser):
         "--plotting-on",
         action="store_true",
         default=False,
-        help="switch on plotting in tests",
+        help="switch on interactive plotting in tests",
     )
     parser.addoption(
         "--longrun",
@@ -67,11 +68,14 @@ def pytest_configure(config):
     """
     Configures pytest with the plotting and longrun command line options.
     """
-    if config.getoption("--plotting-on"):
-        tests.PLOTTING = True
-    else:
+    if not config.getoption("--plotting-on"):
         # We're not displaying plots so use a display-less backend
         mpl.use("Agg")
+        # Disable CAD viewer by mocking out QApplication.exec
+        from PySide2.QtWidgets import QApplication
+
+        app = QApplication([])
+        mock.patch.object(app, "exec_").start()
 
     options = {
         "longrun": config.option.longrun,
--- a/conftest.py
+++ b/conftest.py
@@ -71,11 +71,9 @@ def pytest_configure(config):
     if not config.getoption("--plotting-on"):
         # We're not displaying plots so use a display-less backend
         mpl.use("Agg")
-        # Disable CAD viewer by mocking out QApplication.exec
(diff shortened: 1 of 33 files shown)


je-cook:
Really glad to see this coming to fruition!

conftest.py
```diff
@@ -67,11 +68,14 @@ def pytest_configure(config):
...
     Configures pytest with the plotting and longrun command line options.
     """
-    if config.getoption("--plotting-on"):
-        tests.PLOTTING = True
-    else:
+    if not config.getoption("--plotting-on"):
         # We're not displaying plots so use a display-less backend
         mpl.use("Agg")
+        # Disable CAD viewer by mocking out QApplication.exec
+        from PySide2.QtWidgets import QApplication
```
je-cook, line 75: I dont know if the test error stems from here but you might need to change the backend to 'offscreen'

je-cook (commented):
Generally looks like a nice improvement.

Ive had a few thoughts about having a helper class that we inherit from to do some of the plotting setup and teardown.

In my head it would clean up a lot of the writing of `fig, ax = subplots()`, `plt.show` and `plt.close`. There would probably need to be one that does just `setup_class` and `teardown_class` and one that does `setup` and `teardown`.

What do you think?

tests/balance_of_plant/test_plotting.py
```diff
@@ -31,6 +31,9 @@
 
 
 class TestSuperSankey:
+    def teardown_method(self):
+        plt.close()
```
je-cook, line 35: with `--plotting-on` locally I never see the plot

je-cook (approved):
I will add the plotting test helper function stuff to a nice to have issue.

Just add a show to the sankey and I'm happy

PR #1307, opened 2022-08-04
Cryostat with new builder 

je-cook:
## Linked Issues

e.g. Closes #{ID}

## Description

## Interface Changes

## Checklist

### DIFF
--- a/bluemira/builders/cryostat.py
+++ b/bluemira/builders/cryostat.py
@@ -1,0 +1,150 @@
+# bluemira is an integrated inter-disciplinary design tool for future fusion
+# reactors. It incorporates several modules, some of which rely on other
+# codes, to carry out a range of typical conceptual fusion reactor design
+# activities.
+#
+# Copyright (C) 2021 M. Coleman, J. Cook, F. Franza, I.A. Maione, S. McIntosh, J. Morris,
+#                    D. Short
+#
+# bluemira is free software; you can redistribute it and/or
+# modify it under the terms of the GNU Lesser General Public
+# License as published by the Free Software Foundation; either
+# version 2.1 of the License, or (at your option) any later version.
+#
+# bluemira is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with bluemira; if not, see <https://www.gnu.org/licenses/>.
+
+"""
+Cryostat builder
+"""
+
+
+class Cryostat(ComponentManager):
+    pass
+
+
+class CryostatDesignerParams(ParameterFrame):
+    x_g_support: Parameter[float]
+    x_gs_kink_diff: Parameter[float]  # TODO add to Parameter default = 2
+    g_cr_ts: Parameter[float]
+    tk_cr_vv: Parameter[float]
+    well_depth: Parameter[float]  # TODO add to Parameter default = 5 chickens
+    z_gs: Parameter[
+        float
+    ]  # TODO add to Parameter default (z gravity support) = -15 chickens
+
+
+class CryostatBuilderParams(ParameterFrame):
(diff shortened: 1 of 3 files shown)


bluemira/builders/cryostat.py
```diff
@@ -0,0 +1,200 @@
...
+        Build the cryostat component.
+        """
+        self.x_out, self.z_top = self.designer.run()
+
+        # possibly just
+        # return Cryostat(
+        #     super().build(
+        #         xz=[self.build_xz()], xy=[self.build_xy()], xyz=self.build_xyz()
+        #     )
+        # )
```
hsaunders1904, line 130: My personal preference is to have each builder do their own component stacking, as you've done below. I think it's more explicit and saves having to navigate to the base class to see what's going on. 

Another option is a utility method on the base class; at least that can have a name that's clear about what's going on.

bluemira/builders/cryostat.py
```diff
@@ -0,0 +1,200 @@
...
+    """
+
+    CRYO = "Cryostat VV"
+    param_cls: Type[CryostatBuilderParams] = CryostatBuilderParams
+
+    def build(self) -> Cryostat:
+        """
+        Build the cryostat component.
+        """
+        self.x_out, self.z_top = self.designer.run()
```
hsaunders1904, line 123: In the plasma builder, I've made these local variables and passed them into the `build_*` methods. That way, it's not possible to call, e.g., `build_xz` before `self.x_out` is defined.

tests/builders/test_cryostat.py
```diff
@@ -0,0 +1,69 @@
...
+    @classmethod
+    def setup_class(cls):
+        cls.params = {
+            "g_cr_ts": {"name": "g_cr_ts", "value": 0.3},
+            "n_TF": {"name": "n_TF", "value": 16},
+            "tk_cr_vv": {"name": "tk_cr_vv", "value": 0.3},
+            "well_depth": {"name": "well_depth", "value": 5},
+            "x_g_support": {"name": "x_g_support", "value": 13},
+            "x_gs_kink_diff": {"name": "x_gs_kink_diff", "value": 2},
+            "z_gs": {"name": "z_gs", "value": -15},
```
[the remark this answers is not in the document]
je-cook: Completely agree. I'm sure there are nicer ways than this to set things up too.

bluemira/builders/cryostat.py
je-cook: I had a go with the utility method. In this specific case I'm unsure whether it is cleaner and I'm still leaning towards the below at the moment:

````python 
       component = super().build(
            xz=[xz_cryostat],
            xy=[self.build_xy(x_out)],
            xyz=self.build_xyz(xz_cross_section),
        )
```

tests/builders/test_tools.py
```diff
@@ -44,6 +45,37 @@
...
+    @pytest.mark.parametrize("ttl, sector_degree", zip(np.arange(1, 16), sector_degree))
+    @pytest.mark.parametrize("degree", np.arange(0, 361, step=60))
+    def test_get_n_sectors(self, degree, ttl, sector_degree):
+        s_deg, n_sec = get_n_sectors(ttl, degree)
+        assert np.isclose(s_deg, sector_degree)
+        try:
+            assert np.isclose(degree // sector_degree, n_sec)
+        except AssertionError:
+            # a sector at both ends
+            assert np.isclose((degree // sector_degree) + 1, n_sec)
```
hsaunders1904, line 76: Is this basically doing the same calculations as the function it's testing? Could we consider just calculating a few examples by hand and checking those? I think that's the commonly done thing, rather than duplicating the logic under test. Not a big deal though

je-cook: Wrong PR :) its isolated in #1327. I was fuzzing it a bit but yeah I can do that

je-cook: done in [2cc6a60](https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/1327/commits/2cc6a603b999176dcaba7a06c12ed8cd0464a181)

hsaunders1904 (approved):

PR #1308, opened 2022-08-04
Thermal Shield with new builders

je-cook:
## Linked Issues

e.g. Closes #{ID}

## Description

## Interface Changes

## Checklist

### DIFF
--- a/bluemira/builders/thermal_shield.py
+++ b/bluemira/builders/thermal_shield.py
@@ -36,7 +36,6 @@ from bluemira.base.parameter_frame import NewParameterFrame as ParameterFrame
 from bluemira.base.parameter_frame import parameter_frame
 from bluemira.builders.tools import (
     circular_pattern_component,
-    directional_component_tree,
     find_xy_plane_radii,
     get_n_sectors,
     make_circular_xy_ring,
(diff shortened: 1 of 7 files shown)


je-cook:
all of the topology fixes also live in #1348 otherwise this wont work,the vvts is also just spun through 359 because of #1347 with the code commented for use when fixed

bluemira/builders/thermal_shield.py
```diff
@@ -0,0 +1,356 @@
...
+    """
+
+    VVTS = "VVTS"
+    param_cls: Type[VVTSBuilderParams] = VVTSBuilderParams
+
+    def build(self) -> VacuumVesselThermalShield:
+        """
+        Build the vacuum vessel thermal shield component.
+        """
+        koz = self.designer.run()
```
hsaunders1904, line 106: I've been adding an `__init__` in all concrete builders `Builder`s, partially for typing the designer, but mainly because the `designer` argument in the base class is optional, but in the concrete classes it's either needed or it's not.

bluemira/base/designer.py
```diff
@@ -57,7 +57,11 @@ def run(self) -> _DesignerReturnT:
         pass
 
     def _init_params(self, params: Union[Dict, _ParameterFrameT]) -> _ParameterFrameT:
-        if isinstance(params, dict):
+        if params is None:
+            if self.param_cls is None:
+                return None
+            return self.param_cls.from_dict({})
+        elif isinstance(params, dict):
```
hsaunders1904, line 64: Do you think we should raise in the case of `params is None and param_cls is not None`? I think all it will mean is an error being raised somewhere else down the line.

In any case, this logic should be consistent with the logic in `Builder._init_params`, so one of them is going to need updating. We should probably refactor the logic out - needn't be in this PR though

je-cook: I was thinking about whether someone might want fallback defaults in their specific parameter class (therefore no error will be raised). I suppose that should really not be allowed and they should just update their config file

bluemira/builders/thermal_shield.py
je-cook: sure, done

je-cook:
if this waits for #1349 it will be a bit cleaner

hsaunders1904 (approved):

PR #1349, opened 2022-08-17
♻️ Refactor out some builder and designer methods

je-cook:
## Description

Refactor out some duplicate code. Unsure on the name and unsure whether I need to inherit from `abc` explicitly for `Designer` and `Builder`.

## Interface Changes

## Checklist

### DIFF
--- a/bluemira/base/parameter_frame/_frame.py
+++ b/bluemira/base/parameter_frame/_frame.py
@@ -98,10 +98,10 @@ class NewParameterFrame:
         return cls.from_dict(json.loads(json_in))
 
 
-def parameter_setup(
+def make_parameter_frame(
     params: Union[Dict, NewParameterFrame, None],
     param_cls: Type[NewParameterFrame],
-) -> Union[Type[NewParameterFrame], None]:
+) -> Union[NewParameterFrame, None]:
     """
     Helper function to generate a `ParameterFrame` of a specific type
 
--- a/bluemira/base/parameter_frame/_frame.py
+++ b/bluemira/base/parameter_frame/_frame.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import abc
 import copy
 import json
 from dataclasses import dataclass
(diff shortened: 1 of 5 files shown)


bluemira/base/parameter_frame/_frame.py
```diff
@@ -96,3 +97,45 @@ def from_json(cls, json_in: Union[str, json.SupportsRead]) -> NewParameterFrame:
                 return cls.from_dict(json.load(f))
         # load from a JSON string
         return cls.from_dict(json.loads(json_in))
+
+
+class ParameterSetup(abc.ABC):
```
hsaunders1904, line 102: Can't say I'm mad about using a class here. I feel like this could be a simple `make_parameterframe(cls, params)` (or similar) function that both the builder and designer call. I'd like to avoid the case we had with the previous builders where we need to go multiple inheritance levels deep before we see where `.params` is set. 

I think having a simple factory function is more flexible, and it makes more sense as something that lives in the `parameter_frame` module.

bluemira/base/parameter_frame/_frame.py
```diff
@@ -96,3 +97,45 @@ def from_json(cls, json_in: Union[str, json.SupportsRead]) -> NewParameterFrame:
...
+    ) -> Union[NewParameterFrame, None]:
+        if params is None:
+            if self.param_cls is None:
+                # Case for where there are no parameters associated with the object
+                return params
+            return self.param_cls.from_dict({})
+        elif isinstance(params, dict):
+            return self.param_cls.from_dict(params)
+        elif isinstance(params, NewParameterFrame):
+            return params
```
hsaunders1904, line 138: My mistake, but do you think we should have the below here?

```python
elif isinstance(params, param_cls):
    return params
elif isinstance(params, NewParameterFrame):
    return param_cls.from_frame(params)
```
Otherwise we could end up with the wrong type of `ParameterFrame`

hsaunders1904 (changes requested):

je-cook:
Ok take2?

bluemira/base/parameter_frame/_frame.py
```diff
@@ -96,3 +96,33 @@ def from_json(cls, json_in: Union[str, json.SupportsRead]) -> NewParameterFrame:
...
+    param_cls: Type[NewParameterFrame]
+        The `ParameterFrame` class to generate
+
+
+    """
+    if param_cls is None:
+        if params is None:
+            # Case for where there are no parameters associated with the object
+            return params
+        raise ValueError("Cannot process parameters, 'param_cls' is None.")
```
je-cook, line 121: Turns out to be a larger amount of effort to not error here

bluemira/base/parameter_frame/_frame.py
je-cook: Done

bluemira/base/parameter_frame/_frame.py
```diff
@@ -96,3 +96,33 @@ def from_json(cls, json_in: Union[str, json.SupportsRead]) -> NewParameterFrame:
                 return cls.from_dict(json.load(f))
         # load from a JSON string
         return cls.from_dict(json.loads(json_in))
+
+
+def parameter_setup(
+    params: Union[Dict, NewParameterFrame, None],
+    param_cls: Type[NewParameterFrame],
+) -> Union[Type[NewParameterFrame], None]:
```
hsaunders1904, line 104: I think this should be `Union[NewParameterFrame, None]`.

I think we may need to do a bit of magic with the typing here - I suspect we'll need a `TypeVar`. Don't worry about that, though - I've got a branch that's attempting to fix all this typing stuff.

bluemira/base/parameter_frame/_frame.py
```diff
@@ -96,3 +96,33 @@ def from_json(cls, json_in: Union[str, json.SupportsRead]) -> NewParameterFrame:
                 return cls.from_dict(json.load(f))
         # load from a JSON string
         return cls.from_dict(json.loads(json_in))
+
+
+def parameter_setup(
```
hsaunders1904, line 101: I still feel this should be `make_parameter_frame` (or similar). It's pretty common throughout Bluemira to use `make_` for factory functions, and `parameter` is different to `parameter_frame`

je-cook: done

hsaunders1904 (approved):
Good to go once tests agree

hsaunders1904:
Test failure is a sporadic optimisation failure - since this is a feature branch, I'm just going to merge

PR #1386, opened 2022-08-31
Refactor Designer

hsaunders1904:
## Description

Updates the `Designer` interface to include a mandatory `run` method, and optional `read` and `mock` modes.

Makes `Builder` and the base `Builders` agnostic of `Designer`s - this means we can remove all the in-out designers we were having to define.

I haven't looked at updating the EUDEMO designers yet - maybe we should do those separately.


je-cook (approved):
Looks good to me, yeah we can do the eudemo ones separately

PR #1366, opened 2022-08-23
Write/refactor EUDEMO equlibrium solver

hsaunders1904:
## Description

Refactors the equilibrium calculations out of the EUDEMO plasma builder, and puts it in a `Solver`.

The solver felt a bit 'boilerplate-ey' whilst I was writing it. We may consider whether we need the setup/run/teardown when we're not running external codes - the split can feel a bit arbitrary when you're just writing normal code and not calling out to an external process. The base solver could just be something that has an `execute` with a run-mode and can parse some params + build_config. Potentially not worth worrying about right now, though.

## Checklist

I confirm that I have completed the following checks:

- [ ] Tests run locally and pass `pytest tests --reactor`
- [ ] Code quality checks run locally and pass `flake8` and `black .`
- [ ] Documentation built locally and checked `sphinx-build -W documentation/source documentation/build`

PR #1388, opened 2022-09-01
EUDEMO equilibrium designer

hsaunders1904:
## Description

Adds designer for EUDEMO reactor equilibrium. Takes basically everything in #1366 and puts it in a `Designer` instead of a `Solver`.

I'm not really sure what name this designer should have. I went with `EquilibriumDesigner` as I guess it is **the** equilibrium designer for the EUDEMO example - but I don't know if we could be more specific. This is probably fine, though,


examples/design/EU-DEMO/EUDEMO_tests/equilibria/test_tools.py
```diff
@@ -3,7 +3,7 @@
 # codes, to carry out a range of typical conceptual fusion reactor design
 # activities.
 #
-# Copyright (C) 2022 M. Coleman, J. Cook, F. Franza, I.A. Maione, S. McIntosh, J. Morris,
+# Copyright (C) 2021 M. Coleman, J. Cook, F. Franza, I.A. Maione, S. McIntosh, J. Morris,
```
je-cook, line 6:
```suggestion
# Copyright (C) 2022 M. Coleman, J. Cook, F. Franza, I.A. Maione, S. McIntosh, J. Morris,
```

pyproject.toml
```diff
@@ -30,7 +30,7 @@ exclude = '''
 line_length = 89
 profile = "black"
 skip_gitignore = true
-known_first_party = ["bluemira"]
+known_first_party = ["bluemira", "EUDEMO_builders"]
```
je-cook, line 33: As you're fixing this "EUDEMO_tests" here too? eg if there are any imports between the tests

je-cook (changes requested):
Looks good, a few things:

Are you expecting that there is some mechanism separate to this that put the updated parameters back into the 'global' parameterframe for use in subsequent steps?

Is there a need for another eqdsk? Can we just add one when this gets separated out into its own repo (maybe symlink for now...?). It seems like a lot of lines that will be removed. Maybe I'm being overly cautious

hsaunders1904:
> Are you expecting that there is some mechanism separate to this that puts the updated parameters back into the 'global' parameterframe for use in subsequent steps?

My intention is that these `ParameterFrame`s are just passing around the same `Parameter` objects. So changes within the designer should propagate to the higher level frames. We definitely need some tests to this effect, though. 

Did you have something else in mind? There may well be some other way that's more explicit.

pyproject.toml
hsaunders1904: I can do - we don't do this for the `bluemira` tests, though.

je-cook: Fair point, I see you've changed it now either way. Leave it changed for now and we can remove it when this is a separate repo

je-cook:
> My intention is that these `ParameterFrame`s are just passing around the same `Parameter` objects. So changes within the designer should propagate to the higher level frames. We definitely need some tests to this effect, though.
> 
> Did you have something else in mind? There may well be some other way that's more explicit.

That sounds good to me, any parameterframe is then ephemeral but the parameter itself is not. I dont think it needs to be more explicit but we have to make users aware that after the conversion from the input json you should pass around parameterframe objects and not `dict`s to get that functionality.

PR #1400, opened 2022-09-06
🐛 fix enum specified inner and remove limitations

je-cook:
## Description

`inner` PictureFrame parameterisation adjustment doesnt work if directly specified with enum. This fixes it and adds some tests to that effect.

Also removed limitation so you can now have tapered inner with either domed or flat sections

## Interface Changes

None

## Checklist

### DIFF
--- a/tests/geometry/test_parameterisations.py
+++ b/tests/geometry/test_parameterisations.py
@@ -190,34 +190,66 @@ class TestPictureFrame:
 
 class TestComplexPictureFrame:
     @pytest.mark.parametrize(
-        "upper, lower, result",
+        "upper, lower, inner, result",
         [
-            ["CURVED", "CURVED", 57.6308],
-            ["CURVED", "FLAT", 56.829],
-            ["FLAT", "CURVED", 56.829],
-            [PFrameSection.CURVED, PFrameSection.CURVED, 57.6308],
-            [PFrameSection.CURVED, PFrameSection.FLAT, 56.829],
-            [PFrameSection.FLAT, PFrameSection.CURVED, 56.829],
+            ["CURVED", "CURVED", "TAPERED_INNER", 56.331],
+            ["CURVED", "FLAT", "TAPERED_INNER", 54.714],
+            ["FLAT", "CURVED", "TAPERED_INNER", 54.714],
+            [
+                PFrameSection.CURVED,
+                PFrameSection.CURVED,
+                PFrameSection.TAPERED_INNER,
+                56.331,
+            ],
+            [
+                PFrameSection.CURVED,
+                PFrameSection.FLAT,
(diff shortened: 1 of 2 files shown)


tests/geometry/test_parameterisations.py
```diff
@@ -189,23 +190,66 @@ def test_ordering(self):
...
+        [
+            # should these pass?
+            # ["FLAT", "FLAT", "TAPERED_INNER"],
+            # ["CURVED", "CURVED", "TAPERED_INNER"],
+            # ["CURVED", "FLAT", "TAPERED_INNER"],
+            # ["FLAT", "CURVED", "TAPERED_INNER"],
+            # [PFrameSection.FLAT, PFrameSection.FLAT, PFrameSection.TAPERED_INNER],
+            # [PFrameSection.CURVED, PFrameSection.CURVED, PFrameSection.TAPERED_INNER],
+            # [PFrameSection.CURVED, PFrameSection.FLAT, PFrameSection.TAPERED_INNER],
+            # [PFrameSection.FLAT, PFrameSection.CURVED, PFrameSection.TAPERED_INNER],
```
je-cook, line 240: Should the tapered inner be tangent with the rest of the shape (it wasnt already because `["FLAT", "FLAT", "TAPERED_INNER"]` should pass and it doesnt locally

je-cook:
@CoronelBuendia the first two commits are an actual fix the third I noticed it wasnt too big of a jump to remove the limitation, my only question is the tangency one above, I'm inclined to put xfail on the commented out parameterisations if it is meant to be tangent because it currently isnt either way.

tests/geometry/test_parameterisations.py
CoronelBuendia: I freely confess to not being familiar with these variants on the `PictureFrame` parameterisations. If I look at #1287, I get a better idea but I still can't be certain.

It feels to me that if the inner leg is tapered and the top and bottom legs are flat, the joints between the two cannot be tangent. They will either be a perpendicular joint, or if the tapering doesn't end before the transition into the top/bottom, it will be an obtuse angle joint.

If, on the other hand, there is some transition radius intended between the tapered leg and the flat sections, then there should be circle arcs joining up the two inner leg and the top/bottom. In this case, I would expect a tangential join. From the implementation, I vaguely recall Shail having problems getting the circle arcs to line up, hence (possibly) why there is some discretisation of these transitions along the way. It's possible this discretisation is the reason why this test fails.

If you can confirm this I can think of a few possible fixes with primitive geometry.

hsaunders1904 (approved):
Considering we've had a request to get this in, I'm happy for this to be merged once there's an issue open to resolve the xfail tests

je-cook:
Issue #1610

PR #1402, opened 2022-09-07
Remove repeated name in param definitions

hsaunders1904:
## Description

Removes the requirement to have the `name` key when defining parameters in a dict or JSON file using the new ParameterFrame class.

### DIFF
--- a/bluemira/base/parameter_frame/_frame.py
+++ b/bluemira/base/parameter_frame/_frame.py
@@ -154,6 +154,12 @@ def make_parameter_frame(
 
     param_cls: Type[NewParameterFrame]
         The `ParameterFrame` class to create a new instance of.
+
+    Returns
+    -------
+    Union[ParameterFrame, None]
+        A frame of the type `param_cls`, or `None` if `params` and
+        `param_cls` are both `None`.
     """
     if param_cls is None:
         if params is None:
--- a/bluemira/base/parameter_frame/_frame.py
+++ b/bluemira/base/parameter_frame/_frame.py
@@ -150,7 +150,7 @@ def make_parameter_frame(
             * None:
                 For the case where no parameters are actually required.
                 This is intended for internal use, to aid in validation
-                of parameters in `Builder`s and `Designer`s.
+                of parameters in `Builder`\\s and `Designer`\\s.
 
     param_cls: Type[NewParameterFrame]
         The `ParameterFrame` class to create a new instance of.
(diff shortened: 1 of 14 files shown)


bluemira/base/parameter_frame/_frame.py
```diff
@@ -139,6 +145,8 @@ def make_parameter_frame(
         return param_cls.from_dict(params)
     elif isinstance(params, param_cls):
         return params
+    elif isinstance(params, str):
```
je-cook, line 148: The docstring type needs to be updated to reflect this. It would also be useful to add a note to the docstring about this one because its not immediately obvious that the string could be a json file as a string or a filename.

je-cook (changes requested):
I think it looks good, one minor. The test failures however look genuine and I think you've missed a couple of name keys.

bluemira/base/parameter_frame/_frame.py
```diff
@@ -64,10 +64,12 @@ def from_dict(
                 raise ValueError(f"Data for parameter '{member}' not found.")
 
             value_type = cls._validate_parameter_field(member)
-            kwargs[member] = NewParameter(**param_data, _value_types=value_type)
+            kwargs[member] = NewParameter(
+                name=member, **param_data, _value_types=value_type
+            )
```
je-cook, line 69: Possible extra addition allowing key value pair specification. We may not want to do this for other reasons (units, source etc) but just a thought
```suggestion
            if not isinstance(param_data, Dict):
                param_data = {"value": param_data}
            kwargs[member] = NewParameter(
                name=member, **param_data, _value_types=value_type
            )
```

bluemira/base/parameter_frame/_frame.py
```diff
@@ -117,18 +123,37 @@ def _validate_parameter_field(cls, field: str) -> Tuple[Type, ...]:
...
+                assigned to the new ParameterFrame's parameters. Note
+                that this makes no copies, so updates to parameters in
+                the new frame will propagate to the old, and vice versa.
+            * str:
+                The path to a JSON file, or, if the string starts with
+                '{', a JSON string.
+            * None:
+                For the case where no parameters are actually required.
+                This is intended for internal use, to aid in validation
+                of parameters in `Builder`s and `Designer`s.
```
je-cook, line 153: Looks good but sphinx does like the formatting, although I do :)

je-cook (approved):
Just a thought I had but good to go once tests pass

je-cook (approved):
Looks good!

Issue #1253, opened 2022-07-14
Replace scipy's `minimize` in equilibria/fixed_boundary/utilities::`calculate_plasma_shape_params` with a more robust solution 


CoronelBuendia:
## Description of issue / requirement to address

In order to find plasma extrema in the FEM fixed boundary solution, an optimisation approach is used. At present this uses scipy's `minimize` function, which is not a particularly robust solution and does not follow the standard in the rest of the code base.

This was a "prototype" solution to the problem, for which scipy is a useful too.

## Proposed solution
Replace the scipy optimisations in the FEM fixed boundary equilibrium module with `Optimisers` and add anaytical Jacobians (for robustness). 

Interestingly, this is not a speed issue, as we are only searching for 4 points and get a reasonably good starting guess from the mesh, so no need to implement a custom CG approach.

## Alternative solutions

Leave things as are, but the interface will likely change anyway when we move away from scipy 1.5.3. We've had a whole host of issues in the past with upgrading scipy due to API changes and a lot of stuff in scipy.optimize seemed to just behave differently (IIRC).

PR #1404, opened 2022-09-08
WIP: Improve psi axis finding in fixed boundary equilibrium

### LINKED ISSUE #1255: Improve calculation of `psi_axis` in the fixed boundary equilibrium solver.

At present, in `bluemira/equilibria/fem_fixed_boundary/fem_magnetostatic_2D.py::FemGradShafranovFixedBoundary` we do the following to get the magnetic flux on axis and at the boundary:

```python
    @property
    def psi_ax(self):
        """Poloidal flux on the magnetic axis"""
        return np.max(self.psi.vector()[:])

    @property
    def psi_b(self):
        """Poloidal flux on the boundary"""
        return np.min(self.psi.vector()[:])
```

Right now, we return the maximum from the mesh points, but for a coarse mesh this might be leading to oscillations of the magnetic axis. This could be slowing down convergence.

## Proposed solution

I reckon we can improve the convergence of the fixed boundary G-S solve by using optimisation to find the psi on axis more accurately. 

Much like we now do for finding the plasma shape parameters of the 95th flux surface, we could also use optimisation to find the maximum magnetic flux value in the function space.

I also recommend we cache the two above properties, to avoid repetition of calculations. This is not a bottleneck at all, but if we add in optimisation it will slow things down if called every time psi_norm is calculated.

## Alternative solutions
It does work as is!

## Additional context
Splitting this out as a separate issue because #1067 has been a long time coming and I don't want to slow it down any more than I already have. Plus the functionality therein already works fine. I will pick this up once the FEM stuff goes in.

CoronelBuendia:
Requires #1067 

## Linked Issues
Closes #1255 

## Description
Calculates the psi axis via optimisation rather than taking the maximum of the grid points in the fixed boundary equilibrium solver.

For the example `examples/equilibria/fem_fixed_boundary/equilibrium_example.py` on `ivanmaione/clean_fixed_boundary` exhibits slightly improved convergence behaviour:

image

There are two additional internal G-S iterations (outer iterations 1 and 4), but 7 fewer in total, as the last outer iteration is not required (in this example).

As for the 95th flux surface extremum finding, just using scipy `minimize` for now. Best to pick up #1253 on both counts in a separate future PR (not so urgent).

## Interface Changes

N/A

## Checklist


CoronelBuendia:
Superceded by #1455

PR #1406, opened 2022-09-08
Equilibria GIF bugfix

### LINKED ISSUE #1405: Equilibria GIF failing due to image size inconsistencies
## Describe the bug
When making a GIF from an equilibrium solve, with e.g.:
```python
program = PicardIterator(
    eq, current_opt_problem, fixed_coils=True, relaxation=0.2, plot=True, gif=True, figure_folder="/home/matti/code/"
)
program()
```
the images all save, but the GIF fails to produce with an error.
## Steps to reproduce

1. Using script 'examples/equilibria/single_null.py`
2. Change lines 241-244 to:
```python
program = PicardIterator(
    eq, current_opt_problem, fixed_coils=True, relaxation=0.2, plot=True, gif=True, figure_folder="/home/"
)
program()
```
4. See error:
```python
  File "examples/equilibria/single_null.py", line 244, in <module>
    program()
  File "/home/matti/code/bluemira/bluemira/equilibria/solve.py", line 498, in __call__
    next(iterator)
  File "/home/matti/code/bluemira/bluemira/equilibria/solve.py", line 542, in __next__
    make_gif(self.figure_folder, self.pname)
  File "/home/matti/code/bluemira/bluemira/utilities/plot_tools.py", line 127, in make_gif
    imageio.mimsave(gifname, images, "GIF", **kwargs)
  File "/home/matti/miniforge3/envs/bluemira/lib/python3.8/site-packages/imageio/v2.py", line 331, in mimwrite
    return file.write(ims, **kwargs)
  File "/home/matti/miniforge3/envs/bluemira/lib/python3.8/site-packages/imageio/core/legacy_plugin_wrapper.py", line 198, in write
    ndimage = np.stack(ndimage, axis=0)
  File "<__array_function__ internals>", line 180, in stack
  File "/home/matti/miniforge3/envs/bluemira/lib/python3.8/site-packages/numpy/core/shape_base.py", line 426, in stack
    raise ValueError('all input arrays must have the same shape')
ValueError: all input arrays must have the same shape
```

## Expected behaviour
I expect the GIF to be produced correctly.

## Screenshots / Tracebacks

See above

- OS: Ubuntu 20.04
- Branch (if developing) [e.g. develop]

## Additional context
Haven't used this in a while

CoronelBuendia:
## Linked Issues

Closes #1405 

## Description

Fixes the aforementioned bug by resizing the offending image.

Seems the `bbox_inches="tight"` isn't consistently sizing images down to a pixel when specified with a `dpi` argument. That's as far down the matplotlib rabbithole I went, before bailing and re-sizing with Pillow (existing dependency).

EDIT:
Also:
* default figure folder changed in-line with repo restructuring (`figure_folder=None` no longer raises an error)
* if `gif=True` but `plot=False` `plot` is set to `True` otherwise it doesn't work properly

## Interface Changes
None

## Checklist


bluemira/utilities/plot_tools.py
```diff
@@ -121,6 +122,17 @@ def make_gif(folder, figname, formatt="png", clean=True):
     if clean:
         for fp in ims:
             os.remove(fp)
+
+    if len(images) > 1:
+        # Occasionally bbox_tight is a few pixels off for contours it seems
+        # Resize the offending images
+        sizes = np.array([image.shape for image in images])
+        if not (sizes == sizes[0]).all():
+            shape = tuple(sizes[1][:2])
```
hsaunders1904, line 131: Not a problem, just curious, is there a reason we're using the size from the second image and not the first?

hsaunders1904 (approved):

bluemira/utilities/plot_tools.py
CoronelBuendia: Urgh.. yeah; the first is invariably the offending image. For reasons I don't understand...

je-cook:
For Pillow to be added as a dependency it needs to be added to `setup.py` and the pin moved to `requirements.txt` if it isnt already.
Is it possible to do this with imageio thats already a direct dependency?

CoronelBuendia:
> For Pillow to be added as a dependency it needs to be added to `setup.py` and the pin moved to `requirements.txt` if it isnt already. Is it possible to do this with imageio thats already a direct dependency?
> 
> Also from your traceback in #1405 it looks like we're using a legacy wrapper so maybe thats what needs updating: `imageio/core/legacy_plugin_wrapper.py`
> 
> Just trying to avoid more direct dependencies :)

I did try, but came across this: https://stackoverflow.com/questions/65733362/how-to-resize-an-image-using-imageio

I thought Pillow was a lesser evil than skimage, but I didn't know it wasn't a direct dependency...

CoronelBuendia:
Putting this back to draft as the GIFs look completely different on my home machine, so need to tackle the root cause it seems.

je-cook:
This may be helped by #1452 and the update to mpl 3.6 possibly?

CoronelBuendia:
OK, so I found the fecker. Seems that plt.pause(float) changes the figure in some cases, so now on my home machine it works... haven't tested at work yet!

Good news is now clipping required now, (PIL no longer needed)

bluemira/equilibria/solve.py
```diff
@@ -586,15 +588,14 @@ def update_fig(self):
         """
         self.ax.clear()
         self.eq.plot(ax=self.ax)
-        self.ax.figure.canvas.draw()
+        plt.pause(PLT_PAUSE)
```
hsaunders1904, line 591: What's the reason for the pause here? Could potentially do with a comment

hsaunders1904 (approved):
I wasn't able to recreate the issue initially, so I can't check if it's resolved. The changes all look fine to me, though!

bluemira/equilibria/solve.py
CoronelBuendia: This pause is actually moved up from a line below. It seems this is matplotlib standard practice (before they introduced animations) for updating plots. Introducing animations here is a little more of a refactor than just a simple bugfix though.

je-cook (approved):

PR #1433, opened 2022-09-16
🧑‍💻 Pretty print parameterframe

### LINKED ISSUE #1430: Create `__str__` method for new parameter frame

We should add a pretty printing of at least parameterframe (maybe parameter as well)

## Proposed solution

use `__str__` and tabulate as we have in the old parameterframe

## Alternative solutions

leave just `__repr__`

je-cook:
## Linked Issues

Closes #1430 

## Description

Pretty print of a parameterframe. I havent bothered with Parameter but that could be done too. Dunno if it would add much

## Interface Changes

None

## Checklist

### DIFF
--- a/tests/base/parameterframe/test_parameterframe.py
+++ b/tests/base/parameterframe/test_parameterframe.py
@@ -212,12 +212,13 @@ class TestParameterFrame:
                 except ValueError as ve:
                     if ind in head_keys:
                         raise ve
-            nas = [i for i, x in enumerate(tr) if x == "N/A"]
 
             # Number of 'N/A' equal to headers without name - number of filled keys
             no_name_headers = headers.copy()
             no_name_headers.pop(headers.index("name"))
-            assert len(nas) == len(no_name_headers - data_values[dvi].keys())
+            assert len([i for i, x in enumerate(tr) if x == "N/A"]) == len(
+                no_name_headers - data_values[dvi].keys()
+            )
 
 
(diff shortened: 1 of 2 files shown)


tests/base/parameterframe/test_parameterframe.py
```diff
@@ -182,6 +183,43 @@ def test_update_values_edits_frames_values(self):
         assert frame.age.value == 30
         assert frame.age.source != "a test"
 
+    @pytest.mark.parametrize("head_keys", [None, ["name", "value"]])
+    def test_tabulator_method_formatting(self, head_keys):
```
hsaunders1904, line 187: It's not so easy to understand what this test is doing. Would it be simpler to split this into two tests; one for each case? Then (I think) it's just a case of extracting the args/kwargs and directly asserting what we expect them to be, which should be much easier to parse.

It will be less DRY, but, in my view, tests should be specific and explicit, even if it means repeating yourself a bit.

hsaunders1904 (commented):
Looks good!
image

A couple of minor things

tests/base/parameterframe/test_parameterframe.py
je-cook: Sure :) I did have fun condensing that one :laughing:

je-cook: Hopefully thats easier to understand

Issue #1422, opened 2022-09-13
Replace uses of old `Parameter` and `ParameterFrame` with new versions


hsaunders1904:
## Description of issue / requirement to address

Replace uses of the old `ParameterFrame` with the new one.

This will potentially involve replacing some of the old classes' functionality. In particular, we'll need to deal with parameter mapping. I feel that the mappings shouldn't live on the `Parameter`, the functionality to map the parameters for a code should be defined within its `codes` module.

There aren't too many instances of the old parameter classes being used within bluemira (mostly in `codes`), but there are quite a few uses within examples - we should take care not to miss these!

hsaunders1904:
Closed by #1505

PR #1456, opened 2022-09-21
Separate eqdsk interface from parameterframe

hsaunders1904:
Part of #1422 

## Description

Remove `EQDSKInterface`'s dependence on `ParameterFrame` as part of replacing the old `ParameterFrame`.

Essentially, the `ParameterFrame` within the file interface has been replaces with a dataclass.

Included in this is a fix for a bug where the file interface was reading `psinorm` in as `psinorm`, but writing it out as `pnorm` in JSON files. Because of this, `psinorm` was just being read in as `None` from JSON files.

Note also that the existing test in `test_file` was not actually being run, because the mechanics for finding files to test on was returning an empty list. I've explicitly listed the test files now.

Another note: the changes to the eqdsk JSON files are only changing the key `pnorm` → `psinorm`. The whole file is on the same line though, so the diffs are impossible to read.

## Interface Changes

Instead of calling the reader like this,

```python
e = EQDSKInterface()
e.read("file")
```

you would call it like this

```python
e = EQDSKInterface.from_file("file")
```

The main reason for this is that we now only have defaults for fields that may, or may not, be in an eqdsk file (`qpsi`). So a default constructor no longer makes sense (the previous defaults were `None`, so we would have got an error as soon as those defaults were used anyway - we just get an error earlier now).

You also access the values of the `EQDSKInterface` using attributes (`eqdsk.qpsi`), rather than via `__getitem__` (`eqdsk["qpsi"]`).

## Checklist


hsaunders1904:
@CoronelBuendia I did have a question about [this test](https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/1456/files#diff-6880e179c826d77a919276571732e70f506c3709a89ee96f2a62a1fbb367dcb0R227-R228). I was expecting the assertion to pass, but it seems that the value for `psinorm` in the file does not match the default `psinorm` we calculate. Is this something we need to worry about? Is the file not self-consistent, or are we calculating `psinorm` wrong, or something else?

tests/equilibria/test_file.py
```diff
@@ -97,67 +112,118 @@ def read_flat_array(fortran_format, array_size):
...
+            del data[field]
+
+        open_mock = mock.mock_open(read_data=json.dumps(data))
+        with mock.patch(OPEN, new=open_mock):
+            eqdsk = EQDSKInterface.from_file("/some/file.json")
+
+        np.testing.assert_allclose(eqdsk.x, self.eudemo_sof_data["x"])
+        np.testing.assert_allclose(eqdsk.z, self.eudemo_sof_data["z"])
+        # TODO(hsaunders1904): understand why this fails
+        # np.testing.assert_allclose(eqdsk.psinorm, self.eudemo_sof_data["psinorm"])
```
CoronelBuendia, line 228: I can't remember why, or indeed if there was a reason, but when looking through our JSON files, it seems that the psinorm vector doesn't end in 1.0 (i.e. endpoint is discarded). In `_derive_psinorm` you have `np.linspace(0, 1, len(fpol))`, which would end in 1.0. I vaguely remember that we used not to be able to find the flux surface at psi_norm=1.0 very accurately using normal flux surface finding approaches. We since added the `find_LCFS_separatrix` which does the job. It could be because of this that we used to discard the 1.0 point in EQSDSKs. This feels like the kind of hack I might have 5 years ago... :D

hsaunders1904: This seems to be it (well remembered!). Calculating `psinorm` as `np.linspace(0, 1, len(fpol) + 1)[:-1]` makes the test pass. I've updated the test so that we're checking more than just the size

CoronelBuendia (approved):

PR #1464, opened 2022-09-23
Fixed boundary equilibrium flux surface finding and Solovev test improvements

CoronelBuendia:
## Linked Issues

e.g. Closes #{ID}

## Description

## Interface Changes

## Checklist


bluemira/equilibria/fem_fixed_boundary/utilities.py
```diff
@@ -210,6 +210,80 @@ def get_tricontours(x, z, array, value):
     return contours
 
 
+def find_flux_surface(psi_norm_func, psi_norm, mesh=None, n_points=100):
+    """
+    Find a flux surface in the psi_norm function precisely by normalised psi value.
+
+    Parameters
+    ----------
+    psi_norm_func: Callable
```
hsaunders1904, line 219: What form should this function take? You can give the function signature as part of the typing. If the function takes an int and a numpy array, and returns a float the signature would be:

```python
Callable[[int, np.ndarray], float]
```

But, further to the typing, a bit more in the docstring about what the inputs to the function should be could be useful.

bluemira/equilibria/fem_fixed_boundary/utilities.py
```diff
@@ -326,88 +400,43 @@ def find_extremum(func, x0):
...
-                [-(ro**4) / 8.0, 0.0],
-                [-(rt**4) / 8.0, +(zt**2) / 2.0],
-                [-(rt**2) / 2.0, 0.0],
-            ]
-        )
-        b = b * np.array([self.A1, self.A2])
-        b = np.sum(b, axis=1)
+    Parameters
+    ----------
+    psi_func: Callable
```
hsaunders1904, line 409: Ditto here - I guess it'll be the same form as above

bluemira/equilibria/fem_fixed_boundary/utilities.py
```diff
@@ -210,6 +210,80 @@ def get_tricontours(x, z, array, value):
...
+        bounds = [(d_guess - 2 * search_range, d_guess + 2 * search_range)]
+
+        def get_next_guess(points, distances, i):
+            return np.hypot(points[0, i - 1] - x_axis, points[1, i - 1] - z_axis)
+
+    else:
+        d_guess = 0.5
+        bounds = [(0.1, None)]
+
+        def get_next_guess(points, distances, i):
```
hsaunders1904, line 253: Looks like `points` isn't used here

hsaunders1904 (commented):
Just a couple of comments, otherwise looks good :+1:.

The new tests are quite beefy (take over minute to run on my laptop), potentially candidates for `@pytest.mark.longrun`? Or did you want the tests run regularly?

CoronelBuendia:
> Just a couple of comments, otherwise looks good +1.
> 
> The new tests are quite beefy (take over minute to run on my laptop), potentially candidates for `@pytest.mark.longrun`? Or did you want the tests run regularly?

I'd like to keep it for now. Especially with the move to Fenics-X in the not too distant future. Once we're happy we're always doing the same thing, we can put it into longrun / and or run it at lower resolution. On my desktop it is more like 20+ seconds

bluemira/equilibria/fem_fixed_boundary/utilities.py
CoronelBuendia: Yes this is because of the rather weird requirement to have a different function for different cases, which take different arguments. So I just overloaded them both with different unused arguments. I've added a noqa but happy to improve if you have a better approach!

je-cook (approved):

je-cook:
I will rebase this once #1455 goes in

bluemira/equilibria/fem_fixed_boundary/utilities.py
je-cook: I don't have a better option at the moment. Merging and we may come back to it in the future

PR #1465, opened 2022-09-23
Separate fuel cycle and ParameterFrame

hsaunders1904:
Part of #1422 

## Description

Removes fuel cycle's dependence on `ParameterFrame`.

The fuel cycle model currently in the repo seems to be EUDEMO specific - should this be a solver/designer in the EUDEMO example repo? 

I was a bit unsure of these changes to begin with. We're no longer passing a reference to a set of parameters to the model/life cycle, so the parent parameters are not "updated" as they used to be. I'm not familiar enough with this module to know if that's a big problem in this case; but in general, replacing `ParameterFrame` with dicts/dataclasses loses this feature. However, I think this may be the right approach. The intention being that the modules are independent of any reactor architecture, but are used by solvers/designers, which own `ParameterFrame`s. The solvers/designers are then responsible for parameter bookkeeping, which means the module code can be more standalone.

## Interface Changes

`LifeCycle` and `EUDEMOFuelCycleModel` now take a dict or dataclass as parameter inputs, instead of a `ParameterFrame`.

## Checklist


je-cook (approved):
This whole module has sort of been left behind (the test coverage is ~12%). I agree we should separate out the EUDEMO specific stuff into something else. 

Either way this seem like a good start for now to at least make it standalone

PR #1482, opened 2022-09-29
📝 change badge to release docs

je-cook:
## Description

We should probably link to the release docs rather than dev docs from the badge. Unfortunately I've never found a way to do it dynamically 

## Interface Changes

## Checklist


CoronelBuendia:
What about code coverage? Should that be latest too?

je-cook:
I was wondering about that, develop will always be higher and I would say coverage will be more widely used by devs... I have no strong preference, thoughts anyone?

CoronelBuendia:
> I was wondering about that, develop will always be higher and I would say coverage will be more widely used by devs... I have no strong preference, thoughts anyone?

OK, fair does

CoronelBuendia (approved):

PR #1487, opened 2022-09-29
✨ Local minima with plateau

je-cook:
## Description

I found this buried in a stash. I don't know if we want it or not. I can clean it up if we do. Finds local minima of a function including plateaus

## Interface Changes

None

## Checklist

### DIFF
--- a/bluemira/equilibria/find.py
+++ b/bluemira/equilibria/find.py
@@ -27,6 +27,8 @@ import operator
 
 import numba as nb
 import numpy as np
+import scipy.ndimage.filters as filters
+import scipy.ndimage.morphology as morphology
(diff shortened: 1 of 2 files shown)


CoronelBuendia:
So I've given this some thought. I'm quite confident that non-zero currents on more than one coil it's nearly impossible to have a plateau of minima (in part, thanks to machine (im)precision).

At this stage, I don't think it is worth switching approach, because it has been relatively robust for a couple of years at least.

That said, I think it's worth holding on to this, just in case we need to revisit. Could you add it as a gist on the repo (https://github.com/Fusion-Power-Plant-Framework/bluemira/discussions/600)

je-cook:
Sure will do, I'll close when I've done that

PR #1504, opened 2022-10-03
✨ Add unit conversions back to parameter frame

### LINKED ISSUE #1421: Implement unit management in new ParameterFrame

The new `Parameter` class stores a unit string, but does not perform any sort of unit conversion/validation. We should add functionality to the `Parameter` class such that it can perform unit conversions on the fly.

## Proposed solution

The following should be considered/implemented:

- Use `pint` as a backend to validate that units strings are valid units.
- Make `unit` a compulsory constructor argument in `Parameter`.
- Convert the parameter value to a standard Bluemira unit in the constructor.
    - This actually might be better to do in `ParameterFrame`, as `Parameter` is a bit more general. So have a list of standard units in `ParameterFrame` and perform unit conversions before the frame's parameters are constructed.
- Add a `value_as_unit` method to retrieve values in a specific unit.
- Add `unit` argument to `set_value` method.
    - So we can set parameter values in non-standard units.

je-cook:
## Linked Issues

Closes #1421 

## Description

## Interface Changes

## Checklist

### DIFF
--- a/bluemira/base/constants.py
+++ b/bluemira/base/constants.py
@@ -121,8 +121,8 @@ class BMUnitRegistry(UnitRegistry):
             e_to_t,
             t_units,
             ev_units,
-            lambda ureg, x: x * conversion,
-            lambda ureg, x: x / conversion,
+            lambda _, x: x * conversion,
+            lambda _, x: x / conversion,
         )
 
     @property
@@ -142,7 +142,7 @@ class BMUnitRegistry(UnitRegistry):
         return self._gas_flow_temperature
 
     @gas_flow_temperature.setter
-    def gas_flow_temperature(self, value: [float, Quantity]):
+    def gas_flow_temperature(self, value: Union[float, Quantity]):
         self._gas_flow_temperature = (
             value.to("kelvin")
             if isinstance(value, Quantity)
--- a/bluemira/base/constants.py
+++ b/bluemira/base/constants.py
@@ -23,7 +23,7 @@
 A collection of generic physical constants, conversions, and miscellaneous constants.
 """
 
-from typing import List, Optional, Union
+from typing import Callable, List, Optional, Union
 
 import numpy as np
 from periodictable import elements
@@ -114,13 +114,47 @@ class BMUnitRegistry(UnitRegistry):
         t_units = "[temperature]"
         ev_units = "[energy]"
 
-        return self._transform(e_to_t, t_units, ev_units, self.Quantity("k_B"))
+        conversion = self.Quantity("k_B")
+
+        return self._transform(
+            e_to_t,
+            t_units,
+            ev_units,
+            lambda ureg, x: x * conversion,
+            lambda ureg, x: x / conversion,
+        )
+
+    @property
+    def flow_conversion(self):
+        """Gas flow conversion factor R * T"""
+        return self.Quantity("molar_gas_constant") * self.gas_flow_temperature
+
+    @property
+    def gas_flow_temperature(self):
+        """
+        Gas flow temperature in kelvin
+
+        If Quantity provided to setter it will convert units (naïvely)
+        """
+        if not hasattr(self, "_gas_flow_temperature"):
+            self._gas_flow_temperature = self.Quantity(0, "celsius").to("kelvin")
+        return self._gas_flow_temperature
+
+    @gas_flow_temperature.setter
+    def gas_flow_temperature(self, value: [float, Quantity]):
+        self._gas_flow_temperature = (
+            value.to("kelvin")
+            if isinstance(value, Quantity)
+            else self.Quantity(value, "kelvin")
+        )
 
     def _flow_context(self):
         """
         Convert between flow in mol and Pa m^3
 
-        Pa m^3 = R * 0degC * mol
+        Pa m^3 = R * temperature * mol
+
+        https://en.wikipedia.org/wiki/Standard_temperature_and_pressure#Molar_volume_of_a_gas
 
         Returns
         -------
@@ -132,18 +166,25 @@ class BMUnitRegistry(UnitRegistry):
         mol_units = "[substance]"
         pam3_units = "[energy]"
 
-        conversion_factor = self.Quantity("molar_gas_constant") * self.Quantity(
-            0, "celsius"
-        ).to("kelvin")
-
-        return self._transform(mols_to_pam3, mol_units, pam3_units, conversion_factor)
+        return self._transform(
+            mols_to_pam3,
+            mol_units,
+            pam3_units,
+            lambda ureg, x: x * ureg.flow_conversion,
+            lambda ureg, x: x / ureg.flow_conversion,
+        )
 
     @staticmethod
     def _transform(
         context: Context,
         units_from: str,
         units_to: str,
-        conversion: Union[float, int, complex, Quantity],
+        forward_transform: Callable[
+            [UnitRegistry, Union[float, int, complex, Quantity]], float
+        ],
+        reverse_transform: Callable[
+            [UnitRegistry, Union[float, int, complex, Quantity]], float
+        ],
     ):
 
         formatters = ["{}", "{} / [time]"]
@@ -151,14 +192,10 @@ class BMUnitRegistry(UnitRegistry):
         for form in formatters:
 
             context.add_transformation(
-                form.format(units_from),
-                form.format(units_to),
-                lambda ureg, x: x * conversion,
+                form.format(units_from), form.format(units_to), forward_transform
             )
             context.add_transformation(
-                form.format(units_to),
-                form.format(units_from),
-                lambda ureg, x: x / conversion,
+                form.format(units_to), form.format(units_from), reverse_transform
             )
 
         return context
@@ -321,6 +358,39 @@ def raw_uc(
             ureg.Quantity(value * unit_from_q).to(unit_to_q.units).magnitude
             / unit_to_q.magnitude
         )
+
+
+def gas_flow_uc(
+    value: Union[int, float, np.ndarray],
+    unit_from: Union[str, ureg.Unit],
+    unit_to: Union[str, ureg.Unit],
+    gas_flow_temperature: Optional[Union[float, int, complex, Quantity]] = None,
+) -> Union[int, float, np.ndarray]:
+    """
+    Converts around Standard temperature and pressure for gas unit conversion.
+    Accurate for Ideal gases.
+
+    https://en.wikipedia.org/wiki/Standard_temperature_and_pressure#Molar_volume_of_a_gas
+
+    Parameters
+    ----------
+    value: Union[int, float, np.array]
+        value to convert
+    unit_from: Union[str, Unit]
+        unit to convert from
+    unit_to: Union[str, Unit]
+        unit to convert to
+    gas_flow_temperature: Optional[Union[float, int, complex, Quantity]]
+        Gas flow temperature if not provided is 273.15 K,
+        if units not given kelvin assumed
+
+    Returns
+    -------
+    converted value
+    """
+    if gas_flow_temperature is not None:
+        ureg.gas_flow_temperature = gas_flow_temperature
+    return raw_uc(value, unit_from, unit_to)
 
 
 def to_celsius(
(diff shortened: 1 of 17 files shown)


bluemira/base/constants.py
```diff
@@ -132,33 +167,36 @@ def _flow_context(self):
...
     def _transform(
         context: Context,
         units_from: str,
         units_to: str,
-        conversion: Union[float, int, complex, Quantity],
+        forward_transform: Callable[
+            [UnitRegistry, Union[float, int, complex, Quantity]], float
+        ],
+        reverse_transform: Callable[
+            [UnitRegistry, Union[float, int, complex, Quantity]], float
```
hsaunders1904, line 187: `float` encompasses `int`, so is not strictly necessary here

bluemira/base/parameter_frame/_frame.py
```diff
@@ -115,14 +186,133 @@ def to_dict(self) -> Dict[str, Dict[str, Any]]:
...
             not hasattr(member_type, "__origin__")
-            or member_type.__origin__ is not NewParameter
+            or member_type.__origin__ is not Parameter
         ):
-            raise TypeError(f"Field '{field}' does not have type NewParameter.")
+            raise TypeError(f"Field '{field}' does not have type Parameter.")
         value_types = get_args(member_type)
         return value_types
 
+    @classmethod
```
hsaunders1904, line 134: Does this one need to be a `classmethod`?

bluemira/base/parameter_frame/_frame.py
```diff
@@ -115,14 +186,133 @@ def to_dict(self) -> Dict[str, Dict[str, Any]]:
...
+
+        if unit != quantity.units:
+            val = raw_uc(quantity.magnitude, quantity.units, unit)
+            if isinstance(param_data["value"], int) and int in value_type:
+                val = int(val)
+            param_data["value"] = val
+
+        param_data["unit"] = unit
+
+    @staticmethod
```
hsaunders1904, line 154: This is probably more of a personal preference, but I would usually make a private static method a free function. There's no functional difference, but it doesn't expose the methods to users of the class and makes the class definition shorter and cleaner.

bluemira/base/parameter_frame/_frame.py
```diff
@@ -115,14 +123,133 @@ def to_dict(self) -> Dict[str, Dict[str, Any]]:
...
+            new_unit = modified_unit
+
+        # Deal with angles
+        if ang_unit:
+            return cls._fix_angle_units(new_unit, orig_unit, ang_unit)
+        else:
+            return new_unit
+
+    @staticmethod
+    def _fix_angle_units(modified_unit, orig_unit, angle_unit):
```
hsaunders1904, line 237: This could probably do with some documentation on what we mean by 'fix' and why we need to 'fix' angles.

bluemira/base/parameter_frame/_frame.py
```diff
@@ -115,14 +123,133 @@ def to_dict(self) -> Dict[str, Dict[str, Any]]:
...
+        breaking_units = ["steradian", "square_degree"]
+        new_angle_unit = base_unit_defaults["[angle]"]
+        unit_str = f"{orig_unit:C}"
+        for b in breaking_units:
+            if b in unit_str:
+                raise NotImplementedError(
+                    f"{breaking_units} not supported for conversion"
+                )
+        if f"{angle_unit}**" in unit_str:
+            raise NotImplementedError("Exponent angles are not supported")
```
hsaunders1904, line 247: A bit pedantic, but what about `x**1` or `x**0`?

bluemira/base/parameter_frame/_parameter.py
```diff
@@ -133,8 +133,12 @@ def value(self) -> ParameterValueType:
     def value(self, new_value: ParameterValueType):
         self.set_value(new_value, source="")
 
+    def value_as(self, unit: Union[str, pint.Unit]) -> ParameterValueType:
+        """Return the current value in a given unit"""
+        return raw_uc(self.value, self.unit, unit)
+
     @property
-    def unit(self) -> str:
+    def unit(self) -> pint.Unit:
```
hsaunders1904, line 141: Are we sure we want to expose `pint` in this way? I thought our plan was just to expose strings, rather than objects, and use `pint` under-the-hood.

bluemira/base/parameter_frame/_frame.py
```diff
@@ -115,14 +123,133 @@ def to_dict(self) -> Dict[str, Dict[str, Any]]:
...
+        return pint.Unit(unit)
+
+    @classmethod
+    def _fix_weird_units(
+        cls, modified_unit: pint.Unit, orig_unit: pint.Unit
+    ) -> pint.Unit:
+        """
+        Essentially a crude unit parser for when we have no dimensions
+        or non-comutative dimensions
+        """
```
hsaunders1904, line 177: This function is quite complex - I think it could probably do with some more documentation and being split up a bit.

tests/base/parameterframe/test_parameterframe.py
```diff
@@ -253,6 +254,140 @@ def test_tabulate_method_columns_have_correct_data(self, head_keys):
...
+
+
+@dataclass
+class UnitFrame5(ParameterFrame):
+    wtf1: Parameter[float]
+    wtf2: Parameter[float]
+    wtf3: Parameter[float]
+
+
+class TestParameterFrameUnits:
```
hsaunders1904, line 296: Do we have a/some test(s) for unknown units? E.g., `{"value": 180.5, "unit": "not_a_known_unit"}`. What behaviour do we want? Maybe a warning?

hsaunders1904 (changes requested):
This is looking great. A lot of gotchas that I hadn't thought of... I haven't tried it out yet, but have a few comments to start

bluemira/base/parameter_frame/_parameter.py
je-cook: My thoughts were it would give us the flexibility to display it as we feel like (as a string: https://pint.readthedocs.io/en/stable/formatting.html?highlight=format#unit-format-specifications). I should modify this so the return type is a string though..

bluemira/base/parameter_frame/_frame.py
je-cook: yeah :). If I'm honest I meant to put this as draft until i'd fixed that :sweat_smile:

tests/base/parameterframe/test_parameterframe.py
je-cook: Could be a warning and conversion to dimensionless. Or could just error. I havent taken into account if the unit is the correct for that variable which I dont know might need to be coded somewhere.

bluemira/base/parameter_frame/_frame.py
je-cook: it could be directly a function now with some work I've just done but I set it up as a classmethod so ti can be called after `validate_parameter_field` in the `from_dict` classmethod

tests/base/parameterframe/test_parameterframe.py
hsaunders1904: Discussed offline: throw an error if `pint` does not recognise the unit, but allow any units `pint` recognises, even if we don't have a 'standard' unit conversion for it.

bluemira/base/parameter_frame/_frame.py
```diff
@@ -164,6 +163,151 @@ def __str__(self) -> str:
...
+
+    new_unit = _non_comutative_unit_conversion(
+        dict(modified_unit.dimensionality), unit_str.split("/")[0], dpa, fpy
+    )
+
+    # Deal with angles
+    return _convert_angle_units(new_unit, unit_str, ang_unit) if ang_unit else new_unit
+
+
+def _non_comutative_unit_conversion(dimensionality, numerator, dpa, fpy):
```
hsaunders1904, line 288: What's this doing? Just making sure `dpa` and `fpy` are at the start of the unit?

bluemira/base/parameter_frame/_parameter.py
```diff
@@ -133,10 +137,14 @@ def value(self) -> ParameterValueType:
...
```
hsaunders1904, line 145:
```suggestion
    def unit(self) -> str:
```

bluemira/base/parameter_frame/_parameter.py
```diff
@@ -133,10 +137,14 @@ def value(self) -> ParameterValueType:
     def value(self, new_value: ParameterValueType):
         self.set_value(new_value, source="")
 
+    def value_as(self, unit: Union[str, pint.Unit]) -> ParameterValueType:
+        """Return the current value in a given unit"""
+        return raw_uc(self.value, self.unit, unit)
```
hsaunders1904, line 142: We potentially need a try/catch to deal with cases where the value is not numeric here. We don't want to be exposing `pint` exceptions

bluemira/base/parameter_frame/_parameter.py
```diff
@@ -56,7 +56,7 @@ def __init__(
         self,
         name: str,
         value: ParameterValueType,
-        unit: str = "",
+        unit: Union[str, pint.Unit] = pint.Unit("dimensionless"),
```
hsaunders1904, line 59: I'm beginning to think we shouldn't be allowing `pint` types as inputs. If we do ever want to manage units differently, we'll be making breaking API changes if we move away from `pint`.

It may never be the case that we stop using `pint`, but I feel for something like this it's better practice to encapsulate the way we're managing units. What do you think? Maybe I'm missing an advantage to allowing `pint` types?

bluemira/base/parameter_frame/_frame.py
```diff
@@ -164,6 +163,151 @@ def __str__(self) -> str:
...
+
+
+def _convert_angle_units(
+    modified_unit: pint.Unit, orig_unit_str: str, angle_unit: str
+) -> pint.Unit:
+    """
+    Converts angle units to the base unit default for angles.
+
+    Angles are dimensionless therefore dimensionality conversion
+    from pint doesnt work. Conversions between angle units is also not
```
hsaunders1904, line 228:
```suggestion
    from pint doesn't work. Conversions between angle units is also not
```

hsaunders1904 (changes requested):
Sorry about the brain dump! This is nearly there, just a few thoughts.

bluemira/base/constants.py
```diff
@@ -180,6 +218,48 @@ def _transform(
...
+    "[energy]": {"[length]": 2, "[mass]": 1, "[time]": -2},
+    "[pressure]": {"[length]": -1, "[mass]": 1, "[time]": -2},
+    "[magnetic_field]": {"[current]": -1, "[mass]": 1, "[time]": -2},
+    "[electric_potential]": {"[length]": 2, "[mass]": 1, "[time]": -2},
+    "[power]": {"[length]": 2, "[mass]": 1, "[time]": -3},
+    "[force]": {"[length]": 2, "[mass]": 1, "[time]": -3},
+    "[resistance]": {"[current]": -2, "[length]": 2, "[mass]": 1, "[time]": -3},
+}
+
+ANGLE_units = [
```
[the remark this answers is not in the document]
je-cook: yes it should

bluemira/base/parameter_frame/_frame.py
je-cook: `dpa` is stripped because it is dimensionless and `fpy` will be converted to seconds. They both need to be added back in. i will write a docstring to that effect

tests/base/parameterframe/test_parameter.py
```diff
@@ -91,8 +91,7 @@ def test_int_values_converted_to_float_given_float_value_type(self):
 
     def test_repr_contains_name_value_and_unit(self):
         param = Parameter(**self.SERIALIZED_PARAM)
-
-        assert re.search("my_param=100 dimensionless", repr(param))
+        assert re.search("my_param=100 ", repr(param))
```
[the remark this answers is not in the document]
je-cook: pint's automatic formatting short form removes it which is nice :)

bluemira/base/parameter_frame/_parameter.py
je-cook: I havent got a good argument to the contrary so done

hsaunders1904 (approved):
🎉 Looks good!

PR #1505, opened 2022-10-05
Replace uses of old Parameter and ParameterFrame with new versions

### LINKED ISSUE #1422: Replace uses of old `Parameter` and `ParameterFrame` with new versions

hsaunders1904:
## Linked Issues

Closes #1422 

## Description

Completely removes the old `Parameter` and `ParameterFrame`, and replaces with the new.

Most of the work was already done in previous PRs

- #1456 
- #1465 
- #1472 
- #1489 

This PR deletes the `base.parameter` module, removes leftover imports and renames `NewParameter[Frame]` to `Parameter[Frame]`.

## Checklist


bluemira/base/config.py
```diff
@@ -1,427 +0,0 @@
...
-    Variables specified here should be physical in some way, and not represent
-    how the code is being run.
-    Defaults are also specified here, and overridden later.
-    New variables should be defined here, with a corresponding entry in the
-    ConfigurationSchema, and passed onwards as Parameter objects.
-    """
-
-    # fmt: off
-    params = [
-        ['Name', 'Reactor name', 'Cambridge', 'dimensionless', None, 'Input'],
```
je-cook, line 41: Are the descriptions and info here put anywhere at the moment?

bluemira/utilities/tools.py
```diff
@@ -72,10 +72,6 @@ def floatstr(_floatstr, obj, *args, **kwargs):
         https://bugs.python.org/issue42434
         https://bugs.python.org/issue31466
         """
-        from bluemira.base.parameter import Parameter
```
je-cook, line 75: I assume our new fancy Parameters can do this automatically?

je-cook (approved):
Looks good once the tests pass. Couple of questions though

bluemira/base/config.py
hsaunders1904: No, nowhere at the moment. I'm happy for it to just live in the git history until we need it. Unless you think we should put it somewhere?

je-cook: No we dont need to put it somewhere now just making sure I was aware :)

bluemira/utilities/tools.py
hsaunders1904: Do you know what the original intent of this is? If I try to encode a `Parameter` in the JSON, I get an error, even with this in

je-cook: I had to look it up. It was because a Parameter used to identify as the type it was wrapping so the floatstr conversion needed modifying because json writer used `__repr__` to do type checking instead of `isinstance`. Now the whole function could probably be removed. Also while I remember we can remove `wrapt` as a direct dependency now

hsaunders1904: Ah, that makes sense. I've removed it - let's see what breaks 😄

bluemira/utilities/tools.py
```diff
@@ -62,22 +62,6 @@ def default(self, obj):
...
-        https://bugs.python.org/issue36841
-        https://bugs.python.org/issue42434
-        https://bugs.python.org/issue31466
-        """
-        from bluemira.base.parameter import Parameter
-
-        if isinstance(obj, Parameter):
-            obj = obj.value
-        return _floatstr(obj, *args, **kwargs)
-
```
je-cook, line 80: `iterencode` and `_patcher` I think can be deleted too

je-cook (approved):

PR #1536, opened 2022-10-18
Feature/builder reactor redesign

### LINKED ISSUE #98: Proposal: Refactor of default params logic
Currently BLUEPRINT-style parameters are based off of a set of templates for each class, which are defined via a record list using a set of default values and metadata. The metadata, serialisation, and ability to copy through parameters from calling objects is very useful and should remain. However, the definition of the default values and setup of the set of parameters on a class can be improved on, in particular as there is need to support multiple configurations, which may not have sensible corresponding default values, and the assignment of parameters for a given builder should be a fixed definition.

BLUEPRINT uses a top level `Configuration` class, that contains all of the parameters available for a `Reactor`. From discussions with @je-cook, we would propose to introduce a set of `ParameterConfiguration` classes that are used to reflect the state of the parameterisation of a particular builder (or analysis). These give a more concrete definition of the parameterisation of a builder by acting as a schema. We should make it possible to look up the metadata for a given parameter from a set of central definitions, extend that allowed list of values, and validate that the parameters have been defined as expected. Parameters should be initialised with a None value by default to avoid unanticipated conflicts between different design studies.

### LINKED ISSUE #159: Coupling for TF shaping and PF positioning
In GitLab by @DanShort12 on Mar 8, 2021, 14:28

## Description

Some equilibrium may be hard to be reached for a given rigid TF coil positionning. More specifically, it appears to be challenging to obtain an acceptable divertor shaping without having a little flexibility on TF coil shape. On the other had The PF coil positions must take the presence of the TF coil in a way.

## Action

* [ ] Build up the method

### LINKED ISSUE #167: More descriptive names for build_tweaks entries
In GitLab by @DanShort12 on Mar 24, 2021, 14:44

The `build_tweaks` configuration allows for the fine adjustment of tuning parameters in solvers. However, the current naming does not indicate which solver the tweak corresponds to (although the current set are mainly relevant for TF coils). The keys used in `build_tweaks` should be updated to make it clear where the tweaking parameter is used e.g. `tf_nr` rather than `nr`.

### LINKED ISSUE #195: Adjust ParameterFrame so parameters are updated globally.
In GitLab by @je-cook on May 27, 2021, 09:49

Building on the conversation had in #321 and working towards integration of more codes such as process and jetto into BLUEPRINT.

Some design ideas and hopes after discussions with @DanShort12 :

 - The definition of parameters should only be defined once and only accessed from one location therefore any changes if modified would be global.
 - Parameters should be searchable (or gettable) within the Frame, therefore a Frame will not need to be passed around explicitly
 - Parameters should be defined in the class that first uses them as opposed to currently all parameters are defined in `systems.config`
 - The current full list of names should be kept in `base.config_schema` as it is at the moment

This will lead in to a general setup-run-teardown workflow for any new codes, detailed further in #420

If you have any comments or suggestions on this direction let me know

### LINKED ISSUE #216: TF coil re-architecture
In GitLab by @sebkahn on Jul 19, 2021, 16:36

## Description

It is geting more and more difficult to use the TF coil class in a user friendly way without bugs. Certain features are difficult to deal with, such as 
- The impossiblity to define a TF coil properly without any ripple optimization
- The code is really obscure and there, very few commenting and a lot of one letter variable name
- The way the 2D loops are made produce bad quality roundings making the CAD not fit for purpose (or the initial loops needs to be redone in the CAD maker ...). We may want to use CAD objects as well as loops there.
- We may want to expand the notion of segmented coil
- We may want to add the calculation of resisitivity and who knows, optimize on it with the ripple limit set as constraint.

## Proposal

The class has three main separate functionality
1. The class generation (`__init__`) that builds the geometry loops from params inputs and coil shape/WP/casing option selection. The parametrizaton can be done using dedicated class *a la* MIRA, as it seems to be the proper way and as conductor layout has common features for all coils. `__init__` calls `build_tf` that makes the geom loops
2. The optimization that call call the relevant optimizer object with all the needed setted up properly, runs the optimizer, run `build_tf` and `update_params` that update the value of all relevant params (for example these used in `__ini__`
3. The generation of a output param to be used to update the reactor system. The fonction must be returning a param object after calling `update_params` and a set of `@property` functions that calculate the relevant physical quantities such as stored energy, coil inductance, WP and case volumes, resistive heating (cost?). These function should be the same that are used by the optimizer.

Choosing what should be the inheritance based on is not that trivial as there quite a few combinatorics in the design options:
- Shape type 
- Conductor type/geometry
- Support structure geometry (case)
- Number of and type of joints, please note that having joints also allows to have different conductor types.

Conductor and support will be class in MIRA, so I suggest we go in that direction too, and the number of joints can be setup in a relatively generic manner. So I propose to base the inheritance on coil shape, this will also allow to break down the TF coil object better and avoid the sea of `if` statements.

Rem: The geometry loops should be both set in CAD and Loop objects, as loops seems to case **a lot** of troubles fairly easily.

This will be the right moment to **document** a bit better what is going on as well.

## Detailed Code section descrititon

Will be done soon (already written down in pen and paper)

Addition 05/11/2021
Related issues to be resolved by the refactoring: 
- https://github.com/Fusion-Power-Plant-Framework/bluemira/issues/274
- https://github.com/Fusion-Power-Plant-Framework/bluemira/issues/228
- https://github.com/Fusion-Power-Plant-Framework/bluemira/issues/362

@CoronelBuendia, @shaildesai96, @DanShort12 any thoughts on all that ?   
Seb

### LINKED ISSUE #229: Make test for vacuum vessel offset_segment
In GitLab by @helen-brooks on Aug 25, 2021, 14:12

Create tests cases for inboard and outboard sides that raise the GeometryError on line 482. An "E" shape should suffice.

### LINKED ISSUE #262: Avoid storing config as a ReactorSystem member

Currently we store `config` as an attribute on `ReactorSystem`. However, the values are copied into the `params` in the constructor, so we don't need the `config` elsewhere in that class. Removing from the class will also mean that we're not potentially changing any external state.

## Proposed solution

Remove `config` as a class member and adjust `ReactorSystem.__init__` so that the `params` are always created appropriately.

## Alternative solutions

The `__init__` update isn't strictly necessary, but I think it will make for a cleaner architecture.

### LINKED ISSUE #433: Defaults values in BLUEMIRA
## Description

Separating the builder and the Component, we are getting in a situation with more and more shared Params input between with the heavy risk to have 2 different defaults values resulting into a silent error. If setup with config this problem does not appear, but there is a lot of variables and forgetting is easily done... There is also the fact that the user is not expected to know all the inputs and their best value...

## Proposal

The proposition is to have a check that verify if there is a variable set by default by two different objects, and that throw an error asking for a definition and precising where the default are both used.

### LINKED ISSUE #615: Represent required parameters on Builder as a set

We currently use lists to define the required parameters on a Builder. However, when merging builders together e.g. in a Design, we may end up with duplicate values, which would be avoided if we used a set. We should also consider when it is appropriate to assign required parameters to the class and when they should be set on the instance - the required parameters on the class may be useful for external validation, but should never be changed.

## Proposed solution

Convert _required_params to be a set. Ensure _required_params from the class are copied onto the class.

### LINKED ISSUE #619: Cache parameter mappings

It would probably be good to cache send and recv parameter mappings instead of reaccessing all of them every time. At the moment I doubt its a massive performance hit but could be in the future.

## Proposed solution

Cache `ParameterMapping` section of a `ParameterFrame` using `lrucache`. Making `ParameterFrame` directly hashable seems like a hard task (as it could be cached in different ways eg just for values or just for mappings). Further exploration needed

## Alternative solutions

Leave as is and deal with the slight performance hit

## Additional context

The plasmod example calls the solver multiple times and I thought it could be a time saving

### LINKED ISSUE #620: Simplify method to get default mappings for external code

We used to have hard-coded mappings for PROCESS in the `Configuration` class and correspondingly in json parameterisation files. It certainly makes sense for these default mappings to now be owned by the code to which they correspond, but one side effect of this is that the mappings are now further away from where a user might look to know what the defaults would be.

## Proposed solution

It would be useful to have a common method across external code APIs to extract the default mappings and possibly apply them to a `Configuration` to generate the full parameterisation if that code were to be used. Ideally it would be possible for a user to define their own default parameterisations and extract them in the same (or at least similar) way.

## Alternative solutions

When making `Configuration` instances to use for a design, or for writing to json, we could just access the mappings for the external code and build them in (like is currently done in the EU-DEMO example) but it would be cleaner to have some unified way to do this.

### LINKED ISSUE #629: Review of builder functionality

### LINKED ISSUE #811: Remove reinitialise from builders

We currently have a fair bit of complexity being introduced by trying to maintain state though reinitialisation of builder classes. This drives some complications in API, while at the same time not actively being used (or having a clear requirement until we have cases that repeatedly run a given build).

## Proposed solution

Remove reinitialise methods from builders.

## Alternative solutions

We could keep things as they are, but they will likely need to change anyway when we get round to having designs that repeatedly call a given build stage (such as via global optimisers).

## Additional context

Probably worth putting in an additional issue to cover versioning builders when called multiple times.

### LINKED ISSUE #892: Support multiple build stages for a single component

### LINKED ISSUE #1023: Error constructing a ParameterFrame with defaults
## Describe the bug

When constructing a new `ParameterFrame`, if `with_defaults` is set to `True`, a `KeyError` is raised.

## Steps to reproduce

```python
from bluemira.base.parameter import ParameterFrame

pf = ParameterFrame(with_defaults=True)
```

## Expected behaviour

No error. `ParameterFrame` constructed containing default values - which I guess in this case should be no values.

### LINKED ISSUE #1072: `ParameterFrame`, `Configuration` and the setup of individual `Builder` `Solver` etc.

At present all `Parameter`s are "hosted" with default values in a `Configuration` object. Many of these `Parameter`s are used in several locations in every run, regardless of what `Reactor` is being designed. This is partly because of interfaces to codes like `PROCESS` and `PLASMOD` which require a full set of inputs.

Many of these `Parameter`s can be considered as "globals" in the context of a reactor design, even if some of them are recalculated in subsequent steps or may even deviate at different snapshots in time from the reference values.

However, many more `Parameter`s are specific to a certain build stage, `Builder`, or `Solver`. In some cases, these `Parameter`s are not used if an alternative `Builder` or `Solver` is used for a given build stage. This leaves the "global" `Configuration` pre-populated with `Parameter`s that have no relevance for a given design.

## Proposed solution

Reduce the `Parameter`s hosted in `Configuration` to the "global" `Parameter`s and enable `Builder`s and `Solver`s to be initialised both from the global `Configuration` and from a local `ParameterFrame` or `Configuration`.

This may complicate the initialisation somewhat, but will provide a clearer picture of what `Parameter`s are used where.

### LINKED ISSUE #1075: Proposal: `build_config` configurability

AT present, we initialise a reactor with the full build_config, which can come e.g. from some JSON file.

However, in order to facilitate building a reactor more dynamically (e.g. by changing some build stages at different points), it would be nice to also be able to input the build_config for a specific build stage at will.

## Proposed solution

e.g. 

```python

reactor_builder = Reactor(params, global_build_config)
reactor = reactor_builder.run()
reactor_builder.build_TF_coils(reactor, local_TF_build_config)
...

```

## Alternative solutions

status quo

### LINKED ISSUE #1148: Separation of concerns - `Builder` / `OptimisationProblem` / `["run", "read", "mock"]`

We have a couple of situations in the EU-DEMO reactor where we are solving a design optimisation problem in an `OptimisationProblem` and subsequently building some geometry in a dedicated `Builder`. This is the case for at least:
* TF coils
* PF coils (see #1113)

At present, the `Builder` handles the `["run", "read", "mock"]` logic and in the case of `"run"` actually runs the design optimisation problem. This means that a lot of parameters that are technically only required for the `"run"` case are passed into the `Builder`.

It is also relatively opaque in terms of separation of concerns: in many people's heads a `Builder` is responsible for creating geometry, regardless of what inputs it has been given.

The most complicated situation is that of the TF coils, where the aspects of the builder are directly used in the optimisation problem being solved in the case of run. As such this is perhaps the best testing ground for an alternative solution.

## Proposed solution

Make `Builder`s solely responsible for generating CAD.

Split out the concern of `["run", "read", "mock"]` from the `Builder` , potentially instead being managed by the `design_stage` itself(?).

If necessary, the optimisation problem may have to call the `Builder`, but this would be quite "circular" and should be avoided if possible. in the case of the PF coils, for instance, this should not be necessary at all.

## Alternative solutions

There are many ways of slicing this one. To be discussed in the architecture workshop.

### LINKED ISSUE #1197: Solver architecture for parameters and reactor design stages

Following the two ADRs created in #1194 and #1192, we have some options for how to re-structure reactor builds and their configuration.

This issue is to investigate a new architecture for reactor builds, and to implement a `DesignStage` class that handles the `read`, `run`, `mock` logic for each stage of the reactor build.

## Proposed solution

The `DesignStage` class will need `read`, `run`, `mock` methods, which means it will be very similar to a solver. In many cases, `DesignStage`s will share similar setup (parsing and setting-up configuration), so a new class that inherits from solver, using shared setup may make sense.

## Alternative solutions

See ADRs

### LINKED ISSUE #1389: Inconsistent electrical current units
## Description

Whilst working on #1388, I noticed what looked like a unit conversion from MA to A when instantiating a `BetaIpProfile`, relevant code on develop [here](https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/75912fcab2ca0cb3d4fe3751bfe6aaccf87d95dd/examples/design/EU-DEMO/EUDEMO_builders/plasma.py#L286). Looking through our existing parameters in `Configuration`, the [non-zero electric current quantities](https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/develop/bluemira/base/config.py#L72) are in `MA` but [our conventions](https://bluemira.readthedocs.io/en/latest/conventions.html#unit-convention) say electric current should be in `A`. This is why we've got this case where a conversion is necessary; the function takes current in `A` (as it should), but the electrical current parameters are in `MA`.

We should resolve this by using either `A` or `MA` across the board.

je-cook:
## Linked Issues

Closes #98
Closes #159 - covered by the reactor developer designing their designer/builder
Closes #167 build tweaks no longer exists
Closes #195 ParameterFrame rewritten so no longer applies
Closes #216 same fix as #159
Closes #229 dependent to the designer/builder made
Closes #262 config no longer exists
Closes #433 No longer defaults parameter defaults
Closes #615 architecture direction change
Closes #619 mappings are added as soon as code module loaded (TODO double check)
Closes #620 see above (TODO double check)
Closes #629 architecture change so fixed
Closes #811 reinitialise removed
Closes #892 up to the reactor developer
Closes #1023 no longer defaults
Closes #1072 ParameterFrame scope reduced
Closes #1075 Reactor can now be built dynamically
Closes #1148 designer/builder separation
Closes #1197 (TODO check)
Closes #1389 unit defaults are enforced

## Description

This is the builder designer and parameter redesign as discussed variously

## Interface Changes

## Checklist


hsaunders1904 (approved):
✔️ 🎉 

There are quite a lot of sonarcloud warnings that we may want to look at, at some point. But not worth holding this up. I'll open a ticket to look into them.
image

PR #1540, opened 2022-10-19
➖ Remove pandas as a direct dependency

je-cook:
## Description

Pandas was not really used for anything important so removed the final couple of uses. It still seems to be a dependency of a dependency however

## Interface Changes

None

## Checklist


hsaunders1904 (approved):
All looks good to me ✔️

PR #1565, opened 2022-10-24
💚 Fix ref report download

je-cook:
## Description

Ref report hash was empty. This should fix that


je-cook:
the download now has a commit for the ref but it is the wrong commit... need to get the target branch head commit

hsaunders1904 (approved):
Looks good to me

PR #1575, opened 2022-10-27
🐛 solve components named the same at the same level

### LINKED ISSUE #955: Components can have the same name at the same depth in a tree
## Describe the bug

Components with the same name can have the same depth within a tree. The following tree is possible:

```txt
TFCoils (Component)
├── Sector 1 (Component)
└── Sector 1 (Component)
```

Seeing as we access components by name (via `get_component`), it feels as though we shouldn't have more than one component with the same name at the same level.

## Steps to reproduce

There are two ways to recreate the behaviour:

```python 
parent = Component("TFCoils")
Component("Sector 1", parent=parent)
Component("Sector 1", parent=parent)
print(parent.tree())
```

```python
parent = Component("TFCoils")
parent.add_child(Component("Sector 1"))
parent.add_child(Component("Sector 1"))
print(parent.tree())
```

## Expected behaviour

We should raise an error in this case.

```python
    def test_add_child_ComponentError_given_duplicated_component_name(self):
        parent = Component("TFCoils")
        parent.add_child(Component("Sector 1"))

        with pytest.raises(ComponentError):
            parent.add_child(Component("Sector 1"))

    def test_init_ComponentError_given_duplicated_component_name(self):
        parent = Component("TFCoils")
        Component("Sector 1", parent=parent)

        with pytest.raises(ComponentError):
            Component("Sector 1", parent=parent)
````

je-cook:
## Linked Issues

Closes #955

## Description

Possible fix, other options considered were using `__eq__` but I would only check `self.name==other.name` which made lead to some odd situations. 

## Interface Changes

You cant have the same named components at the same level

## Checklist

### DIFF
--- a/bluemira/base/components.py
+++ b/bluemira/base/components.py
@@ -77,7 +77,12 @@ class Component(NodeMixin, Plottable, DisplayableCAD):
     ):
         super().__init__()
         self.name = name
+
+        if parent is not None and name in (ch.name for ch in parent.children):
+            raise ComponentError(f"Component {name} is already a child of {parent}")
+
         self.parent = parent
+
         if children:
             self.children = children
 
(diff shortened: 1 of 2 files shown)


hsaunders1904 (approved):
Good stuff!

Issue #1611, opened 2022-11-04
 `Solver` should inherit from `Designer` functioning as a multistage `Designer`


je-cook:
## Description of issue / requirement to address

A `Solver` is as of writing only used in Codes where multiple stages are required to complete a problem. Essentially this is a multi stage designer. It should be an extension of the `Designer` class and not its own thing

## Proposed solution

Solver should inherit from Designer in some form

## Alternative solutions

Solver lives as a separate object but would need reworking

## Additional context

#1192

PR #1625, opened 2022-11-09
Remove SolverABC

### LINKED ISSUE #1423: Move SolverABC out of base module

Uses of `SolverABC` felt very boilerplate-y when not interfacing an external code. With the introduction of the new `Designer` class, we will no longer be considering using `SolverABC` outside the `codes` module (a solver is basically now a `Designer` with `setup`, `run`, `teardown`).

We currently have `SolverABC` and `CodesSolver` - I think we don't need both of these classes, and we can merge the functionality of the two. We should do similar for the `Task` and `CodesTask` classes.

Remember to remove the `SolverABC` example from `examples/base`. This example could potentially be repurposed as a `Designer example`

hsaunders1904:
## Linked Issues

Closes #1423 

## Description

Removes `SolverABC` and incorporates its functionality into `CodesSolver`. Fortunately, `SolverABC` wasn't used in many places.

I do think we may be missing a piece of the puzzle, in that we have `Builder`s that make CAD, `Designer`s that design CAD, but we don't have a base 'calculator'. It could be that we don't actually need one, but I think it's worth a discussion at some point. #1611 contains one suggestion. 

## Interface Changes

No more `SolverABC`.

## Checklist


je-cook:
> I've just noticed that `RunMode` is used in `equilibria`, but I've moved it into `codes`. This probably needs to be resolved before we can think about merging this.
> 
> Edit: although, the transport solver is using a `CodesSolver` from `codes` anyway. What do you think, @je-cook?

We could put them both under an `if typing.TYPE_CHECKING:` as a quick thing. It may be quite difficult to resolve otherwise

hsaunders1904:
> We could put them both under an if typing.TYPE_CHECKING: as a quick thing. It may be quite difficult to resolve otherwise.

The concern is more of an architectural one - modules should be as 'standalone' as we can make them, and, where possible, only rely on things in `base` and probably `utilities`. In theory, we could definitely have a transport solver that's not a `CodesSolver` in the case of `fem_fixed_boundary`.

Since I noticed that the dependency on `codes` was already there, I'm less worried about this being merged. But I think we should consider this issue as part of #1611 (or open a new ticket)

je-cook (approved):
I'm happy with this as is with the open issues. 

It may be nice to have a CodeSolver example that isnt dependent on having any code installed. Maybe instead of getting that in here we should open an issue for that. I think it could be made from the example thats deleted here.

PR #1659, opened 2022-11-24
Replace matplotlib private `QuadContourGenerator` with contourpy

### LINKED ISSUE #1452: Consider using contourpy for contours instead of matplotlib

We should consider using [`contourpy`](https://github.com/contourpy/contourpy) for generating contours in the equlibria module.

We're currently [using Matplotlib's private `QuadContourGenerator`](https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/develop/bluemira/equilibria/find.py#L30) to generate contours. This leaves us at the mercy of Matplotlib changing their private API, and has already [caused problems once before](https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/998) when upgrading Matplotlib.

Matplotlib has just released 3.6.0, which includes updates to [contour generation](https://matplotlib.org/stable/users/prev_whats_new/whats_new_3.6.0.html#improved-quad-contour-calculations-via-contourpy) - they're shifting to using contourpy instead of their own implementation. This is likely to cause some disruption when we next update Matplotlib.

Instead of continuing to use Matplotlib to generate contours, we should directly use contourpy, which should have a much more stable API.

This is not urgent, but should be considered in our next few dependency update cycles.

## Proposed solution

Use contourpy for contour generation.

## Alternative solutions

Keep using private Matplotlib contour generation (which will be using contourpy under-the-hood anyway), or use an alternative contour generation package.

CoronelBuendia:
## Linked Issues

Closes #1452 

## Description

Replaces use of matplotlib private functionality for contour generation on quad grids (now removed in more recent matplotlib versions) with contourpy.

Note that `TriContourGenerator` in matplotlib is still presently used.

## Interface Changes

N/A

## Checklist

### DIFF
--- a/bluemira/equilibria/find.py
+++ b/bluemira/equilibria/find.py
@@ -27,7 +27,7 @@ import operator
 
 import numba as nb
 import numpy as np
-from matplotlib._contour import QuadContourGenerator
+from contourpy import LineType, contour_generator
 from scipy.interpolate import RectBivariateSpline
 
 from bluemira.base.look_and_feel import bluemira_warn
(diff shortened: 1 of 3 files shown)


je-cook (changes requested):
Looks good to me but could you add `contourpy` to pyproject.toml @ about L20 as its a direct dependency

je-cook (approved):

PR #1686, opened 2022-12-01
Fixed boundary equilibrium plotting

### LINKED ISSUE #1683: Fixed boundary equilibrium plotting is "all or nothing"

When running a fixed boundary equilibrium solve, there is a fair amount of text displayed, and no plots are shown in `plot=False` is passed in.

If `plot=True` is passed in, there are a huge number of plots that come up at each outer and inner iteration. These plots must be closed for the solution to continue. The plots are not saved.

This comes from a time during development when all these plots were useful for debugging, but as for the purposes of "plotting to display" I'd argue it's overkill, and quite fiddly (closing all the windows every time they pop up).

We have many `plot` flags in various parts of the code, and they tend to provide a relatively concise (opinionated) view of the problem. In the free boundary equilibrium solve, a single plot is updated at each iteration, and does not stop the rest of the code from running.

## Proposed solution

Plotting features we want in the transport/fixed boundary equilibrium solve:
* concise plotting for display purposes (minimal number of figures) (using `plot=True` as in the rest of bluemira)
* detailed plots for debugging purposes which are saved
* the ability to create GIFs 
* the same colour schemes as for the free boundary equilibrium (e.g. Jtor, psi, field, etc. all have a specific colour map in equilibrium/plotting.py)
```python
PLOT_DEFAULTS = {
    "psi": {
        "nlevels": 15,
        "cmap": "viridis",
    },
    "field": {
        "nlevels": 15,
        "cmap": "magma",
    },
    "current": {
        "nlevels": 30,
        "cmap": "plasma",
    },
...
```

## Additional context

@ivanmaione

CoronelBuendia:
## Linked Issues

Closes #1683 

## Description

Fixed boundary equilibrium iteration

Transport iteration

## Checklist

### DIFF
--- a/bluemira/utilities/plot_tools.py
+++ b/bluemira/utilities/plot_tools.py
@@ -96,7 +96,9 @@ def str_to_latex(string):
 
 def make_gif(folder, figname, formatt="png", clean=True):
     """
-    Makes a GIF image from a set of images with similar names in a folder
+    Make a GIF image from a set of images with similar names in a folder.
+    Figures are sorted in increasing order based on a trailing number, e.g.
+    'figure_A'[1, 2, 3, ..].
     Cleans up the temporary figure files (deletes!)
     Creates a GIF file in the folder directory
 
@@ -105,7 +107,7 @@ def make_gif(folder, figname, formatt="png", clean=True):
     folder: str
         Full path folder name
     figname: str
-        Figure name prefix. E.g. 'figure_A'[1, 2, 3, ..]
+        Figure name prefix
     formatt: str (default = 'png')
         Figure filename extension
     clean: bool (default = True)
--- a/bluemira/utilities/plot_tools.py
+++ b/bluemira/utilities/plot_tools.py
@@ -116,10 +116,8 @@ def make_gif(folder, figname, formatt="png", clean=True):
     ims = []
     for filename in os.listdir(folder):
         if filename.startswith(figname) and filename.endswith(formatt):
-            fp = os.path.join(folder, filename)
-            ims.append(fp)
 
-    find_digit = re.compile("(\d+)")
+    find_digit = re.compile("(\\d+)")
     ims = sorted(ims, key=lambda x: int(find_digit.findall(x)[-1]))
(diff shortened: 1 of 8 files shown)


bluemira/utilities/plot_tools.py
```diff
@@ -116,7 +119,7 @@ def make_gif(folder, figname, formatt="png", clean=True):
             if filename.endswith(formatt):
                 fp = os.path.join(folder, filename)
                 ims.append(fp)
-    ims = sorted(ims)
+    ims = sorted(ims, key=lambda x: int(re.sub("\\D", "", x)))
```
CoronelBuendia, line 122: @je-cook please take a look; I suck at regex. The idea is to get the trailing number in a list of filenames and sort based on increasing order. This is effectively to avoid adding natsort as a dependency.

je-cook: It looks fine but maybe find existing instead of not existing might be better? This makes sure we get the last set of digits eg 'f1231ig_123.png' -> 123:
```suggestion
    find_digit = re.compile("(\d+)")
    ims = sorted(ims, key=lambda x: int(find_digit.findall(x)[-1]))
```

je-cook: the only hole I can see in both is if there are no numbers or numbers in weird places but probably catching those is overkill

CoronelBuendia: Thanks!

CoronelBuendia:
I don't know what happened, but there is a problem now with the reconstructed profile of jpar when using fully relaxed q profile:
image
I think it is because before I was getting the raw pprime and ffprime, and now they are scaled, but honestly not sure how long this has been the case, nor why it is happening now.

ivanmaione (approved):
I am fine with this PR. The new plotting output is concise and reports all the information necessary to overview the fixed boundary solution.

je-cook (commented):
Just need to rebase because the notebooks have been removed

je-cook (approved):

PR #1706, opened 2022-12-07
🥅 catch freecad error

### LINKED ISSUE #1309: FreeCAD error raised if geometry.tools.convert fails
## Description of issue

The problem here is that when `geometry.tools.convert` fails to convert a FreeCAD shape to a Bluemira object (e.g., after we've tried to revolve a circle 360 degrees), it raises a FreeCAD error. This means, in order to catch the error, we need to import FreeCAD. We should wrap the convert function in try-catch and raise a `GeometryError` instead.

je-cook:
## Linked Issues

Closes #1309

## Description

I created it as a decorator so that the api would be constant but I'm open to other options. 

Possibly could generalise it further and have the error to catch and the error to raise as input and put the function in utilities

## Interface Changes

`tools.convert` will now raise a geometry error instead of a FreeCADError on failure

## Checklist

### DIFF
--- a/bluemira/codes/_freecadapi.py
+++ b/bluemira/codes/_freecadapi.py
@@ -79,7 +79,7 @@ MAX_PRECISION = 1e-5
 # ======================================================================================
 
 
-def catch_caderr(raise_error):
+def catch_caderr(new_error_type):
     """
     Catch CAD errors with given error
     """
@@ -89,7 +89,7 @@ def catch_caderr(raise_error):
             try:
                 return func(*args, **kwargs)
             except FreeCADError as fe:
-                raise raise_error(fe.args[0]) from None
+                raise new_error_type(fe.args[0]) from fe
 
         return wrapper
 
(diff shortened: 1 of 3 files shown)


bluemira/codes/_freecadapi.py
```diff
@@ -73,6 +73,29 @@
 WORKING_PRECISION = 1e-5
 MIN_PRECISION = 1e-5
 MAX_PRECISION = 1e-5
+
+# ======================================================================================
+# Error catching
+# ======================================================================================
+
+
+def catch_caderr(raise_error):
```
hsaunders1904, line 82: This argument should be an error type, from how it's named I thought it was meant to be a function. Maybe `new_error_type` would be a clearer name?

bluemira/codes/_freecadapi.py
```diff
@@ -73,6 +73,29 @@
...
+    """
+    Catch CAD errors with given error
+    """
+
+    def argswrap(func):
+        def wrapper(*args, **kwargs):
+            try:
+                return func(*args, **kwargs)
+            except FreeCADError as fe:
+                raise raise_error(fe.args[0]) from None
```
hsaunders1904, line 92: If we raise from `None`, do we lose the information the `FreeCADError` is telling us? If so, I think we should still raise from the `FreeCADError`. The issue was basically just saying the error needs to be catchable via `except BluemiraError`

hsaunders1904 (changes requested):

hsaunders1904 (approved):
👍

PR #1707, opened 2022-12-07
🐛 fix indexing bug in inscribed_rect test

je-cook:
## Description

When trying to catch errors the indexing wasnt nested deep enough.
This is very hard to see if it fixes the occasional failure we see because I haven't had a local failing example

## Checklist

### DIFF
--- a/bluemira/geometry/coordinates.py
+++ b/bluemira/geometry/coordinates.py
@@ -1148,7 +1148,7 @@ def _parse_to_xyz_array(xyz_array):
 
 def _parse_array(xyz_array):
     try:
-        xyz_array = np.array(np.atleast_2d(xyz_array), dtype=np.float64)
+        xyz_array = np.array(np.atleast_2d(np.squeeze(xyz_array)), dtype=np.float64)
     except ValueError:
         raise CoordinatesError(
             "Cannot instantiate Coordinates with a ragged (3, N | M) array."
--- a/tests/geometry/test_inscribed_rect.py
+++ b/tests/geometry/test_inscribed_rect.py
@@ -102,7 +102,9 @@ class TestInscribedRectangle:
 
                         if tf is not None:
                             # Some overlaps are points or lines of 0 area
-                            if not all([get_area(*t.xyz) == 0.0 for t in tf]):
+                            if not all(
+                                [get_area(*seg.xyz) == 0.0 for t in tf for seg in t]
+                            ):
                                 self.assertion_error_creator(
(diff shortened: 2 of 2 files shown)


CoronelBuendia (changes requested):
5% failure rate before this PR:
```python
>   if not all([get_area(*t.xyz) == 0.0 for t in tf]):
E   AttributeError: 'list' object has no attribute 'xyz'
```

4% failure rate with this PR:
```python
.0 = <list_iterator object at 0x7f0581b83b80>

        [
>           get_area(*seg.xyz) <= self.MIN_AREA
            for t in tf
            for seg in t
        ]
    ):
E   AttributeError: 'numpy.ndarray' object has no attribute 'xyz'

```

(Rates on 100 runs. There are 5 tests in file, so 500 tests, so technically a lower failure rate in terms of tests)

je-cook:
Getting there! Ok I'll have another go

CoronelBuendia:
5 failures in 100 runs:

```python
tests/geometry/test_inscribed_rect.py:100: in <listcomp>
    Coordinates(seg.discretize(byedges=True, ndiscr=50))
bluemira/geometry/coordinates.py:1244: in __init__
    self._array = _parse_to_xyz_array(xyz_array)
bluemira/geometry/coordinates.py:1143: in _parse_to_xyz_array
    xyz_array = _parse_array(xyz_array)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

xyz_array = array([[[ 3.59999968,  0.        , -4.99999844],
        [ 3.59999968,  0.        , -4.99999856],
        [ 3.59999968...    , -4.99999868],
        [ 3.5999997 ,  0.        , -4.99999856],
        [ 3.59999968,  0.        , -4.99999844]]])

    def _parse_array(xyz_array):
        try:
            xyz_array = np.array(np.atleast_2d(xyz_array), dtype=np.float64)
        except ValueError:
            raise CoordinatesError(
                "Cannot instantiate Coordinates with a ragged (3, N | M) array."
            )
    
        shape = xyz_array.shape
        if len(shape) > 2:
>           raise NotImplementedError
E           NotImplementedError

bluemira/geometry/coordinates.py:1159: NotImplementedError
```

I think you need to treat the point and 0 length segments properly somehow..

je-cook:
I ran it 500 times and got no failures, but then again I didnt get many before so its probably worth another check

PR #1733, opened 2022-12-12
Add fixed - free boundary equilibrium solve to EU-DEMO

### LINKED ISSUE #1476: Fixed boundary to free boundary equilibrium for EU-DEMO

For the EU-DEMO builder sequence, we need to move from a fixed boundary equilibrium solution (coupled from a PLASMOD solve) into a free boundary equilibrium solution.

From the PLASMOD <-> fixed boundary solve we can take:
* p' and FF'
* Plasma current
* a set of points along the boundary
* X-point location

and use these in a unconstrained solve for the reference plasma equilibrium with `n_PF` arbitrarily spaced PF coils and `n_CS` CS coils. This would replace the step prior to the first pass design of the first wall.

## Proposed solution
Implement an EU-DEMO design stage performing the above operation parametrically.

## Alternative solutions

This is part of TSVV-14 Task 6

## Additional context

Best to wait until after the architecture refactor before implementing.

CoronelBuendia:
## Linked Issues

Requires #1892 
Closes #1476 

## Checklist

### DIFF
--- a/eudemo/eudemo/equilibria/_equilibrium.py
+++ b/eudemo/eudemo/equilibria/_equilibrium.py
@@ -127,7 +127,7 @@ def make_reference_equilibrium(
         r_cs=params.r_cs_in.value + params.tk_cs.value / 2,
         tk_cs=params.tk_cs.value / 2,
         n_CS=5,
-        n_PF=12,
+        n_PF=6,
     )
 
     grid = make_grid(
--- a/eudemo/eudemo/equilibria/_equilibrium.py
+++ b/eudemo/eudemo/equilibria/_equilibrium.py
@@ -130,8 +130,8 @@ def make_reference_equilibrium(
         tf_track,
         r_cs=params.r_cs_in.value + 0.5 * params.tk_cs.value,
         tk_cs=0.5 * params.tk_cs.value,
-        n_CS=10,
-        n_PF=20,
+        n_CS=5,
+        n_PF=6,
     )
 
     grid = make_grid(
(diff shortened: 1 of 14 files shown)


eudemo/eudemo/equilibria/_equilibrium.py
```diff
@@ -94,3 +96,79 @@ def make_equilibrium(
...
+        nx=nx,
+        nz=nz,
+    )
+
+    profiles = CustomProfile(
+        p_prime,
+        ff_prime,
+        R_0=params.R_0.value,
+        B_0=params.B_0.value,
+        I_p=params.I_p.value * 1e6,  # This is in fact required, James.
```
je-cook, line 166: Is it always the case that you want I_p in `μA`, or 1e6 times larger than the input?

If so this factor should probably go into CustomProfile as below if all user facing functions are in SI units. Otherwise can we use the below:
 ```suggestion
        I_p=raw_uc(params.I_p.value, "A", "μA")
```

je-cook (changes requested):
Just the one question otherwise looks good to me, maybe some of the TODO's should be pulled out into an issue or two but they mostly look minor

eudemo/eudemo/equilibria/_equilibrium.py
je-cook: It should be in [A] (`params.I_p.unit' should only ever be amps because we autoconvert) , if thats the case the PLASMOD API needs to change, not adding a factor here.

je-cook: If everything is working correctly we already convert it: https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/develop/bluemira/codes/plasmod/mapping.py#L579

CoronelBuendia: Aight sorry it was coming from I_p in mock_PROCESS being set in MA instead of A.

je-cook (approved):

PR #1748, opened 2022-12-16
Run examples in CI

### LINKED ISSUE #1329: Run examples in CI

It would be good to get an early warning of any examples that are broken. We should decide how regularly we want to run the examples. Some take a long time to run, so we may only want to run these occasionally (say once a week), but some are very fast and should be run more regularly.

## Outputs

- [ ] A script to run the examples (we need to be careful of interactive windows blocking execution)
- [ ] Additions to our GitHub workflows to run the examples on a regular basis

hsaunders1904:
## Linked Issues

Closes #1329 

## Description

- Adds a script to run examples in sequence.
- Adds GitHub Actions workflow to run the examples, and fail if there are any errors
  - The workflow will run every Monday at 04:37
  - It can also be triggered manually via 'workflow dispatch'
- Examples using plasmod/process have been disabled

## Interface Changes

Nope

## Checklist


scripts/run_examples.py
```diff
@@ -0,0 +1,130 @@
...
+        exec(compile(source, file_path, "exec"), globals())  # noqa: S102
+    except Exception as e:
+        print(e, file=sys.stderr)
+        return False
+    finally:
+        plt.close("all")
+    return True
+
+
+def run_examples(example_files: List[str], plotting_on=False) -> List[Tuple[str, bool]]:
```
je-cook, line 92:
```suggestion
def run_examples(example_files: List[str], plotting_on: bool = False) -> List[Tuple[str, bool]]:
```

je-cook (approved):
Looks great! One very pedantic comment

PR #1765, opened 2022-12-19
General docs review and code cleanup

### LINKED ISSUE #1418: Documentation Review
We should review and update our documentation for version 1

TODO:

- [ ] Fill list with documentation holes

### LINKED ISSUE #1541: Resolve some sonarcloud code smells introduced in feature branch
## Description

The PR for merging the feature branch into develop (#1536) had quite a few code smell warnings from sonarcloud. Some of these are not issues (e.g., 'name does not match regex' for tests, and or, parmeter names), but there are several that are worth looking into.

Here's the sonarcloud report (please don't expire 🙏 - I can't work out if you can export the report somehow)
https://sonarcloud.io/project/issues?resolved=false&types=CODE_SMELL&sinceLeakPeriod=true&pullRequest=1536&id=Fusion-Power-Plant-Framework_bluemira

je-cook:
## Linked Issues

Closes #1418 
Closes #1541 

## Description

A code cleanup and a documentation review

## Interface Changes

None

## Checklist


bluemira/equilibria/find.py
```diff
@@ -893,13 +890,7 @@ def grid_2d_contour(x, z):
...
             # Add an intermediate point (ccw)
-            if x[i + 1] > xi and z[i + 1] < zi:
-                x_new.append(x[i + 1])
-                z_new.append(zi)
-            elif x[i + 1] > xi and z[i + 1] > zi:
-                x_new.append(xi)
-                z_new.append(z[i + 1])
-            elif x[i + 1] < xi and z[i + 1] > zi:
+            if (x[i + 1] > xi and z[i + 1] < zi) or (x[i + 1] < xi and z[i + 1] > zi):
                 x_new.append(x[i + 1])
```
CoronelBuendia, line 894: This doesn't look right. append z[i+1] to z_new in one case

bluemira/geometry/tools.py
```diff
@@ -1257,14 +1256,10 @@ def deserialize_shape(buffer: dict):
     supported_types = [BluemiraWire, BluemiraFace, BluemiraShell]
 
     def _extract_mesh_options(shape_dict: dict):
-        mesh_options = None
+        mesh_options = meshing.MeshOptions()
```
CoronelBuendia, line 1259: This won't return None if neither lcar or physical group is not in the shape_dict. Perhaps we don't want this anyway, but as is the refactor is not 1:1.

CoronelBuendia (changes requested):
A couple of non-1:1 things that I picked up on. I didn't really look at the docs

bluemira/equilibria/find.py
je-cook: Its a direct copy of what was there as far as I can tell? 
The first half of the if is L896 and the second half is l902. 
L899 is absorbed into the else

bluemira/geometry/tools.py
je-cook: sure I'll rework so its the same for now

bluemira/equilibria/find.py
CoronelBuendia: ah right true; well I guess it is unit-tested anyway! Not sure I find this easier to read TBH but it's all good then

CoronelBuendia (approved):

PR #1767, opened 2022-12-19
Swap out scipy optimiser for NLOpt in FEM fixed boundary equilibria magnetic axis finding

CoronelBuendia:
## Linked Issues

Partially addresses #1253

## Checklist

### DIFF
--- a/bluemira/equilibria/fem_fixed_boundary/equilibrium.py
+++ b/bluemira/equilibria/fem_fixed_boundary/equilibrium.py
@@ -365,11 +365,11 @@ def solve_transport_fixed_boundary(
     else:
         # If we don't break we didn't converge
         message = bluemira_warn
-        line_1 = f"did not converge within {max_iter}"
+        line_1 = f"did not converge within {max_iter} iterations"
         ltgt = ">"
 
     message(
-        f"PLASMOD <-> Fixed boundary G-S {line_1} iterations:\n\t"
+        f"PLASMOD <-> Fixed boundary G-S {line_1}:\n\t"
         f"Target kappa_95: {kappa95_t:.3f}\n\t"
         f"Actual kappa_95: {kappa_95:.3f}\n\t"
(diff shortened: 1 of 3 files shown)


CoronelBuendia:
@ivanmaione, I actually just tried this with the FE derivatives of psi, with something along the lines of:

```python
dpsi_dx = dolfin.project(self.psi.dx(0))
dpsi_dz = dolfin.project(self.psi.dx(1))

...
grad[0] = -dpsi_dx(*x)
grad[1] = -dpsi_dz(*x)
```
Obviously there's a bit of an overhead in doing this project (about 0.5 seconds at the start and 8-9 ms thereafter), and to my surprise it did not perform any better (actually slightly worse!). I guess for psi it is such a smooth function that numerical gradients are not so bad. Or possibly that the FE derivatives are not so good, maybe I need a different order or something. Either way, so far I don't see gains with the "analytical" gradient.

ivanmaione:
> 

Since self.psi is already a function, unless it is not necessary to project it on a different space, the projection should not be necessary (i.e. you already have access to `grad` being calling self.psi.dx). In this way, one would gain that 0.5s.

CoronelBuendia:
> > 
> 
> Since self.psi is already a function, unless it is not necessary to project it on a different space, the projection should not be necessary (i.e. you already have access to `grad` being calling self.psi.dx). In this way, one would gain that 0.5s.

Right, but I couldn't find a way to evaluate `self.psi.dx` at a point. Is there a way?

ivanmaione:
> > Since self.psi is already a function, unless it is not necessary to project it on a different space, the projection should not be necessary (i.e. you already have access to `grad` being calling self.psi.dx). In this way, one would gain that 0.5s.
> 
> Right, but I couldn't find a way to evaluate `self.psi.dx` at a point. Is there a way?

You are right. Unfortunately, I don't think there is a faster way to do that.

ivanmaione (approved):

PR #1769, opened 2022-12-20
Remove scipy optimisation from FEM fixed boundary solve

### LINKED ISSUE #1253: Replace scipy's `minimize` in equilibria/fixed_boundary/utilities::`calculate_plasma_shape_params` with a more robust solution

CoronelBuendia:
## Linked Issues

Closes #1253

## Checklist

### DIFF
--- a/bluemira/equilibria/fem_fixed_boundary/equilibrium.py
+++ b/bluemira/equilibria/fem_fixed_boundary/equilibrium.py
@@ -335,12 +335,24 @@ def solve_transport_fixed_boundary(
 
         gs_solver.solve(plot=plot)
 
+        from bluemira.equilibria.fem_fixed_boundary.utilities import (
+            calculate_plasma_shape_params_new,
+        )
+
         _, kappa_95, delta_95 = calculate_plasma_shape_params(
             gs_solver.psi_norm_2d,
             mesh,
             np.sqrt(0.95),
         )
-
+        print("\n")
+        print(f"Old: {kappa_95=}, {delta_95=}")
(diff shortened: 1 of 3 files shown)


CoronelBuendia:
Still without making use of FE gradients, same discussion as for #1767

CoronelBuendia:
Expect changes in the `equilibrium_example.py` results, as the calculation of delta and kappa 95 depend on this optimisation. I find that the NLOpt SLSQP is not so happy with the problem as the scipy one, because the starting solutions are infeasible, and I imagine scipy has a workaround for this. NLOpt's COBYLA I found to perform better than scipy's SLSQP on this problem, so the kappa and delta 95 values should be _slightly_ more "correct" in the vast majority of cases, which thereby changes results.

ivanmaione (approved):

PR #1794, opened 2023-01-03
✨ Mypy config and typing in documentation modification

je-cook:
## Description

This adds a basic mypy config with the eventual long term aim of enabling mypy in the CI system in some form.

It also modifies the API documentation so that types no longer need to be specified in the docstring. 

Types should now be specified in the function signature, they will be automatically added to both the signature and docstring when the documentation is built. Up for discussion however, we may just want the types in the description or signature.

## Interface Changes

None

## Checklist

### DIFF
--- a/documentation/source/conf.py
+++ b/documentation/source/conf.py
@@ -98,6 +98,8 @@ numfig = True
 
 
 # --- Configuration for sphinx-autoapi ---
+autodoc_typehints = "both"
+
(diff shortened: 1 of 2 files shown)


hsaunders1904 (approved):
Look like reasonable settings to me. Out of interest, are these some base settings from somewhere, or have you come up with for our use-case?

je-cook:
Its strict without `disallow_untyped_defs = true` because of `-> None` things

PR #1829, opened 2023-01-12
Add EQDSK saving for fixed boundary equilibrium

### LINKED ISSUE #1462: Fixed boundary equiliibrum: save to file

At present, it is not possible to save a fixed boundary equilibrium solution to a file. This is a common piece of functionality for all equilibrium solvers, and is desirable output.

## Proposed solution

We could use the `EQDSKInterface` we presently have to create the dreaded but still popular EQDSK format, and our own JSON formats. Obviously there will be no coil information, but this is in any case not in the original G-EQDSK specification.

## Alternative solutions
Not store the output of the fixed boundary equilibrium solver in a file, or just use the mesh file(s).

## Additional context
The `EQDSKInterface` is being refactored to be independent of `ParameterFrame`. Probably best to wait until that is done to minimise churn. This is not urgent.

CoronelBuendia:
## Linked Issues

Closes #1462 

## Checklist


bluemira/equilibria/fem_fixed_boundary/fem_magnetostatic_2D.py
```diff
@@ -350,8 +351,13 @@ def define_g(
             Target current (also used to initialize the solution in case self.psi is
             still 0 and pprime and ffprime are, then, not defined).
             If None, plasma current is calculated and not constrained
+        B_0: float
+            Toroidal field at R_0 [T]. Used when saving to file.
         """
         self._curr_target = curr_target
+        self._B_0 = B_0
+        self._pprime = pprime
+        self._ffprime = ffprime
```
CoronelBuendia, line 360: This is only really required so that an EQDSK can be made from a G-S solver. It would actually be much nicer to pass in the raw data of the pprime and ffprime (if they are arrays), and interpolate them internally to the G-S solver. This is because when it comes to saving an equilibrium, there is no easy way of knowing how many points to specify in the psi_norm vector to retrieve the profiles.

The present solution therefore has inaccurate profiles stored in the equilibrium file.

CoronelBuendia:
So, I needed to do this to split up the PLASMOD <-> fixed boundary from the free boundary in to separate stages in the EU-DEMO build. This will also speed up development as I don't need to run the time-consuming solve before doing the fixed<->free boundary equilibrium step.

However, I didn't want to change the architecture too much, e.g. by adding some some of `FixedBoundaryEquilibrium` class that stores all the necessary data, so I wrote a function that takes a solver and does the necessary handling from there. Without changing the handling of the profiles in the G-S solver, the output file cannot properly interpolate the profiles. I can already see discrepancies in the flux functions. There is a workaround by doing the profile interpolation inside the G-S solver, without affecting architecture too much, but best to discuss first.

PR #1851, opened 2023-01-16
🐛 Fix bug in equilibria run from bad conversion to new coilset

je-cook:
## Description

Missed probably because we don't have many tests of equilibria/run, will add some before I take this out of draft

## Interface Changes

None

## Checklist

### DIFF
--- a/bluemira/equilibria/coils/_grouping.py
+++ b/bluemira/equilibria/coils/_grouping.py
@@ -171,12 +171,42 @@ class CoilGroup(CoilGroupFieldsMixin):
         """
         Fix the sizes of coils in CoilGroup
         """
-        for fx in self.__list_getter("fix_size"):
-            fx()
+        self.__run_func("fix_size")
 
-    def _resize(self, currents):
-        for rs, current in zip(self.__list_getter("_resize"), currents):
-            rs(current)
+    def resize(self, currents: Union[float, List, np.ndarray]):
+        """
+        Resize coils based on their current if their size is not fixed
+        """
+        self.__run_func("resize", currents)
+
+    def _resize(self, currents: Union[float, List, np.ndarray]):
+        """
+        Resize coils based on their current
+
+        Notes
+        -----
+        Ignores any protections on their size
+        """
+        self.__run_func("_resize", currents)
+
+    def __run_func(self, func: str, *args, **kwargs):
+        """Runs a function with no outputs that exists on a coil or coilgroup"""
+        if not args:
+            for ff in self.__list_getter(func):
+                ff(**kwargs)
+        else:
+            args = list(args)
+            funclist = self.__list_getter(func)
+            len_funclist = len(funclist)
+            for no, arg in enumerate(args):
+                if isinstance(arg, (float, int)):
+                    args[no] = np.full(len_funclist, arg)
+                elif len(arg) != len_funclist:
+                    raise ValueError(
+                        f"length of {arg} != number of coilgroups ({len_funclist})"
+                    )
+            for ff, *_args in zip(funclist, *args):
+                ff(*_args, **kwargs)
 
     def add_coil(self, *coils: Union[Coil, CoilGroup[Coil]]):
         """Add coils to the coil group"""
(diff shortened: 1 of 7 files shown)


je-cook:
Ok I'm stopping here . In order to test this properly I'm going to reorganise this slightly in #1852. Now at least the pf coil section of eudemo creates a coilset again. 

Next step will be usng the builders in `bluemira/builders/pfcoils`. At the moment the CS coils have no size in the coilset however.

eudemo/eudemo/pf_coils/tools.py
```diff
@@ -189,7 +189,7 @@ def make_coilset(
     for s in solenoid:
         s.fix_size()
 
-    tf_track = offset_wire(tf_boundary, 1, join="arc")
+    tf_track = offset_wire(tf_boundary, 1)
```
je-cook, line 192: The track only produces a very short segment otherwise

bluemira/equilibria/coils/_grouping.py
```diff
@@ -171,8 +171,42 @@ def fix_sizes(self):
         """
         Fix the sizes of coils in CoilGroup
         """
-        for fx in self.__list_getter("fix_size"):
-            fx()
+        self.__run_func("fix_size")
```
hsaunders1904, line 174: What's the advantage `__run_func` over just calling the function? It's not clear to me what `__run_func` is doing. I think it could do with a bit more explanation on why it's needed/what it offers that directly calling the function(s) does not.

CoronelBuendia (approved):

bluemira/equilibria/coils/_grouping.py
```diff
@@ -460,7 +494,7 @@ def get_max_current(self, max_current: float = np.infty) -> np.ndarray:
         np.ndarray
         """
         return np.where(
-            np.isnan(self.j_max),
+            np.isnan(self.j_max) ^ ~self._flag_sizefix,  # or not
```
hsaunders1904, line 497: Wouldn't 'or not' be ` x | ~y`, not `x ^ ~y`?

Also, isn't `x ^ ~y` equivalent to `x == y`? Which I think could make the intent clearer

hsaunders1904 (commented):
A couple of small things, but otherwise looks fine to me

bluemira/equilibria/coils/_grouping.py
je-cook: In this case its just to avoid repeat code, but in more complex situations you may be passing weirdly shaped object in a nested way into a nested coil groups function eg a circuit within a group being the simplest (complex) situation

bluemira/equilibria/coils/_grouping.py
je-cook: Your right it should be `x | ~y` 

on your second point no not quite because for a XOR -> NOT (XNOR) gate `False + False = True`

hsaunders1904: Not sure what you mean on the second point, both `x ^ ~y` and `x == y` return `True` for `x=False, y=False`.

```pycon
In [1]: import numpy as np

In [2]: x = np.array([True, True, False, False])

In [3]: y = np.array([True, False, True, False])

In [4]: x ^ ~y
Out[4]: array([ True, False, False,  True])

In [5]: x == y
Out[5]: array([ True, False, False,  True])
```

Or am I misunderstanding something?

je-cook: Just because I kept second guessing myself

j_max is nan -> use max current
not size fixed -> use max current

|RAW|||WITH NOT CONVERSION| ||| 
|--|--|--|--|--|--|--| 
|F   |T | F|    -> |  F | F | F|
|T  |F|   T   |-> |  T  |T |  T|
|T | T  | T  | ->  | T | F  | T|
|F|  F  | T |  ->   |F | T |  T|

je-cook: Ahh not proper AND apparently. In that case yes they are the same

https://en.wikipedia.org/wiki/And_gate

PR #1826, opened 2023-01-10
✨ Add notebook to docs

je-cook:
## Linked Issues

Closes #1825 

~~Currently base off #1784~~ 

## Description

Adds a new examples section to the docs. I've done a bit of title cleanup. A few additional dev dependencies also.

Some weirdness to include the examples in sphinx I've added a symlink (I couldn't find a way to get sphinx to include the files otherwise). This means that the link to our data folder in the materials example is broken, hence the suppression in `conf.py`. If we can find a way that we can directly include the examples the link would work.

Secondly I've had to rename the examples (the extension is .ex.py) so that we don't pickup all python files in the documentation with ~nbsphinx~ mynb-st.

~Nbsphinx needs an install of pandoc hence the addition to the docker container and conda env, the pip version doesnt include the binary~ Now using mynb-st which doesnt need pandoc

the remainder of the added deps is for nicety in the docs eg copying buttons on all code blocks

## Checklist

I confirm that I have completed the following checks:

- [ ] Tests run locally and pass `pytest tests --reactor`
- [ ] Code quality checks run locally and pass `flake8` and `black .`
- [ ] Documentation built locally and checked `sphinx-build -W documentation/source documentation/build`

PR #1853, opened 2023-01-17
✨ External Codes example

### LINKED ISSUE #1629: Write a CodesSolver example

We should write an example for setting up a `CodesSolver`, that doesn't use an external  dependency that's not installed by Bluemira.

## Proposed solution

We could potentially re-purpose the [old `SolverABC` example](268271d61516ff19198d1f2258e94487271d8969) that was removed in #1625.

je-cook:
## Linked Issues

Closes #1629 

currently based off #1826 

## Description

Adds an external code example with a python script as the external code, still needs a bit of finessing I think, especially the last bit of explanation

## Checklist


examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,387 @@
...
+
+# %% [markdown]
+# ### Using the solver
+#
+# To run the solver you just need to provide the parameters and the configuration
+# to initialse the object.
+# Be aware `problem_settings` should be used sparingly for options that won't
+# change within the rerunning of the solver,
+# it has the same effect as modifying the default.
+# Also note the units of the parameter values returned back have been updated
```
hsaunders1904, line 327: We get a few warnings when we run this example. Is it worth mentioning them here, or in a comment within the code block? Just to let the users know this is expected and why.

```suggestion
# Also note the units of the parameter values returned back have been updated.
```

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,387 @@
...
+# The program has been named "External Code" for simiplicity and is used to help the user
+# trace where variables come from. It is a simple script that slightly modifies
+# a file provided to it.
+#
+# There are 3 dataclasses containing the commandline options for the code,
+# the input parameters and the output parameters.
+
+# %%
+PRG_NAME = "External Code"
+BINARY = ["python", "ext_code_script.py"]
```
hsaunders1904, line 73: I'm getting an error when I run the example from the root bluemira directory. Could we get the path to the executable script relative to the example file, or will that mess things up if we're generating the notebook?

hsaunders1904 (changes requested):
A few comments, but looks good.

When we print the table, if the "description" or "long_name" are missing, we print `N/A` in the cell, but when a "value" is missing, we leave the cell empty. Is there a distinction between the cases that means we want the different behaviour?

hsaunders1904 (commented):
Oh, also this is producing several extra files within the repo. Could we set the read/write directories to somewhere else? Not sure the best place, maybe `generated_data` or `os.getenv("TMPDIR", "/tmp")`?

examples/codes/external_code.ex.py
je-cook: it should be possible because it will be in the same folder as the script

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+
+# %%
+from ext_code_script import get_filename
+
+from bluemira.base.file import get_bluemira_path
+from bluemira.base.parameter_frame import Parameter, ParameterFrame
+from bluemira.codes.interface import CodesSetup, CodesSolver, CodesTask, CodesTeardown
+from bluemira.codes.interface import RunMode as BaseRunMode
+from bluemira.codes.params import MappedParameterFrame
+from bluemira.codes.utilities import ParameterMapping
```
hsaunders1904, line 55: Any reason these imports are in separate cells?

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+
+# %% [markdown]
+# # External Code Wrapping
+#
+# This example goes though the minimal steps taken to wrap an external code and
+# retrive its outputs
```
hsaunders1904, line 61:
```suggestion
# retrieve its outputs.
```

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+# Firstly we define the options available in the code and its name.
+# In this example we're wrapping the python script
+# [ext_code_script.py](ext_code_script.py).
+# The script reads in a file and writes out to a different file with
+# two possible modifications:
+#   * Add a header
+#   * Add line numbers
+#
+# Unusually the BINARY global variable is a list instead of a string because there are
+# two base commands to run the script from the commandline. We have used a little hack
```
hsaunders1904, line 72:
```suggestion
# two base commands to run the script from the command line. We have used a little hack
```

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+# with the `get_filename` function so you don't have to find the external code script
+# file location.
+# Usually the program is in your PATH or provided by the config.
+#
+# The program has been named "External Code" for simiplicity and is used to help the
```
hsaunders1904, line 77:
```suggestion
# The program has been named "External Code" for simplicity and is used to help the
```

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+# user trace where variables come from. It is a simple script that slightly modifies
+# a file provided to it.
+#
+# There are 3 dataclasses containing the commandline options for the code,
```
hsaunders1904, line 81:
```suggestion
# There are 3 dataclasses containing the command line options for the code,
```

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+
+# %% [markdown]
+# ## Linking the code to bluemira
+#
+# To link an external code to bluemira you need a few bits of machinery
+#
+# * A `MappedParameterFrame` that links bluemira parameter names and units to the
+#   external code
+# * A `RunMode` class to specify the possibly running modes
+# * A `Solver` that orchistrates the running of the code
```
hsaunders1904, line 125:
```suggestion
# * A `Solver` that orchestrates the running of the code
```

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+# * Some task for the `Solver` to run typically there are three:
```
hsaunders1904, line 126:
```suggestion
# * Some task, or tasks, for the `Solver` to run. Typically there are three:
```

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+
+    RUN = auto()
+    READ = auto()
+    MOCK = auto()
+
+
+# %% [markdown]
+# The `Setup` class pulls over the inputs from bluemira as described by the
+# mapping and creates the input file. The output of the run method returns
+# the commandline options list.
```
hsaunders1904, line 181:
```suggestion
# the command line options list.
```

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+        super().__init__(params, PRG_NAME)
+        self.problem_settings = problem_settings
+        self.infile = infile
+
+    def update_inputs(self) -> Dict:
+        """Update inputs from bluemira"""
+        self.inputs = ECInputs()
+        self.options = ECOpts()
+        inp = self._get_new_inputs()
+        for thing in (self.problem_settings.items(), inp.items()):
```
hsaunders1904, line 199: Is there a more descriptive name than `thing` we could use? Are they `code_input`s?

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+            for k, v in thing:
+                if k in self.options.__annotations__:
```
hsaunders1904, line 201: Is  `[f.name for f in dataclasses.fields(self.options]` a more idiomatic way of retrieving this list?

Also, is my understanding right, in that we're doing two different things with `options` and `inputs`, and that's what this `if` is switching on? If so, I think it may be clearer to just have two separate loops, then we don't need the `if` at all

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+                elif v is not None:
+                    getattr(self.inputs, k)
+                    setattr(self.inputs, k, v)
+
+        return {k: v for k, v in asdict(self.inputs).items() if v}
+
+    def run(self) -> List:
+        """Run mode"""
+        inp = self.update_inputs()
+        with open(self.infile, "w") as _if:
```
hsaunders1904, line 215:
```suggestion
        with open(self.infile, "w") as input_file:
```

I think this is a bit more descriptive, and then we don't need the leading `_` on the `if` - I was initially a bit confused why we needed a private variable here, until I realised it was to allow calling the variable `if`

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+            for k, v in inp.items():
+                _if.write(f"{k}  {v}\n")
+        return self.options.to_list()
+
+
+# %% [markdown]
+# `Run` simply runs the code in a subprocess with the given options
```
hsaunders1904, line 222:
```suggestion
# `Run` simply runs the code in a subprocess with the given options.
```

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+        self.outfile = outfile
+
+    def run(self, options: List):
+        """Run mode"""
+        self._run_subprocess([*self.binary, *options, self.infile, self.outfile])
+
+
+# %% [markdown]
+# `Teardown` reads in a given output file or in the case of mock returns a known
+# value sending the new parameter values back to the `ParameterFrame`
```
hsaunders1904, line 243:
```suggestion
# `Teardown` reads in a given output file or, in the case of mock, returns a known
# value, sending the new parameter values back to the `ParameterFrame`.
```

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+class Teardown(CodesTeardown):
+    """Teardown task"""
+
+    def __init__(self, params: ParameterFrame, outfile: str):
+        super().__init__(params, PRG_NAME)
+        self.outfile = outfile
+
+    def _read_file(self):
+        out_params = {}
+        with open(self.outfile, "r") as of:
```
hsaunders1904, line 255: To be consistent with my suggestion above.

```suggestion
        with open(self.outfile, "r") as output_file:
```

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+
+    name = PRG_NAME
+    params = ECParameterFrame
+    setup_cls = Setup
+    run_cls = Run
+    teardown_cls = Teardown
+    run_mode_cls = RunMode
+
+    def __init__(self, params: Union[ParameterFrame, Dict], build_config: Dict):
+
```
hsaunders1904, line 297:
```suggestion
```

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+    def mock(self) -> ParameterFrame:
+        """Mock mode"""
+        self._update_params_with_outputs({"param1": 15})
+        return self.params
+
+
+# %% [markdown]
+# `Solver` combines the three tasks into one object for execution.
+# The execute method has been overridden here for our usecase and returns
```
hsaunders1904, line 282:
```suggestion
# The execute method has been overridden here for our use-case and returns
```

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+        if teardown := self._get_execution_method(self._teardown, run_mode):
+            result = teardown()
+        return result
+
+
+# %% [markdown]
+# ### Using the solver
+#
+# To run the solver you just need to provide the parameters and the configuration
+# to initialse the object.
```
hsaunders1904, line 330:
```suggestion
# to initialise the object.
```

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+# Be aware `problem_settings` should be used sparingly for options that won't
+# change within the rerunning of the solver,
+# it has the same effect as modifying the default.
+# Also note the units of the parameter values returned back have been updated.
+#
+# The files written and read by the external code are stored in the generated_data
+# folder in the root of the bluemira repository.
+#
+# Some warnings will be shown because some of the situations here are usually
+# undesireable.
```
hsaunders1904, line 340:
```suggestion
# undesirable.
```

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+solver = Solver(params, build_config)
+
+# Running in all the different modes
+for mode in ["run", "read", "mock"]:
+    print(mode)
+    out_params = solver.execute(mode)
+    print(out_params)
+
+# %% [markdown]
+# 1 warning this time, we still havent sent a value for `param1` in `run` mode.
```
hsaunders1904, line 378:
```suggestion
# 1 warning this time, we still haven't sent a value for `param1` in `run` mode.
```

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+# Notice how the default for `param2` from our `problem_settings` is used when we
+# turn off the send mapping
```
hsaunders1904, line 380:
```suggestion
# turn off the send mapping.
```

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+
+# %%
+solver.modify_mappings({"param2": {"send": False}})
+print(solver.execute("run"))
+
+# %% [markdown]
+# Again the same warning. This time we have modified the value of `param2` and turned
+# the send mapping back on
```
hsaunders1904, line 388:
```suggestion
# the send mapping back on.
```

hsaunders1904: I'll stop suggesting these now, might be worth going through and ensuring our markdown blocks end with a full-stop where they should

examples/codes/external_code.ex.py
```diff
@@ -0,0 +1,428 @@
...
+# %% [markdown]
+# Turning on the header option (that output wont change) but the source has changed
+# because we havent updated it
+
+# %%
+solver.params.header.value = True
+print(solver.execute("run"))
+
+# %% [markdown]
+# Now we set the `param2` source and only send and not recieve the result
```
hsaunders1904, line 413:
```suggestion
# Now we set the `param2` source and only send and not receive the result
```

hsaunders1904 (changes requested):
Proper look through this time. 

Example looks good 👍. A few typos/spellings to look at, and some minor suggestions.

examples/codes/external_code.ex.py
je-cook: Nope :)

examples/codes/external_code.ex.py
je-cook: the options and inputs can come from either problem_settings or inp but I've made it a bit clearer

hsaunders1904 (approved):

PR #1857, opened 2023-01-19
Allow Union in `Parameter` typing

je-cook:
## Description

Before this fix `Parameter[Union[str, MyEnum]` would fail because the type wouldnt be inspected. I have only done the first level of nesting `Union[Union[...,...], Union[...,...]]` etc wont work

## Checklist

### DIFF
--- a/bluemira/base/parameter_frame/_frame.py
+++ b/bluemira/base/parameter_frame/_frame.py
@@ -215,8 +215,14 @@ def _validate_parameter_field(field, member_type: Type) -> Tuple[Type, ...]:
         not hasattr(member_type, "__origin__") or member_type.__origin__ is not Parameter
     ):
         raise TypeError(f"Field '{field}' does not have type Parameter.")
-    value_types = get_args(member_type)
-    return value_types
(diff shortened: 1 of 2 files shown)


bluemira/base/parameter_frame/_frame.py
```diff
@@ -215,8 +215,14 @@ def _validate_parameter_field(field, member_type: Type) -> Tuple[Type, ...]:
...
-    value_types = get_args(member_type)
-    return value_types
+    value_types = list(get_args(member_type))
+    vt = []
+    for v in value_types:
+        if new_v := get_args(v):
+            vt.extend(new_v)
+        else:
+            vt.append(v)
+    return tuple(vt)
```
hsaunders1904, line 225: What happens here if the type is some container type other than `Union`? Something like `Tuple[str, int]`? Wouldn't we be allowing a `str` or `int`, where really we should be raising an error?

Maybe we should check the type is actually a `Union` before we unpack it; and add a test that we get an error for other container types.

hsaunders1904 (changes requested):

PR #1865, opened 2023-01-25
Add `KuiroukidisLCFS(GeometryParameterisation)`

CoronelBuendia:
## Linked Issues

Closes #{ID}

## Checklist

### DIFF
--- a/bluemira/equilibria/shapes.py
+++ b/bluemira/equilibria/shapes.py
@@ -108,7 +108,7 @@ def flux_surface_manickam(r_0, z_0, a, kappa=1, delta=0, indent=0, n=20):
     return Coordinates({"x": x, "z": z})
 
 
-def flux_surface_kuiroukidis(
+def flux_surface_kuiroukidis_quadrants(
     r_0,
     z_0,
     a,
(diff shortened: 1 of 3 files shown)


bluemira/equilibria/shapes.py
```diff
@@ -119,49 +120,6 @@ def flux_surface_kuiroukidis(
     n_power: int = 8,
     n_points: int = 100,
 ):
```
je-cook, line 122: Could we have a small docstring here even though technically its allowed without by flake, just to direct users to the below function that gives the explanation. I know its only used internally here and not aimed to be exposed but its a reasonably complicated function and because python its still importable.

```suggestion
):
"""
Kuiroukidis flux surface individual quadrants

please see :func:flux_surface_kuiroukidis for more information
"""
```

bluemira/equilibria/shapes.py
```diff
@@ -246,22 +201,198 @@ def flux_surface_kuiroukidis(
...
+    z_upper_right += z_0
+    z_upper_left += z_0
+    z_left += z_0
+    z_right += z_0
+    return [x_upper_right, x_upper_left, x_left, x_right], [
+        z_upper_right,
+        z_upper_left,
+        z_left,
+        z_right,
+    ]
```
je-cook, line 225: Feels cleaner
```suggestion

    return (
        np.array([x_upper_right, x_upper_left, x_left, x_right]),
        np.array([z_upper_right, z_upper_left, z_left, z_right]) + z_0,
    )
```

bluemira/equilibria/shapes.py
```diff
@@ -246,22 +201,198 @@ def flux_surface_kuiroukidis(
...
+        wires = []
+        labels = ["upper_outer", "upper_inner", "lower_inner", "lower_outer"]
+        for x_q, z_q, lab in zip(x_quadrants, z_quadrants, labels):
+            wires.append(
+                interpolate_bspline(
+                    np.array([x_q, np.zeros(len(x_q)), z_q]).T, label=lab
+                )
+            )
+
+        return BluemiraWire(wires, label=label)
```
je-cook, line 393: I think I slightly prefer this as its only creating a single list, feel free to ignore
```suggestion
        return BluemiraWire(
            [
                interpolate_bspline(
                    np.array([x_q, np.zeros(len(x_q)), z_q]).T, label=lab
                )
                for x_q, z_q, lab in zip(x_quadrants, z_quadrants, labels)
            ],
            label=label,
        )
```

je-cook (approved):
Looks good, just a few minors if you want

Issue #1893, opened 2023-02-01
Cannot deepcopy PictureFrame


hsaunders1904:
## Describe the bug

It is not possible to `deepcopy` a `PictureFrame`. It causes a `ValueError`.

## Steps to reproduce

```python
from copy import deepcopy

from bluemira.geometry.parameterisations import PictureFrame

p = PictureFrame()
p2 = deepcopy(p)
```

## Expected behaviour

The copy should succeed.

## Screenshots / Tracebacks

<details>

    ---------------------------------------------------------------------------
    ValueError                                Traceback (most recent call last)
    ----> 1 p2 = deepcopy(p)

    File ~/mambaforge/envs/bluemira/lib/python3.8/copy.py:172, in deepcopy(x, memo, _nil)
        170                 y = x
        171             else:
    --> 172                 y = _reconstruct(x, memo, *rv)
        174 # If is its own copy, don't memoize.
        175 if y is not x:

    File ~/mambaforge/envs/bluemira/lib/python3.8/copy.py:270, in _reconstruct(x, memo, func, args, state, listiter, dictiter, deepcopy)
        268 if state is not None:
        269     if deep:
    --> 270         state = deepcopy(state, memo)
        271     if hasattr(y, '__setstate__'):
        272         y.__setstate__(state)

    File ~/mambaforge/envs/bluemira/lib/python3.8/copy.py:146, in deepcopy(x, memo, _nil)
        144 copier = _deepcopy_dispatch.get(cls)
        145 if copier is not None:
    --> 146     y = copier(x, memo)
        147 else:
        148     if issubclass(cls, type):

    File ~/mambaforge/envs/bluemira/lib/python3.8/copy.py:210, in _deepcopy_tuple(x, memo, deepcopy)
        209 def _deepcopy_tuple(x, memo, deepcopy=deepcopy):
    --> 210     y = [deepcopy(a, memo) for a in x]
        211     # We're not going to put the tuple in the memo, but it's still important we
        212     # check for it, in case the tuple contains recursive mutable structures.
        213     try:

    File ~/mambaforge/envs/bluemira/lib/python3.8/copy.py:210, in <listcomp>(.0)
        209 def _deepcopy_tuple(x, memo, deepcopy=deepcopy):
    --> 210     y = [deepcopy(a, memo) for a in x]
        211     # We're not going to put the tuple in the memo, but it's still important we
        212     # check for it, in case the tuple contains recursive mutable structures.
[...]

PR #1894, opened 2023-02-01
Add new optimisation module

### LINKED ISSUE #1328: Documentation for and design of optimisation classes

The documentation for setting up an optimisation problem is currently lacking. We should address this.

We might also want to have a look at a couple of points in terms of the design:

- Should `OptimisationObjective.apply_objective` and `OptimsationConstraint.apply_constraint` be methods on `OptimisationProblem`? It feels more natural for these methods to live on the objects that are being mutated. 
- Is it better to have the objective function and constraints be outputs of implementations of abstract methods on the `OptimisationProblem` class, rather than defining the objective/constraints in the constructor and calling `super().__init__`? E.g.
    ```python 
    class OptimisationProblem(ABC):
        def __init__(self) -> None:
            self.opt = self.objective()
            self.apply_constraints(self.constraints)
            ...

        @abc.abstractmethod
        def objective(self, x: np.ndarray, grad: np.ndarray, **kwargs) -> OptimisationObjective:
            pass

        @abc.abstractmethod
        def constraints(self) -> List[OptimisationConstraint]:
            pass
    ```
- Is it necessary for users to define a new class for every optimisation problem? Could they compose their own by passing an optimiser, objective function and list of constraints to an `OptimisationProblem`? This could avoid a class explosion of optimisation problems. We may even be able to perform all of this in a single function call.
- We should 'enumify' some of the optimisation options (e.g., `constraint_type`, and the NLOpt algorithms)

## Outputs
- [ ] Documentation for, and examples of, setting up optimisation problems
- [ ] Discussions and decisions regarding the design of `OptimisationProblem` and related classes

hsaunders1904:
## Linked Issues

Closes #1328 

### !WIP!

Adds a new `optimisation` module, which exports functions `optimise` and `optimise_geometry`, along with an `Algorithm` enum that lists available algorithms.

The best place to start is probably the new [documentation](https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/hsaunders1904/1328_documentation_for_and_design_of_optimisation_classes/documentation/source/optimisation/optimisation.rst) and [examples](https://github.com/Fusion-Power-Plant-Framework/bluemira/tree/hsaunders1904/1328_documentation_for_and_design_of_optimisation_classes/examples/optimisation), which show how to use the API.

### TODO

- [ ] Error handling; we need to achieve parity with the old optimisers in terms of error messages.
- [ ] RST documentation for `optimise_geometry`.
- [ ] Tests using equality constraints.
- [x] ~Caching f(x) and passing to approx_gradient to avoid recalculation of objective and constraints. This was done previously, but there was a bug somewhere (99f6a997f24db0862d1bb30e5f72d6d34ccd0855)~. [72f99bf](https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/1894/commits/72f99bfff49e11a3f44bdbf8ab3a190fa5d01fde)
- [ ] Allow customisation of shape discretisation in keep-out/in zones.

## Interface Changes

No changes, but have added a new way to do optimizations.

## Checklist


bluemira/optimisation/_algorithm.py
```diff
@@ -0,0 +1,73 @@
...
+        ValueError
+            If the string does not correspond to an enum name.
+        """
+        try:
+            return getattr(cls, s)
+        except AttributeError:
+            if s == "DIRECT-L":
+                # special case for backward compatibility
+                return Algorithm.DIRECT_L
+            raise ValueError(f"No such Algorithm value '{s}'.")
```
je-cook, line 73: I dont know whether it might be better to modify the `__getitem__` in a metaclass because currently this is possible
Algorithm["DIRECT_L"] but Algorithm["DIRECT-L"] would fail

```python
class AlgorithmMeta(enum.EnumMeta):
    def __getitem__(self, s: str) :
        try:
            return super().__getitem__(s)
        except AttributeError:
            if s == "DIRECT-L":
                # special case for backward compatibility
               return self.DIRECT_L
            raise ValueError(f"No such Algorithm value '{s}'.")

class Algorithm(enum.Enum, metaclass=AlgorithmMeta):
```

examples/optimisation/nonlinearly_constrained_problem.ex.py
```diff
@@ -0,0 +1,86 @@
+# %%
```
je-cook, line 1: Needs this block from https://github.com/Fusion-Power-Plant-Framework/bluemira/tree/develop/examples#maintaining-the-examples
```suggestion
# ---
# jupyter:
#   jupytext:
#     cell_metadata_filter: tags,-all
#     notebook_metadata_filter: -jupytext.text_representation.jupytext_version
#     text_representation:
#       extension: .py
#       format_name: percent
#       format_version: '1.3'
#   kernelspec:
#     display_name: Python 3 (ipykernel)
#     language: python
#     name: python3
# ---

# %% tags=["remove-cell"]
```

bluemira/optimisation/_optimiser.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+
+    @abc.abstractmethod
+    def optimise(self, x0: Optional[np.ndarray] = None) -> OptimiserResult:
+        """
+        Run the optimiser.
+
+        Parameters
+        ----------
+        x0: Optional[np.ndarray]
```
je-cook, line 137: Just fyi we don't need to type in both the function definition and the docstrings anymore. The documentation build automatically will populate both if its in the function definition. 
Not really important though and in some cases we can be more descriptive in the docstring. If the type already exists in the docstring I dont think its replaced by the function definition type.
```suggestion
        x0
```

je-cook (commented):
Some little ones to start, I'm about a third of the way through

examples/optimisation/nonlinearly_constrained_problem.ex.py
hsaunders1904: Done, I think

bluemira/optimisation/_nlopt/constraint.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+class ConstraintType(enum.Enum):
+    """Enumeration of constraint types."""
+
+    EQUALITY = enum.auto()
+    INEQUALITY = enum.auto()
+
+
+class Constraint:
```
je-cook, line 37: This could be premature optimisation (pun unintended :smile:), but this class and `NloptObjectiveFunction` in `objective.py` seem very similar, could they have a common base or is the more trouble than its worth?

documentation/source/examples.rst
```diff
@@ -56,6 +56,8 @@ Other Examples
     examples/magnetostatics/helmholtz_example
     examples/materials/material_definitions
     examples/mesh/mesh_tutorial
+    examples/optimisation/geometry_optimisation
+    examples/optimisation/nonlinearly_constrained_problem
```
je-cook, line 60: We may want to consider putting one of these or a different one in introductory examples

tests/optimisation/test_geometry.py
```diff
@@ -0,0 +1,201 @@
...
+            {
+                "x1": {"value": 5.5, "upper_bound": 10, "lower_bound": 5},
+                "x2": {"value": 16, "upper_bound": 19.5, "lower_bound": 12.5},
+                "dz": {"value": 1, "upper_bound": 1.5, "lower_bound": -1.5},
+            }
+        )
+        # TODO(hsaunders1904): think about whether we want to keep the
+        # original parameterisation constant, or whether we change it
+        # in-place. Add a test for the behaviour
+        original_length = parameterisation.create_shape().length
```
je-cook, line 88: As you obviously are, I am in two minds. I think currently we modify in place though

je-cook (commented):
OK I've been through everything once, looks really good. I havent played with it yet but I will do soon.

bluemira/optimisation/_nlopt/constraint.py
hsaunders1904: You're right, the gradient stuff, at the very least, should be refactored into something common.

tests/optimisation/test_geometry.py
hsaunders1904: I think the only reason I'm currently modifying the input, is that I can't actually copy a geometry parameterisation. I meant to open an issue for this, but `OptVariable` is not copyable; I think because it overrides `__getattribute__` and doesn't implement a `__deepcopy__`. I would have gone with copying, but because of this and #1893, it was easier to just leave it

hsaunders1904:
Mostly as a reminder to myself: there's an issue with the documentation build. I can't get relative path links between notebooks working, I want to link from a deprecated tutorial to the new one. 

There's also flake8 issues that line lengths are too long in some of the examples. This is a bit annoying in `examples/optimisation/nonlinearly_constrained_problem.ex.py` where a URL link means the line is too long, but if I put a `noqa` at the end, the markdown just has a random `noqa` in it. Maybe use a URL shortener?

je-cook:
Not a huge fan on the url shortener, two options as I see it add a per file ignore to `.flake8` for that example and just deal with the modifications or just link to the whole tutorial section rather than specific header? 
I think the first option is probably best, can't find an ignore urls think for flake which would be good!

bluemira/builders/tf_coils.py
```diff
@@ -543,7 +543,7 @@ class RippleConstrainedLengthGOP(GeometryOptimisationProblem):
     ----------
     parameterisation: GeometryParameterisation
         Geometry parameterisation for the winding pack current centreline
-    optimiser: Optimiser
+    optimiser: bluemira.utilities.optimiser.Optimiser
```
je-cook, line 546: It might be worth just moving all of these to typings?

bluemira/equilibria/opt_problems.py
```diff
@@ -83,7 +83,7 @@ class CoilsetOptimisationProblem(OptimisationProblem):
     ----------
     coilset: Coilset
         Coilset to be optimised.
-    optimiser: Optimiser (default: None)
+    optimiser: bluemira.utilities.optimiser.Optimiser (default: None)
```
je-cook, line 86: Same here (if you can be bothered now). We could do it later

bluemira/optimisation/_geometry/_typing.py
```diff
@@ -0,0 +1,50 @@
...
+# License along with bluemira; if not, see <https://www.gnu.org/licenses/>.
+from typing import Protocol, TypedDict
+
+import numpy as np
+from typing_extensions import NotRequired
+
+from bluemira.geometry.parameterisations import GeometryParameterisation
+
+
+class GeomOptimiserObjective(Protocol):
```
je-cook, line 29: Nice to see this getting started too!

je-cook (approved):
Looks so good! I have no real comments :)

hsaunders1904:
I've opened #2189 which lists the tasks for the feature branch. I've added everything from the list of tasks in this PR and everything from the PR comments to the new issue.

Issue #1685, opened 2022-12-01
Update default PROCESS install to more recent version


je-cook:
## Description of issue / requirement to address

We should keep up with the PROCESS changes. A lot has changed recently with PROCESS so it will probably need a little bit of work.

## Proposed solution

Ask PROCESS to cut a new release. Change our default installation to that version

## Alternative solutions

Leave it on v2.3

PR #1895, opened 2023-02-02
⬆️ use the PROCESS hotfix tag

je-cook:
## Description

PROCESS install has needed to be updated before a release because the FORD repository has moved which breaks the PROCESS install. This hotfix tag slightly ahead of 2.3.0. Once PROCESS has a new release and #1685 is done this is irrelevant.

## Checklist

### DIFF
--- a/scripts/install-process.sh
+++ b/scripts/install-process.sh
@@ -7,7 +7,7 @@
 set -e
 
 # Ensure we are working in the (bluemira) conda environment
-source ~/.miniforge-init.sh && conda activate bluemira
+source ~/.mambaforge-init.sh && conda activate bluemira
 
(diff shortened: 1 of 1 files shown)


hsaunders1904 (changes requested):
There's an issue in the `install-process` script. It's still trying to use `.miniforge-init.sh`, but we now create `.mambaforge-init.sh` in our `install-conda` script.

After switching to `mambaforge` it all seems to work though.

I still see the version as `v2.3.0`, with no `hotfix` prefix. Maybe that's expected?
```
> pip list | grep process
process                       2.3.0
```

je-cook:
yep I do too, It probably should be a new tag but that is the tag they've given us

hsaunders1904 (approved):

PR #1918, opened 2023-02-06
✨ Improve STEP file export and add new export filetypes

je-cook:
## Description

This adds how FreeCAD exports STEP files instead of as they call it the 'legacy' version. 
We now have more customisation over STEP files like author, scheme and labelling objects with the possibility of adding colours at some point is suggested to be possible.

I've also added a few other filetypes based on the freecad standard export docs: https://wiki.freecadweb.org/Import_Export/en. There are some that need more FreeCAD than we import or extra modules but they should error gracefully when used through the `save_cad` function.

## Interface Changes

Addition of `save_cad` function. Old `save_as_STP` not removed as it is still a valid save method.

Units of STEP files are now consistent with our units (the files we were exporting were in mm)

## Checklist


bluemira/codes/_freecadapi.py
```diff
@@ -1084,8 +1269,17 @@ def save_as_STP(shapes, filename="test", scale=1):
         Full path filename of the STP assembly
     scale: float (default 1)
         The scale in which to save the Shape objects
+
+    Notes
+    -----
+    This uses the legacy method to save to STP files.
+    It doesnt require freecad documents but also doesnt allow much customisation
+    The scale of Part is in mm by default therefore we scale by 1000 to convert
+    to metres.
```
CoronelBuendia, line 1278: What are the pros and cons here? I vaguely remember us wanting to avoid FreeCAD documents, for some reason, but perhaps that was mostly for viewing CAD, not sure.

bluemira/geometry/tools.py
```diff
@@ -932,10 +932,13 @@ def save_cad(
...
     filename: str
         Full path filename of the STP assembly
-    scale: float, default = 1.0
-        The scale in which to save the Shape objects
+    formatt
+        file format to save as
+    kwargs
+        arguments passed to cadapi save function
     """
-    save_as_STP(get_properties_from_components(components, "shape"), filename, scale)
```
CoronelBuendia, line 938: So this removes the "legacy" method from the user-facing API. Question: do we want to continue to expose the legacy method? This has the potential to break some workflows.

bluemira/codes/_freecadapi.py
```diff
@@ -1098,11 +1292,94 @@ def save_as_STP(shapes, filename="test", scale=1):
...
+        for no, obj in enumerate(objs):
+            objs[no].Shape = scale_shape(obj.Shape.copy(), scale)
+
+
+def save_cad(
+    shapes: Iterable[apiShape],
+    filename: str,
+    formatt: Union[str, CADFileType] = "stp",
+    labels: Iterable[str] = None,
+    scale: int = 1000,
```
CoronelBuendia, line 1331: perhaps use `raw_uc` here? just to make it really explicit from which to which.

CoronelBuendia (commented):
Really superficial review from my side. Implementation looks thorough, but I have not played with it.

From my side, if this does not break any workflows, and improves our saving CAD geometries to files, then I think it is great.

@ivanmaione I reckon should take a look though, as he will perhaps know why we avoided the document route the first time round.

bluemira/codes/_freecadapi.py
je-cook: We already use FreeCAD documents in the CAD viewer. The document is initialised in the viewer we just dont use them when creating our CAD. The FreeCAD document interface isnt very pythonic, its a straight conversion from C++ and also is another layer of abstraction, its I think why we use `Part` directly.

Possibly there could be some issues when running on something without a display but I don't think so as we explicitly initialise without a GUI, and there are probably other errors we havent dealt with in that area. 

In any event the tests pass here when there definitely isnt a display.

bluemira/geometry/tools.py
je-cook: From our basic saving yes, it is still possible to use some interface to the legacy save STP because one the kwargs is legacy (see https://github.com/FreeCAD/FreeCAD/blob/431fafc5337976c43bf50aae7bc75b46ee8037c7/src/Mod/Import/Gui/AppImportGuiPy.cpp#L559). It will use the new interface but the old export method.

I havent documented that but I could?

Alternatively I could add another file type (eg STEP_LEGACY) that uses our old interface which is the `exportSTEP()` method on any underlying Part object.

je-cook: Also I havent removed the save_as_STP function so as long as it is a list of `BluemiraGeo` objects that will still work

je-cook:
`pytest tests/geometry/test_tools.py::TestSavingCAD::test_save_as_STP` passes for me locally. I dont know why it would fail here or pass before I made this change...

bluemira/geometry/tools.py
CoronelBuendia: I don't mind either way, but check with the STEP guys JIC

bluemira/codes/_freecadapi.py
```diff
@@ -1161,14 +1317,96 @@ def save_as_STP(shapes: List[apiShape], filename: str = "test", scale: float = 1
...
+        CAD shape objects to save
+    filename
+        filename (file extension will be forced base on `formatt`)
+    formatt
+        file formatt
+    labels
+        shape labels
+    unit_scale
+        unit to save the objects as.
+    kwargs
```
hsaunders1904, line 1376: Missing colons on these

tests/codes/test_freecadapi.py
```diff
@@ -254,3 +255,138 @@ def func():
...
+            "OPENSCAD",
+            "SIMPLE_MODEL",
+            "STEP",
+            "STEP_2",
+            "STL",
+            "THREED_MANUFACTURING",
+            # ifcopenshell package required
+            pytest.param("IFC_BIM", marks=[pytest.mark.xfail]),
+            pytest.param("IFC_BIM_JSON", marks=[pytest.mark.xfail]),
+            # Part.Feature has no compatible object type TODO find compatible object type
```
hsaunders1904, line 354: Should we have these as options if we know they fail?

hsaunders1904 (approved):
Code looks good to me, one little question, but nothing major

tests/codes/test_freecadapi.py
je-cook: My take was more if they do start working it would be nice to be aware. Maybe there is a better way of flagging this than loads of xfails?

oliverfunk (commented):
I've looked through the code and ran the tests, which passed.
I can't comment on the impl. too much as I don't have much experience with this, the interface looks good and I did a test with save_cad for .stl and .obj and successfully opened them in Blender. They were tiny.
I could also view the test .stp file with some online step viewer

bluemira/geometry/tools.py
```diff
@@ -1032,20 +1035,22 @@ def save_as_STP(
...
-    freecad_shapes = [s.shape for s in shapes]
-    cadapi.save_as_STP(freecad_shapes, filename, scale)
+    cadapi.save_as_STP([s.shape for s in shapes], filename, unit_scale, **kwargs)
 
 
 def save_cad(
-    components: Union[Component, Iterable[Component]], filename: str, scale: float = 1
+    components: Union[Component, Iterable[Component]],
+    filename: str,
+    formatt: str = "stp",
```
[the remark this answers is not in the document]
je-cook: yeah your right done

hsaunders1904 (approved):

PR #1939, opened 2023-02-14
🐛 allow setting None when updating a parameterframe

je-cook:
## Description

When we are updating from an external code sometimes the value doesnt exist in the output we warn and set it to None . This will then fail when we update subsequent frames.

The fix allows the `None` to be passed through harmlessly 

## Checklist

### DIFF
--- a/bluemira/base/parameter_frame/_frame.py
+++ b/bluemira/base/parameter_frame/_frame.py
@@ -96,37 +96,38 @@ class ParameterFrame:
     def update_from_dict(self, new_values: ParamDictT):
         """Update from a dictionary representation of a ``ParameterFrame``"""
         for key, value in new_values.items():
-            param: Parameter = getattr(self, key)
             if "name" in value:
                 del value["name"]
-            value_type = _validate_parameter_field(key, self._types[key])
-            new_param = Parameter(name=key, **value, _value_types=value_type)
-            param.set_value(
-                new_param.value
-                if param.unit == "" or new_param.value is None
-                else new_param.value_as(param.unit),
-                new_param.source,
+            self._set_param(
+                key,
+                Parameter(
+                    name=key,
+                    **value,
+                    _value_types=_validate_parameter_field(key, self._types[key]),
+                ),
             )
-            if new_param.long_name != "":
-                param._long_name = new_param.long_name
-            if new_param.description != "":
-                param._description = new_param.description
 
     def update_from_frame(self, frame: ParameterFrame):
         """Update the frame with the values of another frame"""
         for o_param in frame:
             if hasattr(self, o_param.name):
-                param = getattr(self, o_param.name)
-                param.set_value(
-                    o_param.value
-                    if param.unit == "" or o_param.value is None
-                    else o_param.value_as(param.unit),
-                    o_param.source,
-                )
-                if o_param.long_name != "":
-                    param._long_name = o_param.long_name
-                if o_param.description != "":
-                    param._description = o_param.description
+                self._set_param(o_param.name, o_param)
+
+    def _set_param(self, name: str, o_param: Parameter):
+        """
+        Sets the information from a Parameter to an existing Parameter in this frame.
+        """
+        param = getattr(self, name)
+        param.set_value(
+            o_param.value
+            if param.unit == "" or o_param.value is None
+            else o_param.value_as(param.unit),
+            o_param.source,
+        )
+        if o_param.long_name != "":
+            param._long_name = o_param.long_name
+        if o_param.description != "":
+            param._description = o_param.description
 
     @classmethod
     def from_dict(
@@ -259,7 +260,6 @@ def _validate_parameter_field(field, member_type: Type) -> Tuple[Type, ...]:
 
 
 def _validate_units(param_data: Dict, value_type: Iterable[Type]):
-
     try:
         quantity = pint.Quantity(param_data["value"], param_data["unit"])
     except KeyError as ke:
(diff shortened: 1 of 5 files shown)


bluemira/base/parameter_frame/_frame.py
```diff
@@ -102,7 +102,9 @@ def update_from_dict(self, new_values: ParamDictT):
             value_type = _validate_parameter_field(key, self._types[key])
             new_param = Parameter(name=key, **value, _value_types=value_type)
             param.set_value(
-                new_param.value if param.unit == "" else new_param.value_as(param.unit),
+                new_param.value
+                if param.unit == "" or new_param.value is None
+                else new_param.value_as(param.unit),
```
hsaunders1904, line 107: We've got this logic in two places, I can see the other instance was already not matching this. I'm trying to think of the best way to refactor it out.

Maybe the `None` logic could go in `Parameter.set_value`? So, if the `Parameter`'s value is `None`, just return `None`, regardless of the unit (or maybe we should enforce that the units are compatible?). I can't remember what the `param.unit == ""` is for. I guess to allow the unit not to be specified in the dict if we just want to update the value?

Thoughts? I think if we are to not refactor the logic, we should at least leave a comment in both places saying that these should be consistent.

je-cook: the `unit == ""` will avoid all unit conversions on bools str etc. 

On the refactor yeah you're right. I'll have a play with what you suggested and see what comes out

je-cook: I've opted for a refactor within `ParameterFrame` because of setting long name and description which was already duplicated as you mentioned. 

I havent moved the None check into `Parameter` because we convert units before we set the value.
The other option is that `param.value_as` should only convert on that conditional. 
As I'm writing this I am thinking I could do that too.

Thoughts?

hsaunders1904: I think keeping the logic within `ParameterFrame`, as you have now, is fine. As you say, we need to do something about `param.value_as` when the value is `None`. Currently, you just get an error. Options are:

1. Just return `None` whatever the unit is.
2. Validate that the unit is compatible, before returning `None`, else throw.
3. Just throw an error (unless trying to retrieve the value as its existing unit?).

I'm struggling with which is best 😅. I'm leaning towards 2, as I think it's good for a user to know that they're trying to do something wrong. The value may not be `None` if they run the same code with different settings next time. Better to flag the bug early, I think.

je-cook: I agree, done the second option

bluemira/base/parameter_frame/_parameter.py
```diff
@@ -137,12 +137,28 @@ def value(self) -> ParameterValueType:
     def value(self, new_value: ParameterValueType):
         self.set_value(new_value, source="")
 
-    def value_as(self, unit: Union[str, pint.Unit]) -> ParameterValueType:
-        """Return the current value in a given unit"""
+    def value_as(self, unit: Union[str, pint.Unit]) -> Union[ParameterValueType, None]:
```
hsaunders1904, line 140: 😆 MyPy is not going to like this... Not sure that we have any other option though

bluemira/base/parameter_frame/_parameter.py
```diff
@@ -137,12 +137,28 @@ def value(self) -> ParameterValueType:
...
             return raw_uc(self.value, self.unit, unit)
         except pint.errors.PintError as pe:
             raise ValueError("Unit conversion failed") from pe
+        except TypeError:
+            if self.value is None:
+                try:
+                    raw_uc(1, self.unit, unit)
+                    return
+                except pint.errors.PintError as pe2:
+                    raise ValueError("Unit conversion failed") from pe2
```
hsaunders1904, line 159: I have a feeling we do something like this elsewhere. Could we refactor it into a function, maybe this?

```python
def units_compatible(unit_1: str, unit_2: str) -> bool:
```

I feel the intent of the code will be clearer this way.

bluemira/base/parameter_frame/_parameter.py
```diff
@@ -137,12 +137,28 @@ def value(self) -> ParameterValueType:
...
+        """
         try:
             return raw_uc(self.value, self.unit, unit)
         except pint.errors.PintError as pe:
             raise ValueError("Unit conversion failed") from pe
+        except TypeError:
+            if self.value is None:
+                try:
+                    raw_uc(1, self.unit, unit)
+                    return
```
hsaunders1904, line 157:
```suggestion
                    return None
```

Personal preference, but I think being explicit would be better in this case

bluemira/base/parameter_frame/_parameter.py
je-cook: I'll factor this out either way but the only other use I can find subtly different. In the first case we avoid the error and in the second we actively want it to be thrown:

https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/10c37558cc125b40150e27f82941b205da16a3df/bluemira/base/parameter_frame/_frame.py#L265

https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/917683ae45da037b4cbe21f9ddc6f16d4352cd21/bluemira/base/parameter_frame/_parameter.py#L156

je-cook: The two cases I have for the factoring out of the units check are a little different one uses the pint quantity thats produced and does no conversion while the other does the check

https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/917683ae45da037b4cbe21f9ddc6f16d4352cd21/bluemira/base/parameter_frame/_parameter.py#L156

https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/10c37558cc125b40150e27f82941b205da16a3df/bluemira/base/parameter_frame/_frame.py#L265

I'll factor it out anyway as it is a useful function to have

hsaunders1904 (approved):

PR #1945, opened 2023-02-17
Add support for optional component tree views

CoronelBuendia:
## Linked Issues

At present it is only possible to add a component tree on a builder with `self.component_tree([xz], [xy], [xyz])` if all views exist. There are cases where we will only want to add e.g. xz and xyz views to the resulting component. This is not possible at present.

Closes #{ID}

## Checklist


CoronelBuendia:
Nevermind, I think that was a lie

Issue #1901, opened 2023-02-03
Lower Port


je-cook:
Add Lower port designer and builder

PR #1948, opened 2023-02-17
Lower port designer

je-cook:
## Linked Issues

part of #1901 

## Description

Creates the designer for the lower port.

TODO:

- [ ] offset for allowing movement of the divertor through the port
- [ ] Test 

## Checklist


eudemo/eudemo/maintenance/lower_port.py
```diff
@@ -0,0 +1,122 @@
...
+    divertor_padding: Parameter[float]
+
+
+class LowerPortDesigner(Designer):
+    """
+    Lower Port Designer
+
+    Notes
+    -----
+    Retrictions on angle is ±45° because there are no protections
```
CoronelBuendia, line 52: Not clear where this angle is defined from. If you take it to e.g. the CoG of the divertor silhouette, then [0, -90] would be more appropriate. Not sure what you mean with restrictions keeping the port dimensions constant.

CoronelBuendia: So at present seems defined following the opposite of the normal convention, perhaps we can flip? Still don't see how we'd ever want to bridge though 0 though. Lower ports are either horizontal (angle=0) or not (angle < 0), or vertical (angle=-90).

eudemo/eudemo/maintenance/lower_port.py
```diff
@@ -0,0 +1,122 @@
...
+
+    param_cls: Type[ParameterFrame] = LowerPortDesignerParams
+
+    def __init__(
+        self,
+        params: Union[Dict, ParameterFrame],
+        build_config: Dict,
+        divertor_xz: BluemiraFace,
+        x_wall_inner: float,
+        x_wall_outer: float,
```
CoronelBuendia, line 65: I'm not sure what these mean, nor why they are two of them. If the trajectory returned is that of the divertor, then the first point of that trajectory could be the CoG of the divertor, and the next point is at angle downwards up until some radius at which the port kinks to horizontal, i.e. 3 points. Why 4 points here?

eudemo/eudemo/maintenance/lower_port.py
```diff
@@ -0,0 +1,122 @@
...
+        x_path = np.array(
+            [x_start, self.x_wall_inner, self.x_wall_outer, self.x_extrema]
+        )
+        z_lower_path = np.array([z_start, z_start, z_outer, z_outer])
+
+        traj = make_polygon(
+            Coordinates(
+                {
+                    "x": x_path,
+                    "z": z_lower_path - self.params.divertor_padding.value,
```
CoronelBuendia, line 98: Why is the trajectory not following e.g. the typical path of the divertor, along which we might be able to sweep in future?

eudemo/eudemo/maintenance/lower_port.py
```diff
@@ -0,0 +1,122 @@
...
+        build_config: Dict,
+        divertor_xz: BluemiraFace,
+        x_wall_inner: float,
+        x_wall_outer: float,
+        x_extrema: float,
+    ):
+        super().__init__(params, build_config)
+        self.divertor_xz = divertor_xz
+        self.x_wall_inner = x_wall_inner
+        self.x_wall_outer = x_wall_outer
```
CoronelBuendia, line 71: I note that I can quite quickly make an invalid face by specifying silly values here.

eudemo/eudemo/maintenance/lower_port.py
```diff
@@ -0,0 +1,122 @@
...
+                                + self.params.divertor_padding.value,
+                            )
+                        )[::-1],
+                    ),
+                }
+            ),
+            closed=True,
+        )
+
+        return BluemiraFace(port), traj
```
CoronelBuendia, line 122: Why are both the trajectory and the exclusion zone well below the divertor? Surely the exclusion zone should encompass the divertor, no?

CoronelBuendia (changes requested):

je-cook:
superceeded by #2108

Issue #1347, opened 2022-08-16
Topological Naming - the root cause of most CAD and geometry issues


je-cook:
## Describe the bug

We have many issues (and work arounds) associated with CAD, the open ones I can find are:
 #1319 #1309 #1108 #1090 #1071 #1045 #902 #682 #513 

These seem to all relate to the Topological naming of elements in FreeCAD. This is a known issue for FreeCAD and is one of the main aims for the v0.21/v1 release cycle which is aiming to be released in 2023.

For some background reading the current progress in fixing this is detailed here:
https://github.com/realthunder/FreeCAD_assembly3/wiki/Topological-Naming

with the current FreeCAD branch:
https://github.com/FreeCAD/FreeCAD/compare/master...development/toponaming

Some background on the FreeCAD wiki and forum.
https://wiki.freecadweb.org/Topological_data_scripting
https://wiki.freecadweb.org/Topological_Naming_Project
https://wiki.freecadweb.org/Topological_naming_problem
https://forum.freecadweb.org/viewtopic.php?f=10&t=27278

The workarounds we have are understandably brittle after reading the above, and the option of just caching the shapes as @ivanmaione is working on in #1295 based off the ADR #1196 will hopefully alleviate some of the problems although not all.

It would be nice to keep our Matryoshka CAD structure or at least return to it at a later date once the above issues are (more) solved.

EDIT: The FreeCAD api looks like there will be some changes, the current release notes are here: https://wiki.freecadweb.org/Release_notes_1.0

Work towards this is progressing (they even have a gantt chartt!! https://github.com/FreeCAD/FreeCAD/issues/8432#issuecomment-1452745031)

sebkahn:
Congrats on finding that !!

je-cook:
Current tracking issue https://github.com/FreeCAD/FreeCAD/issues 8432

luzpaz:
Just a heads up that Toponaming won't be fixed in FreeCAD v0.21.x  
Best place to track progress is https://github.com/orgs/FreeCAD/projects/2

je-cook:
Thanks! Will definitely keep an eye on that project.

PR #1979, opened 2023-02-27
Avoid some CAD bugs in EUDEMO

je-cook:
## Description

This is a work around for the bug @hsaunders1904  pointed out in #1965 and a workaround for the divertor dome.
If the outer divertor target was lower than the inner target the dome would be put on the wrong flux line.

There is still a bug in the blanket segmentation, for some reason the boolean cut produces a valid face and an invalid face (L204 eudemo/blanket.py).

## Checklist

### DIFF
--- a/bluemira/builders/plasma.py
+++ b/bluemira/builders/plasma.py
@@ -98,8 +98,8 @@ class PlasmaBuilder(Builder):
         lcfs: BluemiraWire
             LCFS wire
         """
-        inner = make_circle(lcfs.bounding_box.x_min, axis=[0, 1, 0])
-        outer = make_circle(lcfs.bounding_box.x_max, axis=[0, 1, 0])
+        inner = make_circle(lcfs.bounding_box.x_min)
+        outer = make_circle(lcfs.bounding_box.x_max)
         face = BluemiraFace([outer, inner], self.name)
         component = PhysicalComponent(self.LCFS, face)
         component.plot_options.face_options["color"] = BLUE_PALETTE["PL"]
--- a/eudemo/eudemo/ivc/divertor_silhouette.py
+++ b/eudemo/eudemo/ivc/divertor_silhouette.py
@@ -192,14 +192,19 @@ class DivertorSilhouetteDesigner(Designer[Tuple[BluemiraWire, ...]]):
         self, start: Sequence[float], end: Sequence[float], label: str
     ) -> BluemiraWire:
         """
-        Make a dome between the two given points
-        The dome shape follows a constant line of flux that is closest
-        to the input start coordinate. Finally, the nearset point on the
-        flux surface to the end point and the end point are joined.
-        """
-        # Get the flux surface that crosses the through the start point
-        # We can use this surface to guide the shape of the dome
+        Make a dome between the two given points.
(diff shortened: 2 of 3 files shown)


eudemo/eudemo/ivc/divertor_silhouette.py
```diff
@@ -199,7 +199,8 @@ def make_dome(
         """
         # Get the flux surface that crosses the through the start point
         # We can use this surface to guide the shape of the dome
-        psi_start = self.equilibrium.psi(*start)
+
+        psi_start = self.equilibrium.psi(*(start if start[1] < end[1] else end))
```
CoronelBuendia, line 203: Can you put a comment on this, please? I saw something about why somewhere, but if it's not tied to this line we're going to forget!

bluemira/builders/plasma.py
```diff
@@ -98,8 +98,8 @@ def build_xy(self, lcfs: BluemiraWire) -> PhysicalComponent:
         lcfs: BluemiraWire
             LCFS wire
         """
-        inner = make_circle(lcfs.bounding_box.x_min, axis=[0, 1, 0])
-        outer = make_circle(lcfs.bounding_box.x_max, axis=[0, 1, 0])
```
CoronelBuendia, line 102: Oops..

eudemo/eudemo/vacuum_vessel.py
```diff
@@ -107,9 +107,14 @@ def build_xz(
...
-        inner_vv = offset_wire(
-            self.ivc_koz, self.params.g_vv_bb.value, join="arc", open_wire=False
+        inner_vv = _offset_wire_discretised(
+            self.ivc_koz,
+            self.params.g_vv_bb.value,
+            join="arc",
+            open_wire=False,
+            ndiscr=600,
         )
+
```
CoronelBuendia, line 117: I can't help but feeling that we're going to want to fallback onto this the vast majority of the time. It might be time to revive `MixedWireMaker`, because otherwise these massive polygons are going to be dreadfully slow and ugly.

CoronelBuendia: Also, why is the automatic fallback failing? Is the wire produced allegedly valid, or are we not catching an error or what?

je-cook: > I can't help but feeling that we're going to want to fallback onto this the vast majority of the time. It might be time to revive `MixedWireMaker`, because otherwise these massive polygons are going to be dreadfully slow and ugly.
It does feel like that could be a way out of this until #1347 is sorted

> Also, why is the automatic fallback failing? Is the wire produced allegedly valid, or are we not catching an error or what?

The wire is valid but it has a loop in it.

This is the surface that gets produced
Screenshot vv

which is made from a wire that is first offset and then vary offset.
from left to right; variable offset, offset_wire, original wire
Screenshot vv3

CoronelBuendia: That's wild... we could try increasing the wire fix tolerance in this case though. But anyway, this will do for now.

CoronelBuendia (approved):

Issue #1900, opened 2023-02-03
Equatorial Port


je-cook:
Add an equatorial port (a set nxn box)

je-cook:
Closed by #1989

PR #1989, opened 2023-03-03
Equatorial Port

IChiang99:
## Linked Issues

Linked Issue #1900 

## Description

Implemented a Designer and Builder for the Equatorial Port

## Interface Changes

N/A

## Checklist


eudemo/eudemo/equatorial_port.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+@dataclass
+class EquatorialPortDesignerParams(ParameterFrame):
+    """
+    Equatorial Port Designer parameters
+    """
+
+    x_ib: Parameter[float]
+    x_ob: Parameter[float]
```
je-cook, line 65: These should be inputs to the designer as they will not be known before any part of the design is known (it will be based off the previous stages

eudemo/eudemo/equatorial_port.py
```diff
@@ -0,0 +1,200 @@
...
+
+    param_cls: Type[EquatorialPortDesignerParams] = EquatorialPortDesignerParams
+
+    def __init__(
+        self,
+        params: Union[Dict, ParameterFrame, EquatorialPortDesignerParams],
+        build_config: Union[Dict, None],
+    ):
+        super().__init__(params, build_config)
+
```
je-cook, line 93: Obviously this will need to be threaded through.
eg:  
```suggestion
        build_config: Union[Dict, None],
        x_ib: float,
        x_ob: float,
    ):
        super().__init__(params, build_config)
        self.x_ib = x_ib
        self.x_ob = x_ob


```

eudemo/eudemo/equatorial_port.py
```diff
@@ -0,0 +1,200 @@
...
+        self.xz_profile = xz_profile
+
+    def build(self) -> Component:
+        """
+        Build the equatorial port component.
+        """
+
+        # TODO: Integrate corner radii
+        self.r_rad = self.params.ep_r_corner.value
+        self.coords = self.params.ep_coordinates.value
```
je-cook, line 137: having to specify all the coords for the user is a big ask. I think this needs to be the inboard x coordinate, the outboard x coordinate and the z offset of a single castellation. We should be able to build up the castellation from that.

eudemo/eudemo/equatorial_port.py
```diff
@@ -0,0 +1,200 @@
...
+        for r in new_coords:
+            x = r[0]
+            y = [r[1], r[1], r[2], r[2]]
+            z = [r[3], r[4], r[4], r[3]]
+            ext_vec = (x_ob - x, 0, 0)
+            section = BluemiraFace(make_polygon({"x": x, "y": y, "z": z}, closed=True))
+            sections.append(extrude_shape(section, ext_vec))
+
+        ep_shape = boolean_fuse(sections)
+        return circular_pattern(ep_shape, n_shapes=n_ep)
```
je-cook, line 186: circular_pattern_component makes this easier as we can just specify the number of TF coils as an input parameter see this function: https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/ad0ee0bf19bb4e7a1f934a6c7e6b5a63d64f8969/bluemira/builders/divertor.py#L90

je-cook (changes requested):
Thanks very much for this, looks like a good start!
The designer is essentially done and tested just a minor point from me. 
The builder needs a few adjustments

eudemo/eudemo/equatorial_port.py
IChiang99: I agree, but might keep the castellation method handy for future reference.

As well as the specified coordinates, wouldn't it also be necessary to specify the x-position of the castellation? Or is that inferred as halfway or at the outer boundary of the VV?

je-cook: Yes true, so 2 x coordinates one for inside and one for outside of a given vessel. 2 offsets for a single castellation 1 in z and 1 in x. 
I would have them as separate params

je-cook:
Can you rebase off develop again I think we've fixed the CI now (:crossed_fingers:)

IChiang99 (commented):
Have changed parameters, circular_pattern, and castellation method per previous PR review comments

eudemo/eudemo/equatorial_port.py
```diff
@@ -0,0 +1,229 @@
...
+        self.xz_profile = xz_profile
+        self.y_width = y_width
+        self.x_off = x_offsets
+        self.cst = castellation_offsets
+
+    def build(self) -> Component:
+        """
+        Build the equatorial port component.
+        """
+        # TODO: Implement corner radii
```
je-cook, line 143: #1992 should help with this, but dont worry about it now

eudemo/eudemo/equatorial_port.py
```diff
@@ -0,0 +1,229 @@
...
+        )
+        return ep_boundary
+
+
+class EquatorialPortBuilder(Builder):
+    """
+    Equatorial Port Builder
+    """
+
+    NAME_XY = "Equatorial Ports xy"
```
je-cook, line 119: I dont think its necessary to have both a name_xy and name_xyz they are automatically put into the xy or xyz part of the tree:
```
EquatorialPort (Component)
├── xz (Component)
│   └── koz (PhysicalComponent)
├── xy (Component)
│   ├── Sector 1 (Component)
│   │   └── Equatorial Ports xy 1 (PhysicalComponent)
│   ├── Sector 2 (Component)
│   │   └── Equatorial Ports xy 2 (PhysicalComponent)
│   ├── Sector 3 (Component)
│   │   └── Equatorial Ports xy 3 (PhysicalComponent)
│   ├── Sector 4 (Component)
│   │   └── Equatorial Ports xy 4 (PhysicalComponent)
│   ├── Sector 5 (Component)
│   │   └── Equatorial Ports xy 5 (PhysicalComponent)
│   ├── Sector 6 (Component)
│   │   └── Equatorial Ports xy 6 (PhysicalComponent)
│   ├── Sector 7 (Component)
│   │   └── Equatorial Ports xy 7 (PhysicalComponent)
│   ├── Sector 8 (Component)
│   │   └── Equatorial Ports xy 8 (PhysicalComponent)
│   ├── Sector 9 (Component)
│   │   └── Equatorial Ports xy 9 (PhysicalComponent)
│   └── Sector 10 (Component)
│       └── Equatorial Ports xy 10 (PhysicalComponent)
└── xyz (Component)
    ├── Sector 1 (Component)
    │   └── Equatorial Ports xyz 1 (PhysicalComponent)

```

eudemo/eudemo/equatorial_port.py
```diff
@@ -0,0 +1,229 @@
...
+        xz_profile = BluemiraWire(make_polygon({"x": x, "y": 0, "z": z}, closed=True))
+        body = PhysicalComponent(self.KOZ, xz_profile)
+        body.plot_options.face_options["color"] = BLUE_PALETTE["VV"][0]
+        return body
+
+    def build_xy(self, x: list, y: list, n_ep: int = 10) -> List[PhysicalComponent]:
+        """
+        Build the xy components of the equatorial port
+        """
+        ep_xy_profile = BluemiraWire(make_polygon({"x": x, "y": y, "z": 0}, closed=True))
```
je-cook, line 178: It looks like both the xy and xz views dont follow the same shape as the xyz, I added `output.show_cad()` in your test_ep_builder (and only created one xyz shape). This is shown when pytest is run with the `--plotting-on` flag:

Screenshot from 2023-03-14 14-18-10

eudemo/eudemo_tests/test_maintenance.py
```diff
@@ -52,3 +63,85 @@ def test_dummy_blanket_port_opt(self):
...
+        self.designer.params.update_values(param_values)
+        self.designer.x_ib = xi
+        self.designer.x_ob = xo
+        output = self.designer.execute()
+
+        # x = (xi, xo, xo, xi)
+        # z = (-zh / 2, -zh / 2, zh / 2, zh / 2)
+        # expectation = BluemiraFace(make_polygon({"x": x, "y": 0, "z": z}, closed=True))
+
+        # assert math.isclose(output.length, expectation.length)
```
je-cook, line 94: this looks like it works now that the fix in https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/2009 is merged

eudemo/eudemo/equatorial_port.py
```diff
@@ -0,0 +1,229 @@
...
+        for r in coords:
+            x = r[0]
+            y = [r[1], r[1], r[2], r[2]]
+            z = [r[3], r[4], r[4], r[3]]
+            ext_vec = (self.x_ob - x, 0, 0)
+            section = BluemiraFace(make_polygon({"x": x, "y": y, "z": z}, closed=True))
+            sections.append(extrude_shape(section, ext_vec))
+
+        ep_shape = PhysicalComponent(self.NAME_XYZ, boolean_fuse(sections))
+        ep_shape.plot_options.face_options["color"] = BLUE_PALETTE["VV"][0]
```
je-cook, line 199: just for the 3D
```suggestion
        ep_shape.display_cad_options.color = BLUE_PALETTE["VV"][0]
```

eudemo/eudemo/equatorial_port.py
IChiang99: Resolved with a bit of zip() help

eudemo/eudemo/equatorial_port.py
```diff
@@ -0,0 +1,229 @@
...
+        build_config: Union[Dict, None],
+        xz_profile: BluemiraFace,
+        y_width: float,
+        x_offsets: list,
+        castellation_offsets: list,
+    ):
+        super().__init__(params, build_config)
+        self.xz_profile = xz_profile
+        self.y_width = y_width
+        self.x_off = x_offsets
```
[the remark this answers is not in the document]
IChiang99: Sure, I'll change these to relative be relative to the inboard x position

eudemo/eudemo/equatorial_port.py
IChiang99: Yes, but it means that both the xy profile components and the xyz 3d components get called `Equatorial Port 1` to ` Equatorial Port 10`. I've updated the tests so they're not retrieving the wrong component and asserting that the 2d xy profiles have no volume.

eudemo/eudemo_tests/test_maintenance.py
IChiang99: Great, changed the asserts to simple equality conditions

eudemo/eudemo/equatorial_port.py
IChiang99: Cool, I'll look at this later

IChiang99 (commented):
Comments resolved and new code pushed through to my branch

eudemo/eudemo_tests/test_maintenance.py
```diff
@@ -52,3 +63,96 @@ def test_dummy_blanket_port_opt(self):
...
+        z_len = zh
+
+        try:
+            assert output.length == 2 * (x_len + z_len)
+        except AssertionError:
+            assert math.isclose(output.length, 2 * (x_len + z_len))
+        try:
+            assert output.area == x_len * z_len
+        except AssertionError:
+            math.isclose(output.area, x_len * z_len)
```
je-cook, line 108: Ahh I meant here that the comment probably wasnt needed and could be removed. I would leave the isclose checks because they.re more likely to fail than not.

```suggestion
        assert math.isclose(output.length, 2 * (x_len + z_len))
        assert math.isclose(output.area, x_len * z_len)
```

eudemo/eudemo_tests/test_maintenance.py
```diff
@@ -52,3 +63,96 @@ def test_dummy_blanket_port_opt(self):
...
+        self.builder.x_off = x_offset
+        self.builder.cst = c_off
+        output = self.builder.build()
+        out_eq_port = output.get_component("xyz").get_component("Equatorial Port 1")
+        if out_eq_port is None:
+            out_eq_port = output.get_component("xyz").get_component("Equatorial Port")
+        try:
+            assert out_eq_port.shape.volume == exp_v
+        except AssertionError:
+            assert math.isclose(out_eq_port.shape.volume, exp_v)
```
je-cook, line 158: Same here
```suggestion
        assert math.isclose(out_eq_port.shape.volume, exp_v)
```

eudemo/eudemo/equatorial_port.py
```diff
@@ -0,0 +1,235 @@
...
+        yz_profile: BluemiraWire,
+        extrude_vec: tuple,
+        x_offsets: list,
+        castellation_offsets: list,
+    ):
+        super().__init__(params, build_config)
+        self.x_ib = x_ib
+        self.x_ob = x_ob
+        self.yz_profile = yz_profile
+        self.vec = extrude_vec
```
je-cook, line 138:
```suggestion
        self.vec = np.array(extrude_vec)
```

je-cook (approved):
Looks good to me @CoronelBuendia do you want one last look?

eudemo/eudemo/maintenance/equatorial_port.py
```diff
@@ -0,0 +1,229 @@
...
+            xy=self.build_xy(xy_slice),
+            xyz=self.build_xyz(xyz_solid),
+        )
+
+    def build_xz(self, xz: BluemiraFace) -> PhysicalComponent:
+        """
+        Build the xz representation of the castellated component
+        """
+        body = PhysicalComponent(self.NAME, xz)
+        body.plot_options.face_options["color"] = BLUE_PALETTE["VV"][0]
```
CoronelBuendia, line 169: please use:

```python
from bluemira.builders.tools import apply_component_display_options
```
for this, it sets the color correctly in the plotting and the CAD

eudemo/eudemo/maintenance/equatorial_port.py
```diff
@@ -0,0 +1,229 @@
...
+        return body
+
+    def build_xy(self, xy: BluemiraFace, n: int = 10) -> List[PhysicalComponent]:
+        """
+        Build the xy representation of the castellated component
+        """
+        body = PhysicalComponent(self.NAME, xy)
+        body.plot_options.face_options["color"] = BLUE_PALETTE["VV"][0]
```
CoronelBuendia, line 177: ditto here and below

eudemo/eudemo/maintenance/equatorial_port.py
```diff
@@ -0,0 +1,229 @@
...
+        self.vec = np.array(extrude_vec)
+        self.off = offsets
+        self.cst = castellation_offsets
+
+    def build(self) -> Component:
+        """
+        Build the castellated component.
+        """
+        # TODO: Implement corner radii using PR #1992
+        self.r_rad = self.params.cst_r_corner.value
```
CoronelBuendia, line 147: Note to future self: it is correct to have this parameter in this builder.

Explanation: I was originally thinking that you don't need this, seeing as the radii could be built into the profile input, however, if we want the radii to be same on all offset shapes (likely), then offsetting outwards would increase the radius.

Might be worth specifying in the builder docstring that the radii are handled by the builder.

eudemo/eudemo_tests/test_maintenance.py
```diff
@@ -52,3 +66,83 @@ def test_dummy_blanket_port_opt(self):
...
+        """Test Castellation Builder"""
+        y = (yw / 2, -yw / 2, -yw / 2, yw / 2)
+        z = (-zh / 2, -zh / 2, zh / 2, zh / 2)
+        yz_profile = BluemiraFace(make_polygon({"x": xi, "y": y, "z": z}, closed=True))
+
+        self.builder.length = xo - xi
+        self.builder.face = yz_profile
+        self.builder.vec = vec
+        self.builder.off = x_offsets
+        self.builder.cst = c_offsets
```
CoronelBuendia, line 143: This makes me uncomfortable :D. Personally, I like to test things via their intended API, here, rejigging stuff on instantiated things is not we normally interact with builders, and it makes knowledge of the internal design of the Builder required to understand the test, as opposed to beneficial.

eudemo/eudemo/maintenance/equatorial_port.py
```diff
@@ -0,0 +1,229 @@
...
+            vec /= vec_mag
+
+        parameter_array = list(zip(distances, offsets))
+        parameter_array.append((length, 1.0))
+        _prev_dist = 0
+        for dist, offset in parameter_array:
+            ext_vec = np.array(vec) * (dist - _prev_dist)
+            sections.append(extrude_shape(base, ext_vec))
+            base.translate(ext_vec)
+            base = BluemiraFace(offset_wire(base.deepcopy(), offset))
```
CoronelBuendia, line 227: Is the deepcopy necessary here? Perhaps I am missing something

eudemo/eudemo/maintenance/equatorial_port.py
```diff
@@ -0,0 +1,229 @@
...
+        sections = []
+
+        # Normalise vec
+        vec_mag = np.linalg.norm(vec)
+        if vec_mag != 1.0:
+            vec /= vec_mag
+
+        parameter_array = list(zip(distances, offsets))
+        parameter_array.append((length, 1.0))
+        _prev_dist = 0
```
CoronelBuendia, line 222: This like it could be done in a cuter way. The 1.0 at the end is just to enable your zipped loop to have a value with which to offset in the final iteration?

If it must be so, please use a final offset value of 0.0, as this won't do much unnecessary computation (at least compared to an offset of 1.0).

eudemo/eudemo/maintenance/equatorial_port.py
```diff
@@ -0,0 +1,229 @@
...
+        body.plot_options.face_options["color"] = BLUE_PALETTE["VV"][0]
+        return circular_pattern_component(body, n_children=n)
+
+    def build_xyz(self, xyz: BluemiraSolid, n: int = 10) -> List[PhysicalComponent]:
+        """
+        Build the xyz representation of the castellated component
+        """
+        cst_shape = PhysicalComponent(self.NAME, xyz)
+        cst_shape.display_cad_options.color = BLUE_PALETTE["VV"][0]
+        return circular_pattern_component(cst_shape, n_children=n)
```
CoronelBuendia, line 186: @oliverfunk is this still how we do sectored stuff nowadays?

CoronelBuendia (dismissed):
The plugs are looking good and it's nice and clean, I have one API-ish request, in terms of supporting different inputs for lazy people. A few general comments.

This needs a rebase too to use some of the newer stuff in terms of how we set up builders for EUDEMO.

eudemo/eudemo/maintenance/equatorial_port.py
```diff
@@ -0,0 +1,229 @@
...
+
+    def __init__(
+        self,
+        params: Union[Dict, ParameterFrame, CastellationBuilderParams],
+        build_config: Union[Dict, None],
+        length: float,
+        start_profile: BluemiraFace,
+        extrude_vec: tuple,
+        offsets: list,
+        castellation_offsets: list,
```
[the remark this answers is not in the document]
IChiang99: There's a fence-post problem in the suggestion, but that's easily solved by instead checking `sum(depth_offsets) < depth` so the final castellation isn't just a flared face at the port end. 

But this is somewhat solved by making the `depth_offsets` to be in absolute terms rather than relative. Then it's easy to check `depth_offsets[-1] < depth` and may make it clearer to the user when they've entered a final castellation depth that is at or beyond the total depth. I prefer absolute depths, especially since I imagine it could get tedious to adjust values if they're chained. The offset values I'm happy to keep relative since that's more intuitive and to reduce the risk a user designs a impractical/nonsense port plug where it castellates inwards and outwards.

eudemo/eudemo/maintenance/equatorial_port.py
IChiang99: May be worth pre-empting this by labelling or listing wires that require fillet/chamfer operations. 
It's not every wire that needs this (not the inboard and outboard face boundaries), so would like to ask if it's wires defining internal and external corners, or just external corners that would go into the list

eudemo/eudemo/maintenance/equatorial_port.py
oliverfunk: @IChiang99 Bluemira now works by patterning xyz components during show_cad because it cuts down on runtime and let's you specify the number of sectors you wish to view (see: https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/ab336083030976071a4fab9f0b8c26a92acabdab/bluemira/base/reactor.py#L147)

build_xyz should return just 1 castellation, positioned correctly within a single sector.
The sectors are defined by n_TF, sector angle by 360/n_TF.

je-cook (approved):
Merging as is as comments are generally covered. We may need to do some subtle modifications when we bring everything together. 

Thanks very much @IChiang99!

PR #1992, opened 2023-03-03
Add `fillet_wire_2D` and `chamfer_wire_2D`

CoronelBuendia:
## Linked Issues

Closes #{ID}

## Checklist

### DIFF
--- a/bluemira/codes/_freecadapi.py
+++ b/bluemira/codes/_freecadapi.py
@@ -26,6 +26,8 @@ Supporting functions for the bluemira geometry module.
 from __future__ import annotations
 
 import math
+import os
+import sys
 from copy import deepcopy
 from typing import Dict, Iterable, List, Optional, Tuple, Union
 
@@ -1355,7 +1357,12 @@ def fillet_wire_2D(wire: apiWire, radius: float, chamfer: bool = False) -> apiWi
             f"Cannot {'chamfer' if chamfer else 'fillet'} a non-planar wire!"
         )
 
-    return DraftGeomUtils.filletWire(wire, radius, chamfer=chamfer)
+    # Temporarily suppress pesky print statement
+    sys.stdout = open(os.devnull, "w")
+    result = DraftGeomUtils.filletWire(wire, radius, chamfer=chamfer)
+    sys.stdout = sys.__stdout__
+
+    return result
 
 
 # ======================================================================================
--- a/bluemira/codes/_freecadapi.py
+++ b/bluemira/codes/_freecadapi.py
@@ -1347,9 +1347,12 @@ def fillet_wire_2D(wire: apiWire, radius: float, chamfer: bool = False) -> apiWi
     """
     # Temporarily suppress pesky print statement:
     # DraftGeomUtils.fillet: Warning: edges have same direction. Did nothing
-    sys.stdout = open(os.devnull, "w")
-    result = DraftGeomUtils.filletWire(wire, radius, chamfer=chamfer)
-    sys.stdout = sys.__stdout__
+    old_stdout = sys.stdout
+    try:
+        sys.stdout = open(os.devnull, "w")
+        result = DraftGeomUtils.filletWire(wire, radius, chamfer=chamfer)
+    finally:
+        sys.stdout = old_stdout
 
     return result
 
--- a/bluemira/codes/_freecadapi.py
+++ b/bluemira/codes/_freecadapi.py
@@ -1346,7 +1346,8 @@ def fillet_wire_2D(wire: apiWire, radius: float, chamfer: bool = False) -> apiWi
     """
     edges = wire.OrderedEdges
     if len(edges) < 2:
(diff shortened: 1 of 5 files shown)


bluemira/codes/_freecadapi.py
```diff
@@ -1324,6 +1326,31 @@ def sweep_shape(profiles, path, solid=True, frenet=True):
         return solid_result.Shells[0]
 
 
+def fillet_wire_2D(wire: apiWire, radius: float, chamfer: bool = False) -> apiWire:
+    """
```
ivanmaione, line 1330: Is it necessary to check that the wire is planar or is it up to the user?

tests/geometry/test_tools.py
```diff
@@ -712,3 +715,44 @@ def test_awkward_mirror(self, shape):
     def test_bad_direction(self, shape):
         with pytest.raises(GeometryError):
             mirror_shape(shape, base=(0, 0, 0), direction=(EPS, EPS, EPS))
+
+
+class TestFilletChamfer2D:
+    closed_rectangle = make_polygon({"x": [0, 2, 2, 0], "z": [0, 0, 2, 2]}, closed=True)
+    open_rectangle = make_polygon({"x": [0, 2, 2, 0], "z": [0, 0, 2, 2]}, closed=False)
+
```
ivanmaione, line 723: missing a test for negative radius

bluemira/geometry/tools.py
```diff
@@ -758,6 +758,54 @@ def sweep_shape(profiles, path, solid=True, frenet=True, label=""):
...
+    radius:
+        Radius of the fillet operation
+
+    Returns
+    -------
+    filleted_wire:
+        The filleted wire
+    """
+    if radius == 0:
+        return wire.deepcopy()
```
ivanmaione, line 778: I would have raised an error here since the operation cannot be performed. However, if there are particular applications in which this behaviour can be useful, I have nothing in particular against this solution.

bluemira/codes/_freecadapi.py
CoronelBuendia: It does actually work for 3-D wires, but I dare not expose that just yet... I will put in a defense against non-planar wires

bluemira/geometry/tools.py
CoronelBuendia: Yeah that's fair. It seems we've tried to be "forgiving" in most places where we can be. So an offset of 0 also gives a deepcopy. FreeCAD also gives you the same object in this case (but copied), so I just went with the flow

bluemira/codes/_freecadapi.py
```diff
@@ -1324,6 +1328,44 @@ def sweep_shape(profiles, path, solid=True, frenet=True):
...
+    if not _wire_is_planar(wire):
+        raise GeometryError(
+            f"Cannot {'chamfer' if chamfer else 'fillet'} a non-planar wire!"
+        )
+
+    # Temporarily suppress pesky print statement:
+    # DraftGeomUtils.fillet: Warning: edges have same direction. Did nothing
+    sys.stdout = open(os.devnull, "w")
+    result = DraftGeomUtils.filletWire(wire, radius, chamfer=chamfer)
+    sys.stdout = sys.__stdout__
```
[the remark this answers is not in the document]
je-cook: I tried suppressing it with warnings (see https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings) but its apparently not using pythons mechanics so not really.
 
I dont think it should change anything for us, we dont muck with stdout logging just uses the default, especially as you put it back immediately after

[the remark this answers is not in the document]
CoronelBuendia: Nah, it's just a print statement straight from Python in FreeCAD. OK, so all good on the logging front then!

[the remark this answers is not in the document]
hsaunders1904: I would wrap this in a try-finally block, though.

```suggestion
    # Temporarily suppress pesky print statement:
    # DraftGeomUtils.fillet: Warning: edges have same direction. Did nothing
    old_stdout = sys.stdout
    try:
        sys.stdout = open(os.devnull, "w")
        result = DraftGeomUtils.filletWire(wire, radius, chamfer=chamfer)
    finally:
        sys.stdout = old_stdout
```

bluemira/geometry/tools.py
```diff
@@ -758,6 +758,73 @@ def sweep_shape(profiles, path, solid=True, frenet=True, label=""):
...
 
 
+def fillet_chamfer_decorator(chamfer: bool):
+    """
+    Decorator for fillet and chamger operations, checking for validity of wire
+    and radius.
+    """
+
+    def decorator(func):
+        def wrapper(wire, radius):
```
je-cook, line 768: We get all the typing and return types properly and the doctrings get copied over as well if we do this

```suggestion
        @functools.wraps(func)
        def wrapper(wire, radius):
```

je-cook (approved):

ivanmaione (approved):

PR #2005, opened 2023-03-06
Move EUDEMO params and config to the new ReactorConfig object

### LINKED ISSUE #1991: Move EUDEMO's params and config to the new ReactorConfig setup

The params and config files used in EUDEMO are large and unwieldy.

Breaking them down into their respective components allows for better config management and lets users swap out different params/configs easily.

## Proposed solution

- Break down the params/configs into their respective components in the config.json files.
- Use the new ReactorConfig object to load and initialise designers/builders.
- ReactorConfig was developed as part of #1964

oliverfunk:
## Linked Issues

Closes #1991 

## Description

Moves EUDEMO reactor.py to using the new ReactorConfig object.

The ReactorConfig class was changed a bit to make it more ergonomic to use. It no longer throws if you try to access keys that don't exist and only warns if return objects are empty.

This fits in with the way things were done in EUDEMO and makes sense because those parameters/configs might be added in the future.

A path to a json file holding global_params can now optionally be given. This lets one separate global params from the build config which is how EUDEMO has been laid out.

In the future, we could isolate params per component. See Issue #2086.

Options to silence the warnings were also added.

## Interface Changes

ReactorConfig

## Checklist

### DIFF
--- a/bluemira/base/reactor_config.py
+++ b/bluemira/base/reactor_config.py
@@ -35,6 +35,25 @@ class ReactorConfig:
     config_path: str | dict
         The path to the config JSON file or a dict of the data.
 
+    global_params_type: Type
+        The ParameterFrame type for the global params.
+
+    global_params_path: Optional[str | dict]
+        A path to a JSON file
+        holding the global params or a dict of the params.
+
+    warn_on_duplicate_keys: bool
+        Print a warning when duplicate keys are found,
+        whose value will be overwritten.
+
+    warn_on_empty_local_params: bool
+        Print a warning when the local params for some args are empty,
+        when calling params_for(*args)
+
+    warn_on_empty_config: bool
+        Print a warning when the config for some args are empty,
+        when calling config_for(*args)
+
     Example
     -------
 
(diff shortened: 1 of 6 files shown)


bluemira/base/reactor_config.py
```diff
@@ -35,6 +35,25 @@ class ReactorConfig:
     config_path: str | dict
         The path to the config JSON file or a dict of the data.
 
+    global_params_type: Type
```
je-cook, line 38: sphinx auto copies the types into the docstring of the function so we only need to write the types in one place on the function argument definitions (less likely to get out of sync). Can we remove the types from the docstring here and add them all onto the init?

bluemira/base/reactor_config.py
```diff
@@ -75,18 +94,32 @@ def __init__(
         self,
         config_path: Union[str, dict],
         global_params_type: Type[_PfT],
+        global_params_path: Optional[Union[str, dict]] = None,
```
je-cook, line 97: #2086 (just for tagging)

je-cook (commented):
Just one minor otherwise looks good, it seems much more generic now which is nice

bluemira/base/reactor_config.py
oliverfunk: Ah that's great, was always annoying to copy them into the docstring

je-cook (approved):

PR #2034, opened 2023-03-13
Catch `RuntimeError` which occasionally crops up in fixed boundary equilibrium solver.

CoronelBuendia:
## Linked Issues

Closes #{ID}

For bad/difficult profiles, I've occasionally witnessed a RuntimeError, when calculating the plasma shape params. This fails hard, and the mesh point is a fair enough estimate of the extrema anyway, so we can just fall back to it softly with a warning.

## Checklist

### DIFF
--- a/bluemira/equilibria/fem_fixed_boundary/utilities.py
+++ b/bluemira/equilibria/fem_fixed_boundary/utilities.py
@@ -32,6 +32,7 @@ from matplotlib.tri import Triangulation
 from scipy.interpolate import interp1d
 
 from bluemira.base.look_and_feel import bluemira_warn
+from bluemira.utilities.error import ExternalOptError
 from bluemira.utilities.opt_problems import OptimisationConstraint, OptimisationObjective
 from bluemira.utilities.optimiser import Optimiser, approx_derivative
 from bluemira.utilities.tools import is_num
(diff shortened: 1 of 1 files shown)


bluemira/equilibria/fem_fixed_boundary/utilities.py
```diff
@@ -459,7 +460,14 @@ def find_extremum(
         )
 
         optimiser.add_eq_constraints(f_constraint, tolerance=1e-10)
-        return optimiser.optimise(x0)
+        try:
+            x_star = optimiser.optimise(x0)
+        except Exception as e:
```
je-cook, line 465: Can we have a more specific exception? This could easily hide other problems

CoronelBuendia: Done but writing a test for this will be hard, it tends to occur when things go haywire in the fixed boundary, and that only happens after a lot of stuff takes place first. Too much effort for very little reward, I feel.

je-cook: yeah I agree, I think the codecov check can be ignored for merging

ivanmaione (approved):

PR #2040, opened 2023-03-17
Make `Builder` and `Designer` consistent when handling `build_config` in `__init__`

### LINKED ISSUE #2015: Inconsistent input handling in Builder/Designer __init__
## Description of Issue

Designer and Builder initialisation inputs are inconsistent. The base Builder class does not currently accept a build_config input of None, whereas the Designer class considers build_config as optional parameter, and has a ternary operator to handle in case of None input.

## Steps to reproduce

Passing None into the build_config parameter of bluemira>base>builder.py Builder Class will result in an Attribute Error whereby it cannot perform `build_config.get("name", _remove_suffix(self.__class__.__name__, "Builder")` since 'NoneType' doesn't have the attribute `.get()`

## Expected behaviour

Consistency with the Designer class where by default

`build_config: Optional[Dict] = None`

and

`self.build_config = build_config if build_config is not None else {}`

## Proposed Solution

Make Builder and Designer classes consistent in their initialisation.

CoronelBuendia:
## Linked Issues

Closes #2015 

## Checklist

### DIFF
--- a/bluemira/base/builder.py
+++ b/bluemira/base/builder.py
@@ -26,7 +26,7 @@ Interfaces for builder classes.
 from __future__ import annotations
 
 import abc
-from typing import Dict, List, Type, Union
+from typing import Dict, List, Optional, Type, Union
 
 from bluemira.base.components import Component
 from bluemira.base.parameter_frame import ParameterFrame, make_parameter_frame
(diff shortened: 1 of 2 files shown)


bluemira/base/builder.py
```diff
@@ -103,14 +103,14 @@ class Builder(abc.ABC):
     def __init__(
         self,
         params: Union[Dict, ParameterFrame, ConfigParams, None],
-        build_config: Dict,
+        build_config: Optional[Dict] = None,
     ):
         super().__init__()
         self.name = build_config.get(
             "name", _remove_suffix(self.__class__.__name__, "Builder")
```
je-cook, line 110: This will fail as soon as `build_config = None`. Needs some catching and probably a test

PR #2041, opened 2023-03-17
Remove bad tuples in `BluemiraGeo` subclasses.

CoronelBuendia:
## Linked Issues

Closes #{ID}

Bug when getting the list of thing that is a thing..

```python
>>> from bluemira.geometry.tools import make_circle
>>> c = make_circle()
>>> c.wires
  File "/home/matti/code/bluemira/bluemira/geometry/wire.py", line 250, in wires
    return tuple(self)
TypeError: 'BluemiraWire' object is not iterable

```
## Checklist

### DIFF
--- a/bluemira/geometry/shell.py
+++ b/bluemira/geometry/shell.py
@@ -110,7 +110,7 @@ class BluemiraShell(BluemiraGeo):
         """
         The shells of the shell. By definition a list of itself.
         """
-        return tuple(self)
+        return tuple([self])
 
     @property
     def solids(self) -> tuple:
--- a/bluemira/geometry/shell.py
+++ b/bluemira/geometry/shell.py
@@ -108,13 +108,13 @@ class BluemiraShell(BluemiraGeo):
     @property
     def shells(self) -> tuple:
         """
-        The shells of the shell. By definition a list of itself.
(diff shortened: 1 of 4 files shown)


bluemira/geometry/shell.py
```diff
@@ -110,7 +110,7 @@ def shells(self) -> tuple:
         """
         The shells of the shell. By definition a list of itself.
         """
-        return tuple(self)
+        return tuple([self])
```
je-cook, line 113: Does this not also need to be done on edges wires and faces functions?

bluemira/geometry/shell.py
```diff
@@ -110,7 +110,7 @@ def shells(self) -> tuple:
         """
         The shells of the shell. By definition a list of itself.
```
je-cook, line 111: This is now not true, it is a tuple of itself not a list

bluemira/geometry/shell.py
CoronelBuendia: Not sure what you mean, for the others the inputs to the tuple are already iterables

je-cook: Oh sorry, talking rubbish read the diff the wrong way round

je-cook (approved):

PR #2042, opened 2023-03-17
WIP: Thread in various coil structure builders in EUDEMO

CoronelBuendia:
## Linked Issues

Closes #{ID}

## Checklist


bluemira/builders/coil_supports.py
```diff
@@ -323,6 +323,7 @@ def __init__(
         super().__init__(params, build_config)
         self.tf_xz_keep_out_zone = tf_xz_keep_out_zone
         self.pf_coil_xz = pf_coil_xz
+        self.name = f"{self.SUPPORT} {self.build_config.get('name', '0')}"
```
je-cook, line 326: This now changes the use of name compared other builders (especially as this is bluemira internal). What about something like this because self.SUPPORT is the same as the default self.name with spaces in it:
```python
self.name = f"{self.name} {self.build_config.get('support_number', 0)}"
```

eudemo/eudemo/coil_structure.py
```diff
@@ -0,0 +1,95 @@
...
+        params, build_config, tf_coil_xz_face, pf_coil_keep_out_zones
+    )
+    ois_xz_profiles = ois_designer.run()
+    ois_builder = OISBuilder(params, build_config, ois_xz_profiles)
+    ois_component = ois_builder.build()
+
+    tf_koz = tf_coil_xz_face.boundary[0]
+    support_components = []
+    for i, pf_coil in enumerate(pf_coil_xz_wires):
+        bc = {**build_config, "name": str(i)}
```
je-cook, line 83:
```suggestion
        bc = {**build_config, "support_number": i}
```

PR #2059, opened 2023-03-21
Add selection strategies for the points at which TF ripple is constrained

### LINKED ISSUE #1735: Expand options and flexibility of setting the points at which ripple is calculated
## Description of requirement to address

I have recently been incorporating the use of `bluemira.builders.tf_coils::RippleConstraindedLengthGOP` into the STEP-Bluemira reactor design, and encountered two pain points.

1. It currently isn't possible to explicitly set where on the LCFS the ripple should be calculated. In [`_make_ripple_points()`](https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/d447ce94d549777986a2d97c6b103c09c42144ad/bluemira/builders/tf_coils.py#L270), the Wire representing the LCFS is evenly discretized, which means that quite a large number is required if one wants to ensure that the point at which they actually want the ripple to be set as a constraint is used. Indeed, there is a comment to this effect in the code itself:
  > ```python
  > # Real argument to making the points the inputs... but then the plot would look
  > # sad! :D
  > # Can speed this up a lot if you know about your problem... I.e. with a princeton
  > # D I could only check one point and get it right faster.
  > ```

2. The discretization fails if `n_rip_points <= 3`. I don't have the error message to hand and we have now put our own warning in our codes, so I would need to check out an earlier version to find it. Please let me know if the error itself would be helpful.

## Proposed solution

For point 1,  we do have some routines for setting ripple points with BLUEPRINT for STEP that could probably be fairly easily ported over Bluemira. I am happy to contribute a solution if the core developers agree it would be valuable. Unfortunately, I am on AL for three weeks from now (wanted to get this issue down before I forget).

For point 2, it would be fairly simple to add a warning in the method about this restriction on `n_rip_points`, although any modifications done with point 1 would also need to be considered (i.e. it might be possible to use only 1 ripple point if setting it manually, but it isn't possible if discretizing).

## Alternative solutions

N/A

## Additional context

N/A

CoronelBuendia:
## Linked Issues

Requires #2065 
Closes #1735 

OK, I've had a quick bash at this, but before I go much further, I'd like to just summarise what I'm going for, and check it will do what you want, @bielsnohr.

I've added an argument to the `RippleConstrainedLengthGOP` `__init__` in which a `RipplePointSelector` can be provided. This is not the nicest API, and I'm sure there are nicer ways to instantiate these things which I will improve on later.

I kind of need a class or protocol to capture the following options:

* Equi-spaced points at which the ripple shall be calculated
* Equi-spaced points above a certain minimum radius (could be combined with the above)
* Direct specification of the points at which ripple should be calculated (your use case?)
* Finding the point of maximum ripple along a wire

Please run:
```bash
pytest tests/builders/test_tf_coils.py --plotting-on
```

and have a look at the API of the aforementioned strategies to see if they are going to deliver what you need.

## Checklist

### DIFF
--- a/bluemira/builders/tf_coils.py
+++ b/bluemira/builders/tf_coils.py
@@ -161,7 +161,7 @@ class ParameterisedRippleSolver:
         return self.cage.ripple(x, y, z)
 
 
-class RipplePointConstraintStrategy(ABC):
+class RipplePointSelector(ABC):
     @abstractmethod
     def make_ripple_points(self) -> Coordinates:
         pass
@@ -171,7 +171,7 @@ class RipplePointConstraintStrategy(ABC):
         pass
 
 
-class EquispacedRipplePointConstraintStrategy(RipplePointConstraintStrategy):
+class EquispacedSelector(RipplePointSelector):
     def __init__(self, wire: BluemiraWire, n_rip_points: int):
         points = wire.discretize(byedges=True, ndiscr=n_rip_points)
         self.points = points
@@ -181,6 +181,31 @@ class EquispacedRipplePointConstraintStrategy(RipplePointConstraintStrategy):
 
     def make_calc_ripple_func(self) -> callable:
         return ParameterisedRippleSolver.ripple
+
+
+class OutboardEquispacedSelector(RipplePointSelector):
+    def __init__(self, wire: BluemiraWire, n_rip_points: int):
+        points = wire.discretize(byedges=True, ndiscr=n_rip_points)
+        self.points = points
+
+    def make_ripple_points(self) -> Coordinates:
+        return self.points
+
+    def make_calc_ripple_func(self) -> callable:
+        return ParameterisedRippleSolver.ripple
+
+
+class FixedSelector(RipplePointSelector):
+    def __init__(self, points: Coordinates):
+        self.points = points
+
+    def make_ripple_points(self) -> Coordinates:
+        return self.points
+
+
+class MaximiseSelector(RipplePointSelector):
+    def __init__(self, wire: BluemiraWire):
+        self.wire = wire
 
 
 class RippleConstrainedLengthGOP(GeometryOptimisationProblem):
@@ -244,6 +269,7 @@ class RippleConstrainedLengthGOP(GeometryOptimisationProblem):
         ny=1,
         n_rip_points=100,
         n_koz_points=100,
+        ripple_selector=None,
     ):
         self.params = make_parameter_frame(params, self._Params)
         self.separatrix = separatrix
@@ -254,7 +280,10 @@ class RippleConstrainedLengthGOP(GeometryOptimisationProblem):
             minimise_length, f_objective_args={"parameterisation": parameterisation}
         )
 
-        self.ripple_points = self._make_ripple_points(separatrix, n_rip_points)
+        if ripple_selector is None:
+            ripple_selector = EquispacedSelector(separatrix, n_rip_points)
+
+        self.ripple_points = ripple_selector.make_ripple_points(separatrix, n_rip_points)
         self.ripple_values = None
         self.solver = ParameterisedRippleSolver(
             wp_cross_section,
--- a/bluemira/builders/tf_coils.py
+++ b/bluemira/builders/tf_coils.py
@@ -26,13 +26,15 @@ Built-in build steps for making parameterised TF coils.
 from abc import ABC, abstractmethod
 from copy import deepcopy
 from dataclasses import dataclass
+from typing import Optional
 
 import matplotlib
 import matplotlib.pyplot as plt
 import numpy as np
 
+from bluemira.base.constants import EPS
 from bluemira.base.error import BuilderError
-from bluemira.base.look_and_feel import bluemira_debug_flush
+from bluemira.base.look_and_feel import bluemira_debug_flush, bluemira_warn
 from bluemira.base.parameter_frame import Parameter, ParameterFrame, make_parameter_frame
 from bluemira.display import plot_2d
 from bluemira.geometry.coordinates import Coordinates
@@ -42,7 +44,7 @@ from bluemira.geometry.optimisation import (
     constrain_koz,
     minimise_length,
 )
-from bluemira.geometry.tools import offset_wire
+from bluemira.geometry.tools import boolean_cut, make_polygon, offset_wire
 from bluemira.geometry.wire import BluemiraWire
 from bluemira.magnetostatics.biot_savart import BiotSavartFilament
 from bluemira.magnetostatics.circuits import HelmholtzCage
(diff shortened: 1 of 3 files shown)


bielsnohr:
Thanks for this @CoronelBuendia ! I've been a bit swamped, but will take a look tomorrow afternoon (Friday, 24 March) if that's all right.

bielsnohr:
After diving down quite the unpleasant rabbit hole in our STEP code base, I have finally found what I believe is the way that ripple points were _hoped_ to be selected with the "old" BLUEPRINT style builders. The desired interface (which I am uncertain if it was ever operational for ripple constrained optimisation) was to give z values at which the LCFS would be sliced to get intersection points that would then be used as the ripple points. There was also an option to give a _fractional_ height of the LCFS at which this slicing should be done (e.g. 0.5 could be passed in and the z value that is half as tall as the LCFS would be used)

From what I understand, these intersections could be passed directly to the `FixedSelector` that you have implemented, which should satisfy our use case. Alternatively, if we want a `Selector` that uses z heights or fractional z heights, we could implement something like `ZValueSelector` and then pass that into the `RippleConstrainedLengthGOP`. Now that there is an option to set the point selection strategy this within the optimiser class, we have the freedom to do whatever we want, meaning our use case will most certainly be covered no matter what it is.

CoronelBuendia:
> After diving down quite the unpleasant rabbit hole in our STEP code base, I have finally found what I believe is the way that ripple points were _hoped_ to be selected with the "old" BLUEPRINT style builders. The desired interface (which I am uncertain if it was ever operational for ripple constrained optimisation) was to give z values at which the LCFS would be sliced to get intersection points that would then be used as the ripple points. There was also an option to give a _fractional_ height of the LCFS at which this slicing should be done (e.g. 0.5 could be passed in and the z value that is half as tall as the LCFS would be used)
> 
> From what I understand, these intersections could be passed directly to the `FixedSelector` that you have implemented, which should satisfy our use case. Alternatively, if we want a `Selector` that uses z heights or fractional z heights, we could implement something like `ZValueSelector` and then pass that into the `RippleConstrainedLengthGOP`. Now that there is an option to set the point selection strategy within the optimiser class, we have the freedom to do whatever we want, meaning our use case will most certainly be covered no matter what it is.

OK, great, thanks for that. I will have a think about combing the `EquispacedSelector` and possibly throw in something along the lines that you want just cos while I'm at it, why not. I was thinking of putting in an `ExtremaSelector` anyway, which would catch all extremal points, which would probably be best for a ST with a `PictureFrame`, given that peak ripple will likely be at one of the extrema.

But even if I don't, as you say, it should be possible for you to define your own `Selector`.

tests/builders/test_tf_coils.py
```diff
@@ -0,0 +1,122 @@
...
+            "z_0": {"value": 0, "unit": "m", "source": "test"},
+            "B_0": {"value": 6, "unit": "T", "source": "test"},
+            "TF_ripple_limit": {"value": 0.6, "unit": "%", "source": "test"},
+        },
+        RippleConstrainedLengthGOPParams,
+    )
+
+    @classmethod
+    def teardown_method(cls):
+        plt.show()
```
je-cook, line 66: Otherwise we get matplotlib warnings on the CI eventually
```suggestion
        plt.show()
        plt.close()
```

bluemira/builders/tf_coils.py
```diff
@@ -72,15 +76,14 @@ class ParameterisedRippleSolver:
         Points at which to calculate TF ripple
```
je-cook, line 76: ripple points has been removed from the init but not the docstring

bluemira/builders/tf_coils.py
```diff
@@ -140,17 +143,397 @@ def _make_single_circuit(self, wire):
...
+                    {
+                        "x": [0, x_min, x_min, 0],
+                        "y": 0,
+                        "z": [z_min, z_min, z_max, z_max],
+                    },
+                    closed=True,
+                )
+            )
+            wire = boolean_cut(wire, cut_face)[0]
+        self.points = wire.discretize(byedges=True, ndiscr=self.n_rip_points)
```
je-cook, line 325: Is `self.points` not consistent with `self.wire` as you possibly modify the original wire to make the points?

bluemira/builders/tf_coils.py
```diff
@@ -140,17 +143,397 @@ def _make_single_circuit(self, wire):
...
+        arg_x_max = np.argmax(coords.x)
+        arg_z_min = np.argmin(coords.z)
+        arg_z_max = np.argmax(coords.z)
+        self.points = Coordinates(
+            [
+                coords.points[arg_x_min],
+                coords.points[arg_x_max],
+                coords.points[arg_z_min],
+                coords.points[arg_z_max],
+            ]
```
je-cook, line 357:
```suggestion
        self.points = Coordinates(
            [
                coords.points[np.argmin(coords.x)],
                coords.points[np.argmax(coords.x)],
                coords.points[np.argmin(coords.z)],
                coords.points[np.argmax(coords.z)],
            ]
```

bluemira/builders/tf_coils.py
```diff
@@ -140,17 +143,397 @@ def _make_single_circuit(self, wire):
...
+        Parameters
+        ----------
+        wire
+            Wire along which the points will be selected
+        """
+        super().set_wire(wire)
+        points = wire.discretize(byedges=True, ndiscr=200)
+        arg_x_max = np.argmax(points.x)
+        x_max_point = points[:, arg_x_max]
+        self._alpha_0 = wire.parameter_at(x_max_point, tolerance=10 * EPS)
```
je-cook, line 403: probably the same question about inconsistency with `self.wire`, should we use `self.wire` instead of `wire`?

bluemira/builders/tf_coils.py
```diff
@@ -140,17 +143,397 @@ def _make_single_circuit(self, wire):
...
+        vector: np.ndarray
+            Variable vector
+        grad: np.ndarray
+            Jacobian matrix of the constraint (updated in place)
+        parameterisation: GeometryParameterisation
+            Geometry parameterisation
+        solver: ParameterisedHelmholtzSolver
+            TF ripple solver
+        TF_ripple_limit: float
+            Maximum allowable TF ripple
```
je-cook, line 454: Can we finish off the docstring, some of the arguments aren't documented, alternatively this method could be made private (`_constrain...`) as it feels like it wont be used outside of the class.

bluemira/builders/tf_coils.py
```diff
@@ -140,17 +143,397 @@ def _make_single_circuit(self, wire):
...
+        Parameters
+        ----------
+        vector: np.ndarray
+            Variable vector
+        parameterisation: GeometryParameterisation
+            Geometry parameterisation
+        solver: ParameterisedHelmholtzSolver
+            TF ripple solver
+        TF_ripple_limit: float
+            Maximum allowable TF ripple
```
je-cook, line 501: Same with this docstring

je-cook (changes requested):
Hopefully whenever we get back to #1894 the this self thing will vanish...

bluemira/builders/tf_coils.py
CoronelBuendia: `self.wire` should be set with `set_wire`. The wire is not modified by any of the classes, just used as a basis from which to get the points. I will make it a private attribute, to discourage direct attribution.

je-cook (approved):

PR #2060, opened 2023-03-21
EUDEMO first wall panelling

### LINKED ISSUE #1899: First Wall Panelling
## Description

Implement wall panelling for the `eudemo` reactor build.

The designer should solve the problem of minimising the number of tiles. It should reimplement the `Paneller` functionality that was in BLUEPRINT.

## Outputs

- Implement `PanelDesigner`
- Implement `PanelBuilder`
- Panelling added to EUDEMO IVC build

hsaunders1904:
## Linked Issues

Closes #1899 

## Description

Implements a first wall panelling designer for EUDEMO. Given some open boundary, the designer wraps that boundary in a polygon, attempting to minimise the length of the polygon (hence the length of the panels), whilst constraining the minimum length of a single panel, and the maximum angle between two panels.

I'm not sure that I'm very happy with the stability of the thing. It's quite temperamental, and small changes in the problem can lead to constraint violations. Even in the best case, constraints are often violated, but only by a little.

Do run the tests with plotting and stdout on, and you can see in how many of the cases constraints are violated. 

```console
export PYTHONPATH=$(pwd)/eudemo:$PYTHONPATH
pytest eudemo/eudemo_tests/ivc/panelling -s -v --plotting-on
```

There also seems to be a case where constraints are violated quite a lot, but my test doesn't fail - so I think there may be a bug in there somewhere. So opening this as a draft for now.

It may help with stability if we can optimise on number of panels as well, but maybe we should leave that for another PR.

## Interface Changes

Nope

## Checklist


eudemo/eudemo/ivc/panelling/_paneller.py
```diff
@@ -0,0 +1,252 @@
...
+    ----------
+    points
+        A numpy array of points, shape should be (2, N).
+    """
+    dists = np.diff(points, axis=1)
+    sq_dists = np.square(dists, out=dists)
+    summed_dists = np.sum(sq_dists, axis=0)
+    sqrt_dists = np.sqrt(summed_dists, out=summed_dists)
+    cumulative_sum = np.cumsum(summed_dists, out=sqrt_dists)
+    return np.hstack((0, cumulative_sum / cumulative_sum[-1]))
```
je-cook, line 177: What was the thoughts on using the `out` argument? 
It looks like we are performing operations and then renaming the variable without reassigning memory so after running L176 `summed_dists == sqrt_dists == cumlative_sum` . Unless I'm mistaken?

It seems a little confusing, but I assume you're trying to flag the logic flow?

eudemo/eudemo/ivc/panelling/_paneller.py
```diff
@@ -0,0 +1,252 @@
...
+
+    Returns
+    -------
+    a: np.array(2, N)
+        The array of 2-D normal vectors of each side of a polygon
+    """
+    a = -np.array([-side_vectors[1], side_vectors[0]]) / np.sqrt(
+        side_vectors[0] ** 2 + side_vectors[1] ** 2
+    )
+    a[np.isnan(a)] = 0
```
je-cook, line 251: Have you seen any divide by NaN warnings come up here? It seems possible at any rate

eudemo/eudemo/ivc/panelling/_pivot_string.py
```diff
@@ -0,0 +1,103 @@
...
+# Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public
+# License along with bluemira; if not, see <https://www.gnu.org/licenses/>.
+from itertools import count
+from typing import Tuple
+
+import numpy as np
+
+DEG_TO_RAD = np.pi / 180
```
je-cook, line 26: maybe? up to you
```suggestion
DEG_TO_RAD = np.deg2rad(1)
```

je-cook:
Looks like I missed a warning on #2059 oops!

CoronelBuendia (changes requested):
Looking good, a few minor comments. Probably the major comment: I think your tests should include some "half-arches" as in EUDEMO at least that is what we're likely to be doing (i.e. for both inboard and outboard blanket segments).

bluemira/utilities/optimiser.py
```diff
@@ -209,7 +209,7 @@ def add_ineq_constraints(
             tolerance = np.array([tolerance])
         super().add_ineq_constraints(f_constraint, tolerance)
 
-    def check_constraints(self, x):
+    def check_constraints(self, x: np.ndarray, warn: bool = False):
```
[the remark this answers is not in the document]
hsaunders1904: You're right, this should definitely be `True`

eudemo/eudemo/ivc/panelling/_paneller.py
hsaunders1904: It's potentially premature optimisation, but it's for performance. Using `out` means we're not constantly re-allocating memory for new arrays; we're re-using memory we've already allocated.

I admit, when I first wrote this I thought it was going to be used within an optimisation loop, but it's not. Maybe I should remove the `out` arguments for clarity's sake, as I don't think performance matters much here?

eudemo/eudemo/ivc/panelling/_designer.py
```diff
@@ -0,0 +1,203 @@
...
+            # Note we're actually increasing the number of panels by 1
+            # by adding 3 below, as there are two more panels than
+            # optimisation parameters.
+            n_panels = opt_problem.n_opts + 3
+            opt_problem = self._set_up_opt_problem(boundary, n_panels)
+            try:
+                x_opt = opt_problem.optimise(check_constraints=False)
+            except ExternalOptError:
+                x_opt = None
+            iter_num += 1
```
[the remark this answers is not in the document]
hsaunders1904: I did explore that a little. The optimum will always end up being the one with most panels though, as that solution will inevitably hug the boundary more, and have the smallest length. So we may as well just run the optimisation on `2*len(initial_joints)` to begin with. 

The way it's currently set up (which is in the spirit of how the problem appears to be approached in BLUEPRINT), we're getting the minimum number of panels for which the constraints are satisfied. Doing the mixed optimisation problem you're describing (with no upper bound) would result in the maximum number of panels for which the constraints are satisfied (as this would inevitably be the minimum length). Is that what we actually want?

eudemo/eudemo/ivc/panelling/_paneller.py
je-cook: Possibly premature op but I don't mind. I have been bitten before using the same `out` array as the `in` array, you can get some weird bugs depending on the operations that are happening.

eudemo/eudemo/ivc/panelling/_paneller.py
hsaunders1904: Do you mean divide by 0 warnings from calculating `a`? I haven't come across any in my testing. I'll admit this is ripped straight from BLUEPRINT, so I'm fuzzy on the justification for this line

eudemo/eudemo/ivc/panelling/_paneller.py
hsaunders1904: This function has been removed, as there is a similar one in `geometry.coordinates`

eudemo/eudemo/ivc/panelling/_opt_problem.py
```diff
@@ -0,0 +1,150 @@
...
+        these parameters are the normalised length along the paneller's
+        boundary, where the panels and boundary meet.
+    optimiser
+        The :class:`.Optimiser` to perform the optimisation with.
+    """
+
+    def __init__(self, paneller: Paneller, optimiser: Optimiser):
+        self.paneller = paneller
+        self.bounds = (np.zeros_like(self.paneller.x0), np.ones_like(self.paneller.x0))
+        objective = OptimisationObjective(self.objective)
```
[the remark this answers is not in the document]
hsaunders1904: 🤔 The `self` argument is implicit, so it shouldn't need to be static

eudemo/eudemo/ivc/panelling/_paneller.py
je-cook: I did, sure, if you havent seen any that fine

hsaunders1904: This is another one that was already in the `geometry` module https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/2060/files#diff-fddce791a384b13641d79f8580424a09ee09d2f3def690f575c9ba95d67fb44eL101

eudemo/eudemo/ivc/panelling/_designer.py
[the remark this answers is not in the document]
CoronelBuendia: No, you are right. Generally we want few panels. But there's probably a pareto front with a second objective there we are ignoring, which is fine.

eudemo/eudemo/ivc/panelling/_opt_problem.py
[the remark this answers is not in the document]
CoronelBuendia: Sure, normally, but when we wrap functions to go into the optimiser, I thought the first args need to be as the signatures expect, and if they have a self that is taken as the first arg. Are you saying we could get rid of the staticmethods e.g. in https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/2059 `tf_coils.py`?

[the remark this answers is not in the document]
hsaunders1904: As far as I can tell, I don't think those static functions are necessary, no. We should be able to reference `self` in the objective. I would guess all `nlopt` is doing is calling the function the same way we would; `instance.objective(x, grad)`.

hsaunders1904:
I've added some tests that panel just the inboard/outboard.

<details>
  <summary>Plots</summary>

image

image

</details>


@je-cook 
> Locally I get one test failure for `eudemo/eudemo_tests/ivc/panelling/test_designer.py::TestPanellingDesigner::test_panels_fully_enclose_wall_boundary[run]`:

Could you send me a plot and/or the array for the error case?

CoronelBuendia (approved):

je-cook:
End result is the discretisation of the boundary was too coarse for the test (by 1 point 🤯). Making that configurable will bypass the issue.

PR #2067, opened 2023-03-24
More EUDEMO build work

CoronelBuendia:
## Linked Issues

Closes #{ID}

## Checklist


eudemo/config/build_config.json
```diff
@@ -123,7 +123,7 @@
         }
     },
     "PF coils": {
-        "run_mode": "read",
+        "run_mode": "run",
```
je-cook, line 126: If we're changing this to run can we remove the json files or is it still useful to have them stored?

je-cook (commented):
Just the one thing about PF and TF json now its much quicker.

I've also noticed that because we don't set any colours for cad on the xz or probably xy, only for plot colours therefore the colours will all be off for that even after #1033 fixes the display options. I'm sure there is a clever way to improve that.

eudemo/config/build_config.json
CoronelBuendia: I think it is still useful, but yeah it is annoying with all the diffs. Let me try and get to the point where all the saved files are consistent with each other, and then we can gitignore them, so that one can always use "read" and still get something that makes sense.

je-cook (approved):

PR #2087, opened 2023-03-28
Module reshuffle for 1.5-D and API reshuffle for 2-D fixed boundary equilibrium

CoronelBuendia:
## Linked Issues

Closes #{ID}

## Checklist


bluemira/magnetostatics/finite_element_2d.py
```diff
@@ -0,0 +1,326 @@
...
+from typing import Callable, Iterable, List, Optional, Tuple, Union
+
+import dolfin
+import numpy as np
+
+from bluemira.base.constants import MU_0
+
+
+def Bz_coil_axis(x: float, z: float, pz: float, current: float) -> float:
+    """
```
CoronelBuendia, line 35: @ivanmaione  please check this docstring, I dunno if I'm tired or what but I found the signature confusing

examples/equilibria/fixed_boundary.ex.py
```diff
@@ -0,0 +1,137 @@
...
+    I_p=I_p, R_0=R_0, B_0=B_0, p_order=2, max_iter=30, iter_err_max=1e-4, relaxation=0.05
+)
+solver.set_mesh(mesh)
+solver.set_profiles(p_prime, ff_prime)
+equilibrium = solver.solve(plot=True)
+
+
+# %% [markdown]
+# Let's inspect the result
+
```
CoronelBuendia, line 126: This is less trivial than I would like... will till the day I die be a fan of `something.plot()` :)

bluemira/equilibria/fem_fixed_boundary/fem_magnetostatic_2D.py
```diff
@@ -298,7 +110,7 @@ def __init__(
         I_p: Optional[float] = None,
         R_0: Optional[float] = None,
         B_0: Optional[float] = None,
-        p_order: int = 3,
+        p_order: int = 2,
```
CoronelBuendia, line 113: @ivanmaione There's a real argument here to removing all the inputs from p_prime to B_0 from the __init__, seeing as we presently have to use `.set_profiles` after the mesh is set, and to enable different ordering of these calls could get messy.

Like, right now, if you set p_prime and ff_prime in the init, but don't use set_profiles before solve, you get an error.

je-cook:
Is now a good opportunity to fill this in instead of having TBD or at least open an issue to do it? https://bluemira.readthedocs.io/en/latest/magnetostatics/magnetostatics.html#finite-element

CoronelBuendia:
> Is now a good opportunity to fill this in instead of having TBD or at least open an issue to do it? https://bluemira.readthedocs.io/en/latest/magnetostatics/magnetostatics.html#finite-element

Yeah we should add that in a separate PR onto this feature branch. Not sure we have all that much to say at the moment, but @ivanmaione can discuss some minimal documentation.

bluemira/magnetostatics/finite_element_2d.py
ivanmaione: Pretty sure you were tired... and also that the docstring and signature were confusing. Directly modified.

ivanmaione (approved):

PR #2104, opened 2023-04-03
🧑‍💻 Time all design and build stages in debug output

je-cook:
## Description

Adds timing shown only in debugging logging for all designers and builders

## Checklist

### DIFF
--- a/bluemira/base/builder.py
+++ b/bluemira/base/builder.py
@@ -88,12 +88,12 @@ class Builder(abc.ABC):
 
     Parameters
     ----------
-    params: Union[Dict, ParameterFrame]
+    params
         The parameters required by the builder.
-    build_config: Dict
+    build_config
         The build configuration for the builder.
-    designer: Optional[Designer]
-        A designer to solve a design problem required by the builder.
+    verbose
+        control how much logging the designer will output
 
     Notes
     -----
--- a/bluemira/base/builder.py
+++ b/bluemira/base/builder.py
@@ -26,13 +26,12 @@ Interfaces for builder classes.
 from __future__ import annotations
 
 import abc
-import time
 from typing import Dict, List, Optional, Type, Union
 
 from bluemira.base.components import Component
-from bluemira.base.look_and_feel import bluemira_debug, bluemira_print
 from bluemira.base.parameter_frame import ParameterFrame, make_parameter_frame
 from bluemira.base.reactor_config import ConfigParams
+from bluemira.base.tools import timing
 from bluemira.utilities.plot_tools import set_component_view
 
 BuildConfig = Dict[str, Union[int, float, str, "BuildConfig"]]
@@ -106,6 +105,8 @@ class Builder(abc.ABC):
         self,
         params: Union[Dict, ParameterFrame, ConfigParams, None],
         build_config: Optional[Dict] = None,
+        *,
+        verbose=True,
     ):
         super().__init__()
         self.params = make_parameter_frame(params, self.param_cls)
@@ -113,16 +114,9 @@ class Builder(abc.ABC):
         self.name = self.build_config.get(
             "name", _remove_suffix(self.__class__.__name__, "Builder")
         )
-        self._build = self.build
-        self.build = self._build_wrapper
-
-    def _build_wrapper(self) -> Component:
-        """Build the component."""
-        bluemira_print(f"Building {self.name}")
-        t1 = time.perf_counter()
-        out = self._build()
-        bluemira_debug(f"Built in {time.perf_counter() - t1:.2g} s")
-        return out
+        self.build = timing(
+            self.build, "Built in", f"Building {self.name}", print_name=verbose
+        )
 
     @abc.abstractproperty
     def param_cls(self) -> Union[Type[ParameterFrame], None]:
--- a/bluemira/base/builder.py
+++ b/bluemira/base/builder.py
@@ -119,9 +119,9 @@ class Builder(abc.ABC):
     def _build_wrapper(self) -> Component:
         """Build the component."""
         bluemira_print(f"Building {self.name}")
-        t1 = time.time()
+        t1 = time.perf_counter()
         out = self._build()
-        bluemira_debug(f"Built in {time.time() - t1} s")
+        bluemira_debug(f"Built in {time.perf_counter() - t1:.2g} s")
         return out
 
     @abc.abstractproperty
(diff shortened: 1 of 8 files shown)


eudemo/eudemo/reactor.py
```diff
@@ -225,8 +226,7 @@ def _read_json(file_path: str) -> Dict:
 
 
 if __name__ == "__main__":
-    import time
-
+    set_log_level("DEBUG")
```
oliverfunk, line 229: Personally, I think debug is too verbose as a default, even for development purposes.
I'm not normally that interested in the execution timings/all the other stuff that comes with debug (like the messages about the dyn executed modules).

oliverfunk (approved):
LGTM, great improvement

eudemo/eudemo/reactor.py
je-cook: sure I'll leave it there so it can easily be changed but set it to info

bluemira/base/builder.py
```diff
@@ -111,6 +113,16 @@ def __init__(
         self.name = self.build_config.get(
             "name", _remove_suffix(self.__class__.__name__, "Builder")
         )
+        self._build = self.build
+        self.build = self._build_wrapper
+
+    def _build_wrapper(self) -> Component:
```
CoronelBuendia, line 119: I guess decorators have to be imported and applied to subclasses, is that why this is like this?

bluemira/base/builder.py
```diff
@@ -111,6 +113,16 @@ def __init__(
...
+        """Build the component."""
+        bluemira_print(f"Building {self.name}")
+        t1 = time.time()
+        out = self._build()
+        bluemira_debug(f"Built in {time.time() - t1} s")
```
CoronelBuendia, line 124: Could we have some pretty time formatting? i.e. `:.2f` or some `datetime.now().strpf()` or whatever?

bluemira/base/builder.py
je-cook: exactly yeah, I couldn't think of a way to do this implicitly without giving the reactor designer something else to do.

CoronelBuendia:
Sorry still a couple of nits to pick for me!
This time formatting is better, but still difficult to understand for anything over 60-100 seconds.

1.1e2 seconds for example is less intuitive than 1 m 50 seconds.

Also, the granularity works well for large builders, but now we're getting a lot of noise with anything that uses lots of little builders, e.g. the PF coils (with DEBUG logging):

image

Perhaps worth considering when and where this decorated approach should be applied, because right now it would get very noisy, even without DEBUG logging.

je-cook:
> Sorry still a couple of nits to pick for me! This time formatting is better, but still difficult to understand for anything over 60-100 seconds.
> 
> 1.1e2 seconds for example is less intuitive than 1 m 50 seconds.

Ok will have another go

> Also, the granularity works well for large builders, but now we're getting a lot of noise with anything that uses lots of little builders, e.g. the PF coils (with DEBUG logging):
> > Perhaps worth considering when and where this decorated approach should be applied, because right now it would get very noisy, even without DEBUG logging.

Debug should be noisy as thats the point but I can understand little designers (as here) or builders not necessarily wanting a message saying we're running. I will have a think

>

oliverfunk:
@je-cook wen merge? Would like this feature

je-cook:
I need to make it a bit more robust first, should be done for tomorrow

oliverfunk (approved):
I think the current level of output at INFO is good. DEBUG is noisy but that's also the point.

bluemira/base/designer.py
```diff
@@ -55,11 +58,20 @@ def __init__(
...
         params: Union[Dict, ParameterFrame, ConfigParams, None],
         build_config: Optional[Dict] = None,
+        *,
+        verbose=True,
     ):
         self.params = make_parameter_frame(params, self.param_cls)
         self.build_config = build_config if build_config is not None else {}
         self._run_func = self._get_run_func(self.run_mode)
 
+        self.execute = timing(
```
hsaunders1904, line 68: I think we're going to lose the typing on `execute` doing this (and in `Builder.build`). As `timing` returns a vanilla `Callable`, type checkers will now think `execute` just returns `Any`.

The only thing I can think of is to use a `TypeVar` in the `timing` signature. Something like this.

```python
_T = TypeVar("_T")
def timing(func: Callable[..., _T], ...) -> Callable[..., _T]:
```

This at least gets us the return type. I can't work out how to get it to recognise the argument types, though.

There may be alternative solutions, though. Would a decorator work? Or did you try that?

bluemira/base/tools.py
```diff
@@ -22,11 +22,36 @@
...
 from bluemira.base.components import Component, PhysicalComponent
+from bluemira.base.look_and_feel import bluemira_debug, bluemira_print
 from bluemira.geometry.compound import BluemiraCompound
 from bluemira.geometry.tools import serialize_shape
 
 
+def timing(
+    func: Callable, timing_str: str, print_str: str = "", print_name: bool = True
+) -> Callable:
+    """Time a function and push to logging"""
```
hsaunders1904, line 37: This is part of our public API (should it be?), so I think the arguments could do with being documented. It's not immediately obvious to me what they're doing from their names.

hsaunders1904 (approved):
This is a good feature to have. Just a couple of comments on documentation and typing

bluemira/base/designer.py
je-cook: It could be used as a decorator here but we will lose all the customisation.  I've added `wraps` which should help and changed the location of the wrapping for the designer. That should take care of the typing for the designer. For the builder I dont really know the best way as the user writes the `build` function

bluemira/base/tools.py
je-cook: I've documented and remove it from the public api also renamed the variables that hopefully make more sense

bluemira/base/designer.py
```diff
@@ -55,18 +58,25 @@ def __init__(
...
 
     def execute(self) -> _DesignerReturnT:
         """
         Execute the designer with the run mode specified by the build config.
 
         By default the run mode is 'run'.
         """
-        return self._run_func()
+        return _timing(
+            self._get_run_func(self.run_mode),
```
je-cook, line 75: This change means that the run function is not defined until execute is called, but is does mean that the run function can get dynamic which is wasnt before, thoughts?

hsaunders1904: This seems sensible to me. It means you can update the run mode in the build config and the change is honoured when you call `execute`.

PR #2108, opened 2023-04-03
clean up code

oliverfunk:
## Linked Issues

Closes #{ID}

## Checklist


oliverfunk:
@CoronelBuendia could you perhaps take a look and see if I'm on the right track.

This PR is mostly an RFC.
James and I had a chat about what's required to build the face that will be extruded radially outwards to build the XYZ shape

eudemo/eudemo/maintenance/lower_port/designer.py
```diff
@@ -0,0 +1,595 @@
...
+
+        div_x_inner = self.divertor_face.bounding_box.x_min
+        div_x_outer = self.divertor_face.bounding_box.x_max
+
+        div_diag_len, _ = distance_to(
+            [div_x_outer, 0, div_z_top],
+            [div_x_inner, 0, div_z_bot],
+        )
+
+        points = LowerPortDesigner._xz_points_dist_away_from(
```
je-cook, line 187:
```suggestion
        points = self_xz_points_dist_away_from(
```

eudemo/eudemo/maintenance/lower_port/designer.py
```diff
@@ -0,0 +1,595 @@
...
+            (div_x_inner, div_z_bot),
+            self._angled_duct_gradient,
+            div_diag_len,
+        )
+        wire = LowerPortDesigner._make_xz_wire_from_points(points[0], points[1])
```
je-cook, line 192:
```suggestion
        wire = self._make_xz_wire_from_points(points[0], points[1])
```

eudemo/eudemo/maintenance/lower_port/designer.py
```diff
@@ -0,0 +1,595 @@
...
+
+        c_pts = LowerPortDesigner._closest_points(wire, self.divertor_face)
```
je-cook, line 194:
```suggestion
        c_pts = self._closest_points(wire, self.divertor_face)
```

eudemo/eudemo/maintenance/lower_port/designer.py
```diff
@@ -0,0 +1,595 @@
...
+        return make_polygon(
+            {
+                "x": np.array(
+                    [
+                        x_point,
+                        x_point,
+                        x_point,
+                        x_point,
+                    ]
+                ),
```
je-cook, line 278:
```suggestion
                "x": x_point,
```

eudemo/eudemo/maintenance/lower_port/designer.py
```diff
@@ -0,0 +1,595 @@
...
+                        y_size,
+                    ]
+                ),
+                "z": np.array(
+                    [
+                        start_top_point[1],
+                        start_top_point[1],
+                        start_bot_point[1],
+                        start_bot_point[1],
+                    ]
```
je-cook, line 293: for readability it would maybe be better to create these arrays before this function

eudemo/eudemo/maintenance/lower_port/designer.py
```diff
@@ -0,0 +1,595 @@
...
+            z_top_point=straight_duct_z_top,
+        )
+        straight_duct = BluemiraFace(straight_duct_boundary)
+
+        angled_cuts = boolean_cut(
+            angled_duct,
+            [straight_duct],
+        )
+        top_piece = angled_cuts[0]
+        duct_inner = boolean_fuse([top_piece, straight_duct])
```
je-cook, line 404: Possibly overkill but a bit of rearrangment would be good as angled pieces is not really used
```suggestion
        duct_inner = boolean_fuse([boolean_cut(angled_duct, straight_duct)[0], straight_duct])
```

je-cook (commented):
First run through still getting my head through it but a bit of cleanup to do

eudemo/eudemo/maintenance/lower_port/designer.py
```diff
@@ -0,0 +1,595 @@
...
+            },
+        )
+
+    @staticmethod
+    def _intersection_points(
+        shape_a: BluemiraGeo,
+        shape_b: BluemiraGeo,
+    ) -> List[Tuple]:
+        dist, vects = distance_to(shape_a, shape_b)
+        if dist > 0.0001:  # not intersecting
```
je-cook, line 575: the geometry constant D_TOLERANCE should probably be used here https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/f1518a172c7a15348d7854635e4667d4c37ecae1/bluemira/geometry/constants.py#L27

eudemo/eudemo/maintenance/lower_port/designer.py
```diff
@@ -0,0 +1,595 @@
...
+            pois.append((v[0], v[1], v[2]))
+        return pois
+
+    @staticmethod
+    def _closest_points(
+        shape_a: BluemiraGeo,
+        shape_b: BluemiraGeo,
+    ) -> List[Tuple]:
+        dist, vects = distance_to(shape_a, shape_b)
+        if dist < 0.0001:  # intersecting, return intersection points
```
je-cook, line 589: Same here

je-cook: also you are essentially doing this check twice, once here and the reverse in the function you call in the return. Is it possible to adjust to its only checked once?

eudemo/eudemo/reactor.py
```diff
@@ -219,14 +249,8 @@ def build_radiation_shield(params, build_config, cryostat_koz) -> RadiationShiel
...
 
-def _read_json(file_path: str) -> Dict:
-    """Read a JSON file to a dictionary."""
-    with open(file_path, "r") as f:
-        return json.load(f)
-
-
 if __name__ == "__main__":
-    set_log_level("INFO")
+    set_log_level("WARNING")
```
je-cook, line 253: Can we leave this on INFO for the final merge

je-cook (changes requested):
A few things but overall looks good. Other than questions we've already talked about it looks nearly done, just some tests :)

eudemo/eudemo/maintenance/lower_port/designer.py
oliverfunk: Yea, thought it was more clear the way it is, but will change

eudemo/eudemo/maintenance/lower_port/designer.py
je-cook: Sure no problem

eudemo/eudemo/maintenance/lower_port/designer.py
je-cook: you could just get rid of the angled_cut variable as a half way house

eudemo/eudemo/maintenance/lower_port/designer.py
oliverfunk: Ah I think it's more readable this way :/

oliverfunk:
Putting these here for clarity

padding
params
y_offset

eudemo/eudemo/maintenance/lower_port/designer.py
```diff
@@ -0,0 +1,588 @@
...
+    boolean_fuse,
+    distance_to,
+    make_polygon,
+    offset_wire,
+)
+from bluemira.geometry.wire import BluemiraWire
+from eudemo.maintenance.lower_port.parameterisations import LowerPortDesignerParams
+
+
+class LowerPortDesigner(Designer):
```
CoronelBuendia, line 46: Not entirely clear to me what the scope of this designer is; it seems there are a lot of return values.

For the upper port, we have designer that just makes the keep-out zone, and then the builder handles everything relating to the y dimension. This is because we always make the ports as wide as they can be in y. Perhaps a similar separation can be achieved here.

This might be the best way to do it, but I think we can get away with a simpler parameterisation for the lower port keep out zone. Something like this:
image
To which you'd have to add various offsets for thicknesses and gaps of and between the thermal shield and VV.

CoronelBuendia (changes requested):
Hi @oliverfunk, this is looking pretty complete and works out of the box, so well done. I think we can reduce the scope of this PR a little by not including the integration with the reactor as you've done. There are also a few "scaffolding" elements that can be removed.

The parameterisation I think can also be slightly modified to be more in-line with the design intent of the port, and the builder could probably handle the y-related size calculations separately to the designer of the lower port keep out zone.

We also need a couple of unit tests of the designer and builder. Once this is in I can take care of all the threading up with the rest of the reactor. It's hard to say beforehand what needs to change, as we need to make this work for the TS and VV with all the gaps etc.

oliverfunk:
Linked issue #2253

oliverfunk:
If you want to see the lower port with the reactor, add the following:

```py
class EUDEMO(Reactor):
    """EUDEMO reactor definition."""

    plasma: Plasma
    vacuum_vessel: VacuumVessel
    vv_thermal: VacuumVesselThermalShield
    divertor: Divertor
    blanket: Blanket
    tf_coils: TFCoil
    pf_coils: PFCoil
    coil_structures: CoilStructures
    cryostat: Cryostat
    cryostat_thermal: CryostatThermalShield
    radiation_shield: RadiationShield
    lower_port: ComponentManager
```

After lower_port is returned from the function:
```py
reactor.lower_port = ComponentManager(lower_port)
```

Here are some screenshots for clarity:

image

image

image

CoronelBuendia (approved):
I haven't looked at this in detail since we talked, but as it was nearly there last time, I reckon it should be good to go.

When I start integrating the ports together, I will come knocking! It's possible we might need to tweak stuff then, but it's hard to say before I give it a serious go.

je-cook (approved):

PR #2120, opened 2023-04-11
✨ Rearrange Reactor and component Manager to have a common base

### LINKED ISSUE #526: Simplify plotting and displays for multiple Components

In many cases we will have a Component tree that contains nodes like "xy", "xz", "xyz" that may exist under system-like components. However, to plot these it is necessary to add the components separately to an axis, and to display them it is necessary to add each component to a component displayer. Neither of these approaches are particularly user friendly.

## Proposed solution

Allow the name to be plotted / displayed to be added to `show_cad`, `plot_2d`, `plot_3d` and perform the corresponding operations on all components corresponding to that name.

## Alternative solutions

We could keep things as they are, but it would be nice to hide some of the implementation details away.

je-cook:
## Linked Issue

Closes #526 as plotting functionality equal to that on reactor would exist on a component too

## Description

This slightly rearranges `Reactor` and `ComponentManager` to have a common base among some other small changes. Open for comment more than immediate merge.

- Rearranges Reactor and ComponentManager to have a common base
- Adds some plotting functionality to the BaseManager
- Adds a build timer to Reactor
- Adds the ability to call tree on a ComponentManager and Reactor

~TODO probably write some tests~

separately we should consider the location of `ComponentManager` because its location in `/base/builder.py` doesnt really make sense. If this PR goes in it could live in `base/reactor.py` (or either way). In any case that movement would coincide with a bump to v2 

## Interface Changes

Should be None, more addition of new functionality

## Checklist


bluemira/base/reactor.py
```diff
@@ -94,6 +275,12 @@ def component(
         """Return the component tree."""
         return self._build_component_tree(with_components)
 
+    def build_time(self) -> float:
+        """
+        Get time since initialisation
+        """
+        return time.perf_counter() - self.start_time
```
hsaunders1904, line 282: This method may be a bit misleading if the `Reactor` is being used in an interactive environment, e.g., in a Jupyter Notebook. The docstring is clear that it's time since initialization, not time spent building, but the function name is less clear on that. `time_since_init` is more explicit, but naming this is a bit tricky, I'll admit. Happy to leave it if you are.

hsaunders1904 (approved):
This all looks good to me. Makes all those empty `ComponentManager`s we have useful - nice one!

Small comment on a function name, but not a dealbreaker.

bluemira/base/reactor.py
je-cook: yep prefer your name, changed!

Issue #181, opened 2021-09-15
Include relevant paths below header


DanShort12:
In GitLab by @je-cook on Apr 27, 2021, 09:50

Add below the banner header all the relevant paths for the user. 

You then don't need to spout any messages or warnings about save locations and it is all there for posterity if you come back to a run and you've saved stdout somewhere.

CoronelBuendia:
@je-cook is this still necessary?

je-cook:
It depends whether we want to start printing the header again when running a reactor design, if so it would be nice to have all the locations we're writing to in the logs via this header. 

Probably worth thinking about for general usability when we come to niceties. Also a wider question is whether we should start using the `FileManager` again for its stated use :)

PR #2136, opened 2023-04-18
Reorganise logging and add back banner

je-cook:
## Linked Issues

This adds the opportunity to solve #181 if we organise a place to save data

## Description

This change means we would log INFO and more verbose to stdout and WARNING and less verbose to stderr instead of all to stderr.

It also adds back the banner printing on the program initialisation (well if anything from look and feel is imported). It feels a bit hacky but it works. Possible future downsides it its current form you cant disable all logging before we output text. 

## Checklist


CoronelBuendia:
Obsolete, can add back banner separately if desired

PR #2158, opened 2023-04-25
✅ Allow all tests to run in parallel

je-cook:
## Description

With #1949 we could no longer run tests in parallel with `pytest-xdist` because the mesh file tries to be created multiple times on test collection. Once tests are parameterised in the collection they get split up logically so this problem is sidestepped. 

This modifies a couple of tests to make lazy style fixtures and use setup_method where appropriate.

## Checklist


hsaunders1904 (approved):
Looks fine to me. This does change the behaviour slightly, as it's now running the setup before every test, instead of once. It's probably better this way, though, as it isolates the tests more, and it doesn't take too much longer (~0.1s difference on my machine).

PR #2159, opened 2023-04-25
EUDEMO blanket builder with panels

### LINKED ISSUE #2091: EUDEMO First wall panelling Builder

Following the implementation of the `PanellingDesigner` in #2060, we should implement a builder to build the CAD of the panels.

hsaunders1904:
## Linked Issues

Closes #2091 

## Description

Updates the EUDEMO blanket builder. It uses the new `PanellingDesigner` to cut panels into the existing blanket build.

## Interface Changes

The EUDEMO blanket builder now takes the blanket silhouette as two faces - the inboard and outboard.

## Checklist


eudemo/eudemo/ivc/plasma_face.py
```diff
@@ -91,7 +92,7 @@ def run(self) -> Tuple[BluemiraFace, BluemiraFace]:
         rm_clearance_face = _make_clearance_face(
             vessel_bbox.x_min,
             vessel_bbox.x_max,
-            self.wall_boundary.bounding_box.z_min,
+            self.wall_boundary.start_point().z,
```
CoronelBuendia, line 95: This may not be very robust, as it will rely on the orientation, or rather, direction of the wire. I appreciate bounding_boxes are also not particularly robust!

eudemo/eudemo_tests/blanket/test_designer.py
```diff
@@ -0,0 +1,203 @@
...
+            self.silhouette,
+            r_inner_cut=r_inner_cut,
+            cut_angle=cut_angle,
+        )
+
+        with mock.patch.object(designer, "segment_blanket") as sb_mock:
+            sb_mock.return_value = blanket
+            ib, ob = designer.run()
+
+        # These areas were (painstakingly) worked out by hand
```
CoronelBuendia, line 157: :D nice

eudemo/eudemo/blanket/designer.py
```diff
@@ -0,0 +1,210 @@
...
+                f"BB poloidal boundary segmentation only returned '{len(boundary_cut)}' "
+                "wire(s)."
+            )
+        if len(boundary_cut) > 2:
+            bluemira_warn(
+                "The BB poloidal boundary segmentation operation returned more than 2 "
+                f"wires ({len(boundary_cut)}); only taking the first two..."
+            )
+        ib_bound, ob_bound = sorted(boundary_cut, key=lambda x: x.center_of_mass[0])[:2]
+        return ib_bound, ob_bound
```
CoronelBuendia, line 210: Sorry to be a pain, but this could be the same function, but with some minor handling on the error message text. Or you could just lump the two together in once function and halve the number of warnings and errors, as if one warns/fails so will the other.

CoronelBuendia (changes requested):
Looks really good, mostly minor stuff.

eudemo/config/params.json
```diff
@@ -284,6 +284,20 @@
     "source": "Input",
     "long_name": "Non-inductive current drive fraction"
   },
+  "fw_a_max": {
+    "value": 25,
+    "unit": "degree",
+    "source": "Input",
+    "long_name": "Maximum panel angle",
+    "description": "Maximum angle of rotation between adjacent first wall panels"
```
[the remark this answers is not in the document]
hsaunders1904: I've made an attempt to clarify

eudemo/eudemo/ivc/plasma_face.py
hsaunders1904: The shape _should_ be an open wire, where both ends are equal in z. I've changed it to take the minimum of the start and end and added a comment.

eudemo/eudemo/blanket/designer.py
hsaunders1904: You're right. I remember being lazy and just copy-and-pasting and thinking I'd come back to it - but obviously forgot 😅

eudemo/eudemo/ivc/plasma_face.py
CoronelBuendia: Both ends will be equal in z on in the case of a horizontal divertor splitting line. In future (read, near future) we're likely going to want to split the divertor from the blankets based e.g. on the angle of the lower port. This will undoubtedly result in both ends not having the same z value.

CoronelBuendia (approved):

je-cook (approved):

PR #2164, opened 2023-04-28
WIP / probably not what we want:  Don't plot/display `PhysicalComponent` with `Void` material

### LINKED ISSUE #2163: Support "Void" material for construction geometry and void volume

We need a way to pass around void volumes, which can be used for construction or e.g. for working out things like enclosed volume in the case of the VV, and other stuff.

We do not want to see these void volume when displaying CAD though.

## Proposed solution

Add a Void material which can be applied to a `PhysicalComponent` via it's `__init__` which are stored in the tree of a `Component` (as usual) but are not shown when using `show_cad`.

## Additional context

@ivanmaione FYI

CoronelBuendia:
## Linked Issues

Closes #2163 

## Checklist

### DIFF
--- a/bluemira/display/displayer.py
+++ b/bluemira/display/displayer.py
@@ -291,6 +291,31 @@ class ComponentDisplayer(BaseDisplayer):
     CAD displayer for Components
     """
 
+    def _get_component_display_info(self, comps):
+        import bluemira.base.components as bm_comp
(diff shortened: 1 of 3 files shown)


CoronelBuendia:
This is not what we want: we do not want to hard-code stuff like this.

PR #2166, opened 2023-04-28
Add another way to get bounding boxes that is more accurate.

### LINKED ISSUE #2162: Bounding box bug
## Describe the bug

While working on #2159 I was coming across a bug where the blanket I was building had a 'lip' at the bottom that wasn't being panelled. After some investigation, it turned out that, in the `design_ivc` function, when we create the 'cut wall boundary' wire (i.e., the wall boundary cut below the equilibrium's x-point, to make room for the divertor), the wire's bounding box was not consistent with the rest of the object. The bounding box's z-value was ~0.2 smaller than it should have been.

The cut wall boundary looks like this:

The green line shows the z given by `cut_wall_boundary.start_point().z[0]` the red shows the z given by `cut_wall_boundary.bounding_box.z_min`.

See [34228d3](https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/2159/commits/34228d30325b55078210690acd6b4529acf2da36) for more context.

## Steps to reproduce

I've tried to get an MRE, but I haven't managed to reproduce the bug 😞. Here are the steps to recreate the bug in the way that I found it.

1. Checkout [34228d3](https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/2159/commits/34228d30325b55078210690acd6b4529acf2da36) in Git (`git checkout 34228d3`).
2. Add a break point to `eudemo/eudemo/ivc/__init__.py` line 76  in your IDE/debugger. 
3. Debug the EUDEMO reactor script.
4. To view the issue, run the following in the debug console:

	```pycon
	>>> cut_wall_boundary.bounding_box.z_min
	-5.701225441085458
	>>> cut_wall_boundary.start_point().z[0]
	-5.525853343944378
	>>> from bluemira.display import show_cad
	>>> show_cad([plasma_face, cut_wall_boundary], backend="POLYSCOPE")
	>>> cut_wall_boundary.bounding_box.z_min
	-5.525853343944378
	```

As you can see, when plotted, the shape corrects itself somehow. Honestly, this seems quite inconsistent, I'm still not 100% sure what's going on with it.

## Expected behaviour

The bounding box should be consistent with the start and end points of the wire.

- OS:  Ubuntu 22.04 (WSL2)

CoronelBuendia:
## Linked Issues

Closes #2162 

## Description

Adds a method onto `BluemiraGeo` to get an optimised bounding box. We haven't discussed the API of this @ivanmaione, I was just looking into making a MWE and found an easy solution. Happy to switch this around if necessary.

I have played with some 3-D solids and it works too but obviously the tesselation is much slower.

## Checklist

### DIFF
--- a/bluemira/codes/_freecadapi.py
+++ b/bluemira/codes/_freecadapi.py
@@ -667,22 +667,31 @@ def bounding_box(obj: apiShape) -> Tuple[float, float, float, float, float, floa
     return box.XMin, box.YMin, box.ZMin, box.XMax, box.YMax, box.ZMax
 
 
-def tesselated_bounding_box(
-    obj: apiShape, tolerance: float = 1.0
-) -> Tuple[float, float, float, float, float, float]:
+def tessellate(obj: apiShape, tolerance: float) -> None:
     """
-    Object's bounding box passing through tesselation
+    Tessellate a geometry object.
+
+    Parameters
+    ----------
+    obj:
+        Shape to tessellate
+    tolerance:
+        Tolerance with which to perform the operation
+
+    Raises
+    ------
+    ValueError:
+        If the tolerance is <= 0.0
 
     Notes
     -----
-    Tesselating less trivial geometry objects is the only way to get a more
-    accurate bounding box. Once tesselated, various attributes are modified
-    according to the the tolerance used.
-
-    Lower tolerances will not update the tesselation.
+    Once tesselated an object's properties may change. Tesselation cannot be reverted
+    to a previous lower value, but can be increased (irreversibly).
     """
+    if tolerance <= 0.0:
+        raise ValueError("Cannot have a tolerance that is less than or equal to 0.0")
+
     obj.tessellate(tolerance)
-    return bounding_box(obj)
 
 
 def start_point(obj: apiShape) -> np.ndarray:
--- a/bluemira/codes/_freecadapi.py
+++ b/bluemira/codes/_freecadapi.py
@@ -667,6 +667,24 @@ def bounding_box(obj: apiShape) -> Tuple[float, float, float, float, float, floa
     return box.XMin, box.YMin, box.ZMin, box.XMax, box.YMax, box.ZMax
 
 
+def tesselated_bounding_box(
+    obj: apiShape, tolerance: float = 1.0
+) -> Tuple[float, float, float, float, float, float]:
+    """
+    Object's bounding box passing through tesselation
+
+    Notes
+    -----
+    Tesselating less trivial geometry objects is the only way to get a more
+    accurate bounding box. Once tesselated, various attributes are modified
+    according to the the tolerance used.
+
+    Lower tolerances will not update the tesselation.
+    """
+    obj.tessellate(tolerance)
+    return bounding_box(obj)
+
+
 def start_point(obj: apiShape) -> np.ndarray:
     """The start point of the object"""
     point = obj.Edges[0].firstVertex().Point
--- a/bluemira/geometry/base.py
+++ b/bluemira/geometry/base.py
@@ -212,8 +212,28 @@ class BluemiraGeo(ABC, GeoMeshable):
     @property
     def bounding_box(self) -> BoundingBox:
         """
-        The bounding box of the shape."""
+        The bounding box of the shape.
+        """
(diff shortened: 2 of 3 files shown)


bluemira/codes/_freecadapi.py
```diff
@@ -667,6 +667,24 @@ def bounding_box(obj) -> Tuple[float, float, float, float, float, float]:
...
+
+    Notes
+    -----
+    Tesselating less trivial geometry objects is the only way to get a more
+    accurate bounding box. Once tesselated, various attributes are modified
+    according to the the tolerance used.
+
+    Lower tolerances will not update the tesselation.
+    """
+    obj.tessellate(tolerance)
```
CoronelBuendia, line 684: alternatively, we could just expose a `tesselate` method, @ivanmaione ?

ivanmaione: I am fine with exposing `tasselate`  (or, maybe, `_tasselate` seems more appropriate to me in this case). @je-cook, any thoughts?

CoronelBuendia: AS discussed with Ivan, went for _tessellate as a private method on `BluemiraGeo`, left the method for `get_optimal_bounding_box` as a short-hand, so that a user does not need to know about tesselate, but can still access a better bounding box.

Interestingly, as far as I can tell, tessellation does not appear to affect length, area or volume or COG calculations. Just the bounding box, but there may well be situations in which this does occur, or other properties affected, hence the various disclaimers regarding tessellation

bluemira/codes/_freecadapi.py
```diff
@@ -667,6 +667,33 @@ def bounding_box(obj: apiShape) -> Tuple[float, float, float, float, float, floa
...
+
+    Raises
+    ------
+    ValueError:
+        If the tolerance is <= 0.0
+
+    Notes
+    -----
+    Once tesselated an object's properties may change. Tesselation cannot be reverted
+    to a previous lower value, but can be increased (irreversibly).
```
je-cook, line 689: We currently use tessellate in the `collect_verts_faces` function for polyscope viewing. It does not copy the shape currently as it was not functionality exposed easily to the user. With this its now possible to get some weird ordering conditions with different tolerances. 

I think some more things need to be done if this is the way forward and I'm tentative about just adding a copy into polyscope because of speed implications.

Its a weird issue that it doesnt every warn if you try to decrease the tesselation otherwise this would be fine.

bluemira/codes/_freecadapi.py
je-cook: See my above comment, but I'm a bit tentative about this whole thing

bluemira/codes/_freecadapi.py
CoronelBuendia: OK, can confirm that appearance is irreversibly modified if tessellate is called with a higher tolerance than the default used in PolyScope. Visually this does not appear to be a problem, but when you turn off the smoothing it's immediately apparent.

Red shape: tessellation = 0.05
Blue shape: tessellation = 0.001

image

```python
from bluemira.geometry.tools import *
from bluemira.geometry.face import BluemiraFace
from bluemira.geometry.parameterisations import PolySpline
from bluemira.display import show_cad
from bluemira.display.displayer import DisplayCADOptions
from bluemira.base.components import PhysicalComponent, Component

w = PolySpline().create_shape()
f = BluemiraFace(w)
s = revolve_shape(f, degree=180)  # default tessellation = 0.05
s2 = s.deepcopy()
s2.rotate(degree=180)
s2._tessellate(tolerance=0.001)
s2._tessellate(tolerance=0.05)

from bluemira.display.displayer import ViewerBackend

p1 = PhysicalComponent("t005", s)
p1.display_cad_options = DisplayCADOptions(colour="r", backend=ViewerBackend.POLYSCOPE, tesselation=0.05)
p2 = PhysicalComponent("t0001", s2)
p2.display_cad_options = DisplayCADOptions(colour="b")
component = Component("test", children=[p1, p2])
component.show_cad(backend="POLYSCOPE")
```

will investigate getting the bounding box by tessellating a deepcopied version of `BluemiraGeo` such that the properties are not irreversibly modified and appearance is more consistent when viewed in PolyScope

PR #2194, opened 2023-05-04
Docstring convention enforcement marathon - part 2

CoronelBuendia:
## Linked Issues

Closes #{ID}

## Checklist

### DIFF
--- a/bluemira/mesh/meshing.py
+++ b/bluemira/mesh/meshing.py
@@ -677,7 +677,7 @@ class _FreeCADGmsh:
     @staticmethod
     def _fragment(
         dim: List[int] = [2, 1, 0],
-        all_ent=None,
+        all_ent: Optional[List[int]] = None,
         tools=[],
         remove_object: bool = True,
         remove_tool: bool = True,
@@ -702,7 +702,7 @@ class _FreeCADGmsh:
         return all_ent, oo, oov
 
     @staticmethod
-    def _map_mesh_dict(mesh_dict: dict, all_ent, oov):
+    def _map_mesh_dict(mesh_dict: dict, all_ent, oov: list = []):
         dim_dict = {
             "points_tag": 0,
             "cntrpoints_tag": 0,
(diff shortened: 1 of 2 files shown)


bluemira/mesh/meshing.py
```diff
@@ -668,7 +676,11 @@ def create_gmsh_curve(buffer):
 
     @staticmethod
     def _fragment(
-        dim=[2, 1, 0], all_ent=None, tools=[], remove_object=True, remove_tool=True
+        dim: List[int] = [2, 1, 0],
+        all_ent=None,
```
oliverfunk, line 680:
```suggestion
        all_ent: Optional[List[int]] = None,
```

oliverfunk, line 680:
```suggestion
        all_ent: Optional[List[int]] = None,
```

bluemira/mesh/meshing.py
```diff
@@ -690,7 +702,7 @@ def _fragment(
         return all_ent, oo, oov
 
     @staticmethod
-    def _map_mesh_dict(mesh_dict, all_ent, oov):
+    def _map_mesh_dict(mesh_dict: dict, all_ent, oov):
```
oliverfunk, line 705:
```suggestion
    def _map_mesh_dict(mesh_dict: dict, all_ent, oov: list = []):
```

oliverfunk (approved):
LGTM, two small additional types in the comments

CoronelBuendia:
> LGTM, two small additional types in the comments.
> 
> I've just noticed some inconsistent use of `dict` vs `Dict` (for example). How do you decide on when to use which?
> 
> We might need a larger discussion around moving to the lowercase types, after moving to py3.10+ As the `typing` module will be deprecated at some point https://docs.python.org/3/library/typing.html#deprecation-timeline-of-major-features

Bleurgh.. didnt know that.. We've been using upper case from typing for now, down the line I guess we'll have to change, its ok to have minor inconsistency. We discussed having a wider issue/discussion on typing.

PR #2224, opened 2023-05-09
Radiation source model and coupling with cherab

DarioV86:
branch: dvaccaro/import-radiation-model

Import radiation model from bluemira step-reactor

Radiation model to calculate the line radiation source [MW/m3] and functions to calculate the wall radiation heat flux [MW/m2] via cherab.

The theory about the models is:
Core radiation - Kovari, M., et al. "“PROCESS”: A systems code for fusion power plants—Part 1: Physics." Fusion Engineering and Design 89.12 (2014): 3054-3069.

SOL radiation - Stangeby, Peter C. The plasma boundary of magnetic fusion devices. Vol. 224. Philadelphia, Pennsylvania: Institute of Physics Pub., 2000.

### To install cherab:
`pip install cherab`


bluemira/radiation_transport/radiation_profile.py
je-cook, line 1: Can we split the functions in the first 1100 lines into a new file eg radiation_profile_tools or a better name and leave this file for the classes?

je-cook (commented):
first file looks good, just a couple of comments. Second file needs to be split

bluemira/radiation_transport/flux_surfaces_maker.py
```diff
@@ -0,0 +1,316 @@
...
+            first_wall.close()
+        return first_wall
+
+    @staticmethod
+    def _get_array_x_mp(flux_surfaces):
+        """
+        Get x_mp array of flux surface values.
+        """
+        x_mp = np.array([fs.x_start for fs in flux_surfaces])
+        return x_mp
```
je-cook, line 127: and for the others 
```suggestion
        return np.array([fs.x_start for fs in flux_surfaces])
```

bluemira/radiation_transport/flux_surfaces_maker.py
```diff
@@ -0,0 +1,316 @@
...
+        self.x_sep_omp, x_out_omp = self._get_sep_out_intersection(outboard=True)
+
+        self.flux_surfaces_ob_lfs = []
+        self.flux_surfaces_ob_hfs = []
+
+        start = self.x_sep_omp + self.dx_mp
+        stop = x_out_omp - EPS
+        step = ((x_out_omp - EPS - (self.x_sep_omp + self.dx_mp)) // self.dx_mp).astype(int)
+        range = np.linspace(start, stop, step)
+        for x in range:
```
je-cook, line 214: A bit simpler? The double division forces it to be an integer
```suggestion
        start = self.x_sep_omp + self.dx_mp
        for x in np.arange((x_out_omp - EPS - start) // self.dx_mp) + start:
```

DarioV86: It doesn't work in that way.
Cannot I do this?
`start = self.x_sep_omp + self.dx_mp`
`stop = x_out_omp - EPS`
`for x in np.arange(start, stop, self.dx_mp):`

je-cook: Oops yeah, sure if it works but there is no need to create new variables start and stop, just do it all on one line (un;less you need start/stop): 
`for x in np.arange(self.x_sep_omp + self.dx_mp, x_out_omp - EPS, self.dx_mp):`

je-cook:
At the moment cherab is only in these three functions at the bottom of `radiation_tools.py` in the main code. To me it looks definitely possible to have only 1 or 2 of them actually need cherab, but either way its makes it much easier to isolate. Need to have a think about the best way forward with that:

https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/7adf3662f5e077b5419317d5dc1b0bb016cabeb5/bluemira/radiation_transport/radiation_tools.py#L819
https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/7adf3662f5e077b5419317d5dc1b0bb016cabeb5/bluemira/radiation_transport/radiation_tools.py#L912
https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/7adf3662f5e077b5419317d5dc1b0bb016cabeb5/bluemira/radiation_transport/radiation_tools.py#L1010

There are some more imports in the examples however

je-cook:
Update just for clarity the example cherab imports now live in `radiation_tools.py` as well. The `Solver` will essentially be the `calculate_fw_rad_loads` function: 

https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/e39fc61403aec36e759a5668e0043cc69ae587bf/bluemira/radiation_transport/radiation_tools.py#L1099

CoronelBuendia:
Hi @DarioV86,

Apologies, I have not had time to look at this in detail.

The first major thing I can see needs to be changed is the treatment of the core "radiation profiles". Here it seems you have a lot of code dedicated to handling things like generating core temperature, density, and (possibly?) impurity profiles.

These are very common, and we use things like this all over the place. Rather than taking a certain parameterisation for `T(rho)` and `n(rho)`, it seems far simpler and more flexible to me take simply take the profiles as inputs.

At present we have some profile objects in `equilibria`, but there are not exactly what you will need, as you need profiles from transport solvers, e.g. PLASMOD. The simplest form for such profiles is just two arrays of length `n`: `rho: np.ndarray` and `value: np.ndarray`. Subsequently you can use linear interpolation to obtain further values should you require them. We can even discuss a base `Profile` class, which we already use in `equilibria` and to some extent in the PLASMOD interface. In future, I hope this will also be a feature in PROCESS.

The second major thing: it seems a real shame we can't get a 2-D axisymmetric radiation solution without adding a relatively heavy-duty dependency (CHERAB). I'm not against having such a dependency (for 3-D it seems like the best way). It's the difference between having an optional dependency for heavy-lifting, and a necessary dependency for light-weight stuff.

Finally, the duration it takes to review and converge upon a pull request is normally $`time \propto size^n`$ where $`n`$ is usually rather large..! In future, it'd be great if we could break this kind of stuff down into smaller pieces.

It's going to take me a while to understand what CHERAB is doing under the hood here, and it doesn't look like there are any mechanics to help me install it here (separate PR please!). In any case, I look forward to running it in September.

Best,
Matti

DarioV86:
Hi @CoronelBuendia ,
sorry for my late reply, got a bit busy this week.
About the first point (T and n profiles), yes, we already discussed about it. It should be relatively simple, and I am going to do it.
About the second point, CHERAB, it may be a bit more difficult and I may need some help. I searched a bit more about how it works, and actually it may already be in 2-D. I will try to talk again with Stuart Henderson next week. He used CHERAB in the past for the same purpose.
Thanks!
See you in September. Enjoy the summer.
Dario

bluemira/radiation_transport/flux_surfaces_maker.py
```diff
@@ -0,0 +1,303 @@
...
+        self.dx_mp = dx_mp
+
+        # Constructors
+        self.first_wall = None
+        self.flux_surfaces_ob_lfs = None
+        self.flux_surfaces_ob_hfs = None
+        self.flux_surfaces_ib_lfs = None
+        self.flux_surfaces_ib_hfs = None
+        self.x_sep_omp = None
+        self.x_sep_imp = None
```
je-cook, line 74: I think all of these should be calculated on init based on whether the provided equilibrium is an single null or a double null.

If the equilibrium is changed they can be recalculated but otherwise the first wall changing is the only thing analysed therefore everything else can be setup beforehand.

bluemira/radiation_transport/flux_surfaces_maker.py
```diff
@@ -0,0 +1,303 @@
...
+        """
+        Get z_fw array of flux surface values.
+        """
+        return np.array([fs.z_end for fs in flux_surfaces])
+    
+    def _get_array_alpha(flux_surfaces):
+        """
+        Get alpha angle array of flux surface values.
+        """
+        return np.array([fs.alpha for fs in flux_surfaces])
```
je-cook, line 150: All of these functions are currently unused. Also they all need the staticmethod decorator otherwise wont work.
I can see they could be useful at some point though.

Couldnt they all use `self.flux_surfaces` directly and therefore not need to be staticmethods or private methods? 

eg:
```python
    def get_array_alpha(self):
        """
        Get alpha angle array of flux surface values.
        """
        return np.array([fs.alpha for fs in self.flux_surfaces])
```

the same is true of `_get_array_x_mp` but it is used currently

bluemira/radiation_transport/flux_surfaces_maker.py
```diff
@@ -0,0 +1,303 @@
...
+    ----------
+    config: Dict[str, float]
+        The parameters for running the transport model. See
+        :class:`ChargedParticleSolverParams` for available parameters
+        and their defaults.
+    equilibrium: Equilibrium
+        The equilibrium defining flux surfaces.
+    dx_mp: float (optional)
+        The midplane spatial resolution between flux surfaces [m]
+        (default: 0.001).
```
je-cook, line 60: this doesnt reflect the init

bluemira/radiation_transport/flux_surfaces_maker.py
```diff
@@ -0,0 +1,303 @@
...
+            The z coordinates of the flux surface intersections
+        heat_flux: np.array
+            The perpendicular heat fluxes at the intersection points [MW/m^2]
+        """
+        self.first_wall = self._process_first_wall(first_wall)
+
+        if self.eq.is_double_null:
+            self._analyse_DN(first_wall)
+        else:
+            self._analyse_SN(first_wall)
```
je-cook, line 267: dx_omp should be returned from the single null and both dx_omp and dx_impo should be returned from the double null not stored on the class
```suggestion
        if self.eq.is_double_null:
            return self._analyse_DN(first_wall)
        else:
            return self._analyse_SN(first_wall)
```

bluemira/radiation_transport/flux_surfaces_maker.py
```diff
@@ -0,0 +1,303 @@
...
+
+        # Find the intersections of the flux surfaces with the first wall
+        self._clip_flux_surfaces(first_wall)
+
+        x_omp = self._get_array_x_mp(
+            self.flux_surfaces_ob_lfs
+        )
+
+        # Calculate values at OMP
+        self.dx_omp = x_omp - self.x_sep_omp
```
je-cook, line 283:
```suggestion
       return x_omp - self.x_sep_omp
```

bluemira/radiation_transport/flux_surfaces_maker.py
```diff
@@ -0,0 +1,303 @@
...
+
+        x_omp = self._get_array_x_mp(self.flux_surfaces_ob_lfs)
+
+        x_imp = self._get_array_x_mp(self.flux_surfaces_ib_lfs)
+
+        # Calculate values at OMP
+        self.dx_omp = x_omp - self.x_sep_omp
+
+        # Calculate values at IMP
+        self.dx_imp = abs(x_imp - self.x_sep_imp)
```
je-cook, line 303:
```suggestion
        dx_omp = x_omp - self.x_sep_omp

        # Calculate values at IMP
        dx_imp = abs(x_imp - self.x_sep_imp)
        
        return dx_omp, dx_imp
        
```

je-cook (changes requested):
I'm going to go through file by file here is `flux_surfaces_maker`

Summary
- The flux surfaces should be setup in the init based on the provided equilibrium
- analyse should return values not set them on the class

bluemira/radiation_transport/radiation_tools.py
```diff
@@ -0,0 +1,1121 @@
...
+from raysect.optical.observer import PowerPipeline0D
+from raysect.optical.observer.nonimaging.pixel import Pixel
+from raysect.optical.material import VolumeTransform
+from raysect.primitive import Cylinder
+
+# Scipy import
+from scipy.interpolate import LinearNDInterpolator, interp1d, interp2d
+
+
+SEP_CORRECTOR = 5e-3  # [m] DN: 5e-3, SN: 5e-2
```
je-cook, line 56:
```python
from dataclasses import dataclass

@dataclass
class SeparationCorrector:
    DN: float = 5e-3
    SN: float = 5e-2
```

bluemira/radiation_transport/radiation_tools.py
```diff
@@ -0,0 +1,1121 @@
...
+    p_sol: float,
+    lambda_q_near: float,
+    eq: Equilibrium,
+    r_sep_mp: float,
+    z_mp: float,
+    k_0: float,
+    firstwall_geom: Grid,
+    lfs=True,
+    connection_length=None,
+):
```
je-cook, line 257:
```suggestion
    separation_correction: float,
):
```

bluemira/radiation_transport/radiation_tools.py
```diff
@@ -0,0 +1,1121 @@
...
+    if lfs:
+        d = SEP_CORRECTOR
+    else:
+        d = -SEP_CORRECTOR
+
+    # Distance between the chosen point and the the target
+    if lfs:
+        forward = True
+    else:
+        forward = False
```
je-cook, line 318:
```suggestion
    # Distinction between lfs and hfs
    d =separation_corrector if lfs else -separation_corrector
    
    # Distance between the chosen point and the the target
```

bluemira/radiation_transport/radiation_tools.py
```diff
@@ -0,0 +1,1121 @@
...
+    l_p = calculate_connection_length_flt(
+        eq,
+        x_p + (d * f_exp),
+        z_p,
+        forward=forward,
```
je-cook, line 323:
```suggestion
        forward=lfs,
```

bluemira/radiation_transport/radiation_tools.py
```diff
@@ -0,0 +1,1121 @@
...
+    s_p = l_tot - l_p
+    if round(abs(z_p)) == 0:
+        s_p = 0
+    # Local temperature
+    t_p = ((t_u**3.5) - 3.5 * (q_par / k_0) * s_p) ** (2 / 7)
+
+    # From eV to keV
+    t_p = constants.raw_uc(t_p, "eV", "keV")
+
+    return t_p
```
je-cook, line 347:
```suggestion
    return  raw_uc(((t_u**3.5) - 3.5 * (q_par / k_0) * s_p) ** (2 / 7), "eV", "keV")
```

je-cook (changes requested):
Not huge quantities in this one. That may change when we isolate cherab.

- Need to run the pre-commit hooks.
- separation_connector should be used as a dataclass and selected by the user a default could be used.

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,2151 @@
...
+    Initial and generic class (no distinction between core and SOL)
+    to calculate radiation source within the flux tubes.
+    """
+
+    # fmt: off
+    default_params = [
+        ["n_e_sep", "Electron density at separatrix", 3e+19, "1/m^3", None, "Input"],
+        ["T_e_sep", "Electron temperature at separatrix", 2e-01, "keV", None, "Input"],
+    ]
+    # fmt: on
```
je-cook, line 80: All these default_params lists need to go now

```suggestion
    # fmt: off
    default_params = [
        ["n_e_sep", "Electron density at separatrix", 3e+19, "1/m^3", None, "Input"],
        ["T_e_sep", "Electron temperature at separatrix", 2e-01, "keV", None, "Input"],
    ]
    # fmt: on
```

```python
from dataclasses import dataclass

@dataclass
class RadiationParameterFrame(ParameterFrame):
    n_e_sep: Parameter[float] = Parameter(name="n_e_sep", long_name="Electron density at separatrix", value=3e19, unit='1/m^3', source="default")
```

this will convert all the values to SI so be warned.
We might want to not set any defaults with all the data in a config file somewhere but for now the above will work.
For the others you can do eg:

```python
@dataclass
class CoreRadiationParameterFrame(RadiationParameterFrame):
    n_e_0: Parameter[float] =  Parameter(name="n_e_0", long_name="Electron density on axis", value=1.81e+20, unit="1/m^3", source="default")

File: bluemira/radiation_transport/radiation_profile.py [hunk 15]
```diff
@@ -0,0 +1,2151 @@
...
+
+    def __init__(
+        self,
+        eq: Equilibrium,
+        flux_surf_solver: TempFsSolver,
```
je-cook, line 85: Need to change the type here as you now have a flux surface maker

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,2151 @@
...
+            self.flux_surf_solver,
+            self.params,
+            self.psi_n,
+            self.ne_mp,
+            self.te_mp,
+            self.imp_content_core,
+            self.imp_data_core,
+        )
+        core_rad.build_mp_rad_profile()
+        rad_tot = np.sum(np.array(core_rad.rad, dtype=object), axis=0).tolist()
```
je-cook, line 1895: I doubt you need to convert it to a list (you might need to call `.ravel()` if it needs to be 1d)
```suggestion
        rad_tot = np.sum(np.array(core_rad.rad, dtype=object), axis=0)
```

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,2151 @@
...
+        f_rad = interp1d(core_rad.rho_core, rad_tot)
+        rho_new = np.sqrt(psi_n)
+        rad_new = f_rad(rho_new)
+        return rad_new
```
je-cook, line 1899:
```suggestion
        return f_rad(np.sqrt(psi_n))
```

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,2151 @@
...
+            data["L_ref"] for key, data in impurity_data.items() if key != "Ar"
+        ]
+        self.impurity_symbols = impurity_content.keys()
+
+        # Midplane profiles
+        self.psi_n = psi_n
+        self.ne_mp = ne_mp
+        self.te_mp = te_mp
+
+    def build_mp_radiation_profile(self):
```
je-cook, line 398: I thought all of these profiles were meant to be coming in as inputs now? Something like this object or its children in equilibria:
https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/develop/bluemira/equilibria/profiles.py#L363

je-cook (changes requested):
A bit of work still to do here, the profiles should be passed in as ready made object to the radiation solver if I understand correctly

bluemira/radiation_transport/radiation_tools.py
DarioV86: Cannot I just write `sep_corrector` = SeparationCorrector.DN if eq.is_double_null else SeparationCorrector.SN` instead of giving it as parameter?

je-cook: It depends whether there will ever be a situation where it needs to be specified, (eg for a custom value) if not then its not needed.

Also your above snippet wont work because the class in not initialised, if you want it to work like that it should be an enum instead

bluemira/radiation_transport/flux_surfaces_maker.py
DarioV86: A few line became a big change 😅

je-cook: yeah sorry 😇

bluemira/radiation_transport/radiation_tools.py
```diff
@@ -0,0 +1,1109 @@
...
+    )
+
+    # Plot the wall and radiation distribution
+    fig, ax = plt.subplots(figsize=(12, 6))
+
+    gs = plt.GridSpec(2, 2, top=0.93, wspace=0.23)
+
+    ax1 = plt.subplot(gs[0:2, 0])
+    ax1.imshow(
+        np.squeeze(t_samples).transpose() * 1.0e-6,
```
je-cook, line 1027: this will fix the orientation of the equilibria but I dont think it is the root of the problem...
```suggestion
        np.squeeze(t_samples)[::-1].transptose() * 1.0e-6,
```

bluemira/radiation_transport/radiation_tools.py
```diff
@@ -0,0 +1,1109 @@
...
+        if p1 != p2:  # Ignore duplicate points
+
+            # evaluate y_vector
+            y_vector_full = p1.vector_to(p2)
+            y_vector = y_vector_full.normalise()
+            y_width = y_vector_full.length
+
+            # Check if the wall element is small enough
+            if y_width > max_wall_len:
+                n_split = np.int(np.ceil(y_width / max_wall_len))
```
je-cook, line 943:
```suggestion
                n_split = int(np.ceil(y_width / max_wall_len))
```

CoronelBuendia:
OK, @DarioV86 @je-cook, I have looked over the example a bit more, and read more of this PR. As mentioned the other day, I think that the user-facing API (as seen in the example) needs to be simplified. I also think that there are some terminological issues.

Below is a proposal, to be discussed, which revolves around my impression that at the moment the `RadiationSolver` is in fact a grouping of source terms, much like in e.g. magnetostatics. This is a bit of terminological grey zone, seeing as the sources do "solve" for the e.g. the field or the radiation, but I find this source distinction helpful in the sense that you have multiple sources of radiation here, which one does not necessarily need to use together. I skip over the combination of the core and SOL sources for now, keeping more or less the same interface as in the example, but moving them to the `RadiationSource` API below.

```python
source = RadiationSource(
          equilibrium: Equilibrium,
          profiles: Profiles,
          firstwall_shape: BluemiraWire,
          core_impurities: Dict[str, float],
          sol_impurities: Dict[str, float],
          params: ParameterFrame,
          settings: Optional[Dict] = None,
)
```
In other words, all of the extraction of flux surfaces and the retrieval of the impurity species data should be under the hood.

Methods for getting the source term values [W/m^3] could include:
```python
source.radiation_at_points(x, z)
source.radiation_at_psin(psi_n)
```
This would allow people to build their own custom callables for the source term when running in a radiation transport solver (e.g. CHREAB). You could also offer an easy and configurable API to make the source callable that is later used in CHERAB, a la:

```python
source_func: Callable[[float, float], float] = source.interpolator(n_fs_core, n_fs_SOL, dx_mp_sol, ...)
```

Plotting on sources technically should not be anything other than the actual geometry of the source. If I want to see what it would look like, then I'd do something like:
```python
x = np.linspace(4, 15, 100)
z = np.linspace(-10, 10, 100)
xx, zz = np.meshgrid(x, z)

radiation = source.radiation_at_points(xx, zz)
f, ax = plt.subplots()
ax.contourf(xx, zz, radiation)
```
Of course, one could make a more sophisticated discretisation.

A brief note on grouping, could be as simple as:
```python
source_group = RadiationSourceGroup(sources: List[RadiationSource] = [core_radiation_source, sol_radiation_source])
```

For me, a `RadiationSolver` could in fact be used instead of `calculate_fw_rad_loads`, perhaps with a name like `FirstWallRadiationSolver`, which does the heavy lifting in CHERAB. This would be a useful abstraction, because we could end up using a different radiation transport solver in future.

```python
solver = FirstWallRadiationSolver(source_func: callable, firstwall_shape: BluemiraWire)
solver.solve() -> x, z, hf
```

I'm still not 100% on where the first wall geometry should be specified, and it may be that it has to be embedded in e.g. the SOL `RadiationSource` but if we can avoid that it would be great, because then you wouldn't need to re-specify the source every time you want to solve with a different geometry, but the same equilibrium.

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,2173 @@
...
+        if self.eq.is_double_null:
+            sep_loop = self.sep_lfs if lfs else self.sep_hfs
+        else:
+            sep_loop = self.sep_ob if lfs else self.sep_ib
+        if z_main > z_pfr:
+            reg_i = np.where((sep_loop.z < z_main) & (sep_loop.z > z_pfr))[0]
+            i_in = np.where(sep_loop.z == np.max(sep_loop.z[reg_i]))[0]
+            i_out = np.where(sep_loop.z == np.min(sep_loop.z[reg_i]))[0]
+        else:
+            reg_i = np.where((sep_loop.z > z_main) & (sep_loop.z < z_pfr))[0]
```
je-cook, line 635: should these have a >= or <= and the else includes `z_main <= z_pfr` not just `z_main < z_pfr`?

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,2173 @@
...
+        -------
+        rad_i: np.array
+            indexes of the points within the radiation region
+        rec_i: np.array
+            indexes pf the points within the recycling region
+        """
+        if lower:
+            rad_i = np.where((flux_tube.z < z_main) & (flux_tube.z > z_pfr))[0]
+            rec_i = np.where(flux_tube.z < z_pfr)[0]
+        else:
```
je-cook, line 674: Similar question here, should somethere there be a <=/>=

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,2173 @@
...
+
+        Returns
+        -------
+        te_prof: np.array
+            radial decayed temperatures through the SoL. Unit [keV]
+        ne_prof: np.array
+            radial decayed densities through the SoL. Unit [1/m^3]
+        """
+        # Distinction between lfs and hfs
+        if lfs is True or not self.eq.is_double_null:
```
je-cook, line 760: already a bool
```suggestion
        if lfs or not self.eq.is_double_null:
```

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,2173 @@
...
+            target temperature [keV]
+        ne_t: np.array
+            target density [1/m^3]
+        """
+        if not detachment:
+            te_t = te_div
+            f_m = f_m
+        else:
+            te_t = [self.params.det_t] * len(te_div)
+            f_m = 0.1
```
je-cook, line 830:
```suggestion
        if detachment:
            te_t = [self.params.det_t] * len(te_div)
            f_m = 0.1
        else:
            te_t = te_div
```

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,2173 @@
...
+
+        # Coverting needed parameter units
+        t_u_ev = constants.raw_uc(t_u_kev, "keV", "eV")
+        p_sol = constants.raw_uc(self.params.P_sep, "MW", "W")
+        f_ion_t = constants.raw_uc(self.params.f_ion_t, "keV", "eV")
+
+        if lfs and self.eq.is_double_null:
+            p_sol = p_sol*self.params.lfs_p_fraction
+        elif not lfs and self.eq.is_double_null:
+            p_sol = p_sol*(1-self.params.lfs_p_fraction)
```
je-cook, line 967:
```suggestion
        if self.eq.is_double_null:
            p_sol *= self.params.lfs_p_fraction if lfs else (1-self.params.lfs_p_fraction)
```

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,2173 @@
...
+            plt.title("Temperature along flux surfaces")
+            plt.ylabel(r"$T_e~[keV]$")
+        else:
+            plt.title("Density along flux surfaces")
+            plt.ylabel(r"$n_e~[m^{-3}]$")
+
+        [
+            ax.plot(np.linspace(0, len(flux_tube.coords.x), len(flux_tube.coords.x)), val)
+            for flux_tube, val in zip(flux_tubes, property)
+        ]
```
je-cook, line 1169:
```suggestion
        for flux_tube, val in zip(flux_tubes, property):
            ax.plot(np.linspace(0, len(flux_tube.coords.x), len(flux_tube.coords.x)), val)
```

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,2173 @@
...
+            frac for key, frac in impurity_content.items() if key != "H"
+        ]
+
+        self.imp_data_t_ref = [
+            data["T_ref"] for key, data in impurity_data.items() if key != "H"
+        ]
+        self.imp_data_l_ref = [
+            data["L_ref"] for key, data in impurity_data.items() if key != "H"
+        ]
+        self.eq=eq
```
je-cook, line 1544: It already happens in the super
```suggestion
```

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,2173 @@
...
+            "hfs": {"density": t_and_n_pol_hfs[1], "loss": loss["hfs"]},
+        }
+        for side, ft in self.rad.items():
+            self.rad[side] = [
+                [
+                    calculate_line_radiation_loss(n, l_f, fi)
+                    for n, l_f in zip(ft["density"], f)
+                ]
+                for f, fi in zip(ft["loss"], self.impurities_content)
+            ]
```
je-cook, line 1702: bit weird to make a dictionary and then destroy
```suggestion
        rad_data = {
            "lfs": {"density": t_and_n_pol_lfs[1], "loss": loss["lfs"]},
            "hfs": {"density": t_and_n_pol_hfs[1], "loss": loss["hfs"]},
        }
        self.rad = {}
        for side, ft in rad_data.items():
            self.rad[side] = [
                [
                    calculate_line_radiation_loss(n, l_f, fi)
                    for n, l_f in zip(ft["density"], f)
                ]
                for f, fi in zip(ft["loss"], self.impurities_content)
            ]
```

je-cook (commented):
Couple of questions about radiation_profile inputs. Are the defaults Params needed? would you expect a user to provide them or know about them?

Also at the moment as written it wont work from a ParameterFrame I think.

lts much easier to read than it has been in past, well done!

Still a lot to digest so will come round again next week.

Also you need to run our pre-commit hooks `pre-commit run --all-files`. I would rebase on develop before you do that.

bluemira/radiation_transport/radiation_profile.py
DarioV86: @je-cook corretcted. I had just assumed that it was unlike to get exactly the same value of `z_main` or `z_pfr`

je-cook: yeah definitely unlikely but could lead to a weird bug

CoronelBuendia (changes requested):
Hi @DarioV86, 

Thanks for this, it's looking so much better now! I have a few points still, but I haven't yet really had the chance to play around with it more than just running the example... The example runs and looks nice, but it is perhaps quite a high resolution (slow runtime for me) for a quick example.

Hopefully later this week I will find the time to look at some of the other API. Frankly, from just looking at the code, it feels like there are quite a few user-facing (i.e. not private _methods) on the `Radiation` classes. It might be wise to keep these to a minimum, at least for the time being.

There are still a few hard-coded magic numbers (e.g. 0.95, 0.1, etc.) which are still in there, which might also be better defined, or at least made into some form of constant.

Anyway, I am in favour of getting this in ASAP, and we can do some of the finer points down the line. That's part of the motivation for suggesting to keep the user-facing API to a minimum, because it means we are less "boxed in" later when it comes to API-breaking changes.

bluemira/radiation_transport/midplane_temperature_density.py
```diff
@@ -0,0 +1,190 @@
...
+            electron temperature at the mid-plane. Unit [keV]
+        psi_n:
+            rho_core**2
+
+        Notes
+        -----
+        The region that extends through the plasma core until its
+        outer layer, named pedestal, is referred as "interior".
+        The region that extends from the pedestal to the separatrix
+        is referred as "exterior".
```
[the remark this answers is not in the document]
DarioV86: Fair. I don't know. I will rename them

bluemira/radiation_transport/midplane_temperature_density.py
```diff
@@ -0,0 +1,190 @@
...
+            - (
+                (rho_core[i_interior] ** self.params.t_beta.value)
+                / (self.rho_ped**self.params.t_beta.value)
+            )
+        ) ** self.params.alpha_t.value
+
+        ne_i = self.params.n_e_ped.value + (n_grad_ped0 * rho_ratio_n)
+        te_i = teped_keV + (t_grad_ped0 * rho_ratio_t)
+
+        i_exterior = np.where((rho_core > self.rho_ped) & (rho_core <= 1))[0]
```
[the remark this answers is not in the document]
DarioV86: I cannot remember why the "=". It's probably as you said, protection. But I removed them and it works.

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -0,0 +1,2167 @@
...
+    upstream_temperature,
+)
+
+if TYPE_CHECKING:
+    from bluemira.equilibria.equilibrium import Equilibrium
+    from bluemira.equilibria.grid import Grid
+    from bluemira.geometry.wire import BluemiraWire
+
+
+class SeparationCorrector(Enum):
```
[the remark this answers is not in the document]
DarioV86: It's now a user input

bluemira/radiation_transport/midplane_temperature_density.py
```diff
@@ -0,0 +1,190 @@
...
+        Calculation of core dimensionless radial coordinate rho.
+
+        Returns
+        -------
+        rho_core:
+            dimensionless core radius. Values between 0 and 1
+        """
+        # The plasma bulk is divided into plasma core and plasma mantle according to rho
+        # rho is a nondimensional radial coordinate: rho = r/a (r varies from 0 to a)
+        self.rho_ped = (self.params.rho_ped_n.value + self.params.rho_ped_t.value) / 2.0
```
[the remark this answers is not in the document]
DarioV86: Honestly, I just took the calculation from PROCESS. Relying on the fact that PROCESS and the core radiation specifically are documented.

CoronelBuendia (approved):
Quick look LGTM but two things for future ASAP:
* Add benchmark test for the case checked with PFCFLUX
* Get rid of duplication in flux surface tools

je-cook (approved):

PR #2167, opened 2023-04-28
Add voidspace to VV and TS in builders and EUDEMO

CoronelBuendia:
## Linked Issues

Closes #{ID}

We need to keep track of void spaces for construction purposes and also for later analysis of enclosed volumes.

This adds some voidspace tracking to the thermal shields and vacuum vessel (not including ports).

This will make stuff look ugly until #2164 is in however...

image

## Checklist

I confirm that I have completed the following checks:

- [x] Tests run locally and pass `pytest tests --reactor`
- [x] Code quality checks run locally and pass `flake8` and `black .`
- [x] Documentation built locally and checked `sphinx-build -W documentation/source documentation/build`

PR #2236, opened 2023-05-11
Material filtering on Managers

### LINKED ISSUE #2163: Support "Void" material for construction geometry and void volume

### LINKED ISSUE #2207: `Component` display options: filtering leaves by certain rules

At the moment, when we view a `Component` with either `show_cad` or `plot_2d`, we have no control over what components are shown.

Given we are now starting to keep track of void geometries, it would be nice to have flexible ways of filtering which components are shown, perhaps with some sensible defaults (i.e. don't show voids by default).

For example, we may in future want to only show the children of a component which are made from e.g. EUROfer.

Note that all solutions should work for `show_cad` and `Component.show_cad`, same for plotting.

## Proposed solution

The underlying mechanics can be as necessary to provide the below behaviour.

Something like (APDL-like API), for example:
```python
from typing import Optional, Tuple, Any, List
from bluemira.materials.cache import Void, EUROfer
from from bluemira.base.components import Component

my_massive_component.show_cad("xyz", filters: Optional[List[Tuple[str, str, Any]]]]=[["u", "material", Void]])
# show everything except voids

my_massive_component.show_cad("xyz", filters: Optional[List[Tuple[str, str, Any]]]]=[["s", "material", EUROfer]])
# show only EUROfer stuff
```

Making this work for the `show_cad` and `plot_2d` functions could be ugly however, as technically, we can pass a list of Components, which would each need their filter rules in some iterable.

Alternative suggestions for API very welcome! Could perhaps also use a `filters=List[Callable[[PhysicalComonent], bool]]`,

## Additional context

#2164 was a lacklustre attempt to hide Voids, but is not flexible at all. I will close it shortly.
@ivanmaione @je-cook  FYI, please comment, as soon as we converge on an API, @je-cook is happy to address.

je-cook:
## Linked Issues

Closes #2207 
Closes #2163 

## Description

This PR adds some filtering opportunities to the `ComponentManager` and `Reactor` it doesn't use the  `Component._get_thing` filtering mechanism that I add last year (and forgot about) so that the tree is kept intact in the filtering. 

At the moment you can only filter on `PhysicalComponent` properties which I will try and find a way around but as `Component` doesnt hold and information and is just a node linker that may be acceptable. We can already filter by node with the combination of `with_component` and `dims`.

UPDATE: restriction removed

## Checklist

### DIFF
--- a/bluemira/base/components.py
+++ b/bluemira/base/components.py
@@ -92,7 +92,7 @@ class Component(NodeMixin, Plottable, DisplayableCAD):
     def filter_components(
         self,
         names: Iterable[str],
-        filter_: Optional[Callable[[anytree.Node], bool]] = None,
+        filter_: Optional[Callable[[PhysicalComponent], bool]] = None,
     ):
         """
         Removes all components from the tree, starting at this component,
@@ -103,6 +103,8 @@ class Component(NodeMixin, Plottable, DisplayableCAD):
         ----------
         names:
             The list of names of each component to search for.
+        filter_:
+            A callable to filter PhysicalComponents from the Component tree
 
         Notes
         -----
@@ -201,7 +203,7 @@ class Component(NodeMixin, Plottable, DisplayableCAD):
 
     def get_component(
         self, name: str, first: bool = True, full_tree: bool = False
-    ) -> Union[Component, Iterable[Component], None]:
+    ) -> Union[Component, Tuple[Component], None]:
         """
         Find the components with the specified name.
 
@@ -283,7 +285,7 @@ class Component(NodeMixin, Plottable, DisplayableCAD):
 
     def _get_thing(
         self, filter_: Union[Callable, None], first: bool, full_tree: bool
-    ) -> Union[Component, Iterable[Component], None]:
+    ) -> Union[Component, Tuple[Component], None]:
         found_nodes = anytree.search.findall(
             self.root if full_tree else self, filter_=filter_
         )
--- a/bluemira/base/components.py
+++ b/bluemira/base/components.py
@@ -89,7 +89,7 @@ class Component(NodeMixin, Plottable, DisplayableCAD):
         """
         return self.name + " (" + self.__class__.__name__ + ")"
 
-    def filter_components(self, names: Iterable[str], filter_: Union[Callable, None]):
+    def filter_components(self, names: Iterable[str], filter_: Optional[Callable[[anytree.Node], bool]] = None):
         """
         Removes all components from the tree, starting at this component,
         that are siblings of each component specified in `names`
--- a/bluemira/base/components.py
+++ b/bluemira/base/components.py
@@ -89,7 +89,7 @@ class Component(NodeMixin, Plottable, DisplayableCAD):
         """
         return self.name + " (" + self.__class__.__name__ + ")"
 
-    def filter_components(self, names: Iterable[str]):
+    def filter_components(self, names: Iterable[str], filter_: Union[Callable, None]):
         """
         Removes all components from the tree, starting at this component,
         that are siblings of each component specified in `names`
@@ -120,6 +120,14 @@ class Component(NodeMixin, Plottable, DisplayableCAD):
                 for c_sib in c.siblings:
                     if c_sib.name not in names:
                         c_sib.parent = None
+
+                for c_child in c.descendants:
+                    if (
+                        isinstance(c_child, PhysicalComponent)
+                        and filter_ is not None
+                        and not filter_(c_child)
+                    ):
+                        c_child.parent = None
 
     def tree(self) -> str:
         """
@@ -189,7 +197,7 @@ class Component(NodeMixin, Plottable, DisplayableCAD):
 
     def get_component(
         self, name: str, first: bool = True, full_tree: bool = False
-    ) -> Union[Component, Tuple[Component], None]:
+    ) -> Union[Component, Iterable[Component], None]:
         """
         Find the components with the specified name.
 
@@ -269,7 +277,9 @@ class Component(NodeMixin, Plottable, DisplayableCAD):
             ]
             return tuple(map(list, zip(*node_properties)))
 
-    def _get_thing(self, filter_: Union[Callable, None], first: bool, full_tree: bool):
+    def _get_thing(
+        self, filter_: Union[Callable, None], first: bool, full_tree: bool
+    ) -> Union[Component, Iterable[Component], None]:
         found_nodes = anytree.search.findall(
             self.root if full_tree else self, filter_=filter_
         )
--- a/bluemira/base/components.py
+++ b/bluemira/base/components.py
@@ -92,7 +92,7 @@ class Component(NodeMixin, Plottable, DisplayableCAD):
     def filter_components(
         self,
         names: Iterable[str],
-        filter_: Optional[Callable[[PhysicalComponent], bool]] = None,
+        filter_: Optional[Callable[[Component], bool]] = None,
     ):
         """
         Removes all components from the tree, starting at this component,
@@ -104,7 +104,8 @@ class Component(NodeMixin, Plottable, DisplayableCAD):
         names:
             The list of names of each component to search for.
         filter_:
-            A callable to filter PhysicalComponents from the Component tree
+            A callable to filter Components from the Component tree,
+            returning True keeps the node False removes it
 
         Notes
         -----
@@ -128,11 +129,7 @@ class Component(NodeMixin, Plottable, DisplayableCAD):
                         c_sib.parent = None
 
                 for c_child in c.descendants:
-                    if (
-                        isinstance(c_child, PhysicalComponent)
-                        and filter_ is not None
-                        and not filter_(c_child)
-                    ):
+                    if filter_ is not None and not filter_(c_child):
                         c_child.parent = None
 
     def tree(self) -> str:
--- a/bluemira/base/components.py
+++ b/bluemira/base/components.py
@@ -269,7 +269,7 @@ class Component(NodeMixin, Plottable, DisplayableCAD):
             ]
             return tuple(map(list, zip(*node_properties)))
 
-    def _get_thing(self, filter_: Callable, first: bool, full_tree: bool):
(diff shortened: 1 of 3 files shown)


CoronelBuendia:
Here is my review from a user perspective (only!):

I've pulled this into #2167 for testing / API interface purposes.

`reactor.vacuum_vessel.show_cad()`
image

`reactor.vacuum_vessel.show_cad(filter_=None)`
image

`reactor.vacuum_vessel.show_cad("xz")`
image

`reactor.vacuum_vessel.show_cad("xz", filter=None)`
image

I can confirm that
`reactor.vacuum_vessel.plot("xz")` and `reactor.vacuum_vessel.plot("xz", filter_=None)` work in exactly the same way.

Please note that the color (black) of the void is not due to this PR but to #2167.

Summary:
* My desired default behaviour works for component managers
* The void `PhysicalComponents` exist in the same tree as everything else
* One can easily override the default behaviour
* One can with _relative_ ease specify one's own filter `Callable([PhysicalComponent], bool)`

I am happy with this from a user perspective.

bluemira/base/components.py
```diff
@@ -89,7 +89,7 @@ def __repr__(self) -> str:
         """
         return self.name + " (" + self.__class__.__name__ + ")"
 
-    def filter_components(self, names: Iterable[str]):
+    def filter_components(self, names: Iterable[str], filter_: Union[Callable, None]):
```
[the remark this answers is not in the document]
je-cook: thanks!

CoronelBuendia (approved):

bluemira/base/components.py
```diff
@@ -189,7 +201,7 @@ def copy_children(
 
     def get_component(
         self, name: str, first: bool = True, full_tree: bool = False
-    ) -> Union[Component, Tuple[Component], None]:
+    ) -> Union[Component, Iterable[Component], None]:
```
hsaunders1904, line 204: I think we should be as specific as possible when it comes to return types, and not use general types like `Iterable`. This function is a bit confusing, though. It sometimes returns a list and sometimes a tuple. 🤔  

In my ideal world, this would _always_ return a list. I think it puts an undue burden on the caller to need to check their return types every time they use the function. Not to mention how much our use of this function upsets MyPy.

Back to reality, though, my take is that we should revert the change to this line, and update [line 267](https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/2236/files#diff-4557c02eb70464b83b973f6f28f0ae18a4ce0b351f00dddbb96c3c1acedc5cc2L264) to return a tuple (technically an API change, but we can probably get away with it). Thoughts?

bluemira/base/components.py
```diff
@@ -269,7 +281,9 @@ def filter_(node, properties):
             ]
             return tuple(map(list, zip(*node_properties)))
 
-    def _get_thing(self, filter_: Callable, first: bool, full_tree: bool):
+    def _get_thing(
+        self, filter_: Union[Callable, None], first: bool, full_tree: bool
+    ) -> Union[Component, Iterable[Component], None]:
```
hsaunders1904, line 286: Again, we should be specific about the return type (which I think is `Tuple` here?)

bluemira/base/reactor.py
```diff
@@ -72,7 +75,7 @@ def show_cad(
         """
 
     @abc.abstractmethod
-    def plot(self, *dims: str):
+    def plot(self, *dims: str, filter_: Union[Callable, None]):
```
hsaunders1904, line 78: What form should this callable take? It also needs adding to the docstring

bluemira/base/reactor.py
```diff
@@ -129,7 +132,12 @@ def _validate_plot_dims(self, *dims) -> Tuple[str, ...]:
 
         return dims_to_show
 
-    def _filter_tree(self, comp: Component, dims_to_show: Tuple[str, ...]) -> Component:
+    def _filter_tree(
+        self,
+        comp: Component,
+        dims_to_show: Tuple[str, ...],
+        filter_: Union[Callable, None],
```
hsaunders1904, line 139: Form of the callable?

bluemira/base/reactor.py
```diff
@@ -340,6 +397,7 @@ def show_cad(
         *dims: str,
         with_components: Optional[List[ComponentManager]] = None,
         n_sectors: Optional[int] = None,
+        filter_: Union[Callable, None] = FilterMaterial(),
```
hsaunders1904, line 400: Ditto

hsaunders1904 (changes requested):

bluemira/base/reactor.py
```diff
@@ -376,7 +435,12 @@ def show_cad(
 
         ComponentDisplayer().show_cad(comp_copy, **kwargs)
 
-    def plot(self, *dims: str, with_components: Optional[List[ComponentManager]] = None):
+    def plot(
+        self,
+        *dims: str,
+        with_components: Optional[List[ComponentManager]] = None,
+        filter_: Union[Callable, None] = FilterMaterial(),
```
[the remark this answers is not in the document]
je-cook: I was thinking the same but I havent come up with a nice way to avoid as None is a valid value of 'no filter'. I'll have another think

bluemira/base/components.py
je-cook: It always returns a tuple at the moment (I just was mucking about with stuff..). 

`get_component_properties` would need to change L284 not L276 if you wanted to return lists but thats a bit unrelated to this change. 
That function at the moment returns lists for the grouping of a selected property and a tuple of there is more than 1 property, the same as the normal return of any function that has multiple returns (it also returns a single property if that is all thats found).

bluemira/base/reactor.py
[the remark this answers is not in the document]
je-cook: How do you feel about, the class is then basically immutable unless you try really hard...a bit of a hack though:

```python
class FilterMaterial:

    def __init__(self, material):
        super().__setattr__("material", material)

    def __call__(self, node) -> bool:
        return print(self.material)

    def __setattr__(self, name: str, value: Any):
        raise AttributeError(f"{type(self).__name__} is immutable")
```

bluemira/base/components.py
hsaunders1904: Ah, sorry, I think I was getting a bit confused between this function and `get_component_properties`, my bad.

bluemira/base/reactor.py
```diff
@@ -139,17 +153,76 @@ def _filter_tree(self, comp: Component, dims_to_show: Tuple[str, ...]) -> Compon
...
+        Override setattr to force immutability
+
+        This method makes the class nearly immutable as no new attributes
+        can be modified or added by standard methods.
+
+        See #2236 discussion_r1191246003 for further details
+        """
+        raise AttributeError(f"{type(self).__name__} is immutable")
+
+    def _filterer(
```
oliverfunk, line 212: Think the name is bit funny, maybe _apply_filters?

bluemira/base/reactor.py
```diff
@@ -139,17 +153,76 @@ def _filter_tree(self, comp: Component, dims_to_show: Tuple[str, ...]) -> Compon
...
+        self, material: Union[SerialisedMaterial, Tuple[SerialisedMaterial]]
+    ) -> bool:
+        bool_store = True
+
+        if self.material is not None:
+            bool_store = isinstance(material, self.material)
```
oliverfunk, line 218: But if self.material is a Tuple, this won't work.

oliverfunk (commented):
Left a few comments

bluemira/base/components.py
```diff
@@ -89,7 +89,11 @@ def __repr__(self) -> str:
         """
         return self.name + " (" + self.__class__.__name__ + ")"
 
-    def filter_components(self, names: Iterable[str]):
+    def filter_components(
+        self,
+        names: Iterable[str],
+        filter_: Optional[Callable[[PhysicalComponent], bool]] = None,
```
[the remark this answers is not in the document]
je-cook: Because filter is a builtin: https://docs.python.org/3/library/functions.html?highlight=isinstance#filter its also the same variable name that anytree uses

bluemira/base/reactor.py
je-cook: yeah it does isinstance accepts types or tuples of types (or Union's of types too by the looks of it) https://docs.python.org/3/library/functions.html?highlight=isinstance#isinstance

bluemira/base/components.py
```diff
@@ -99,6 +103,8 @@ def filter_components(self, names: Iterable[str]):
         ----------
         names:
             The list of names of each component to search for.
+        filter_:
+            A callable to filter PhysicalComponents from the Component tree
```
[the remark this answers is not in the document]
je-cook: Sure I'll improve the docstring

bluemira/base/reactor.py
je-cook: I suppose, less fun though :smile:

bluemira/base/reactor.py
```diff
@@ -139,17 +153,76 @@ def _filter_tree(self, comp: Component, dims_to_show: Tuple[str, ...]) -> Compon
...
+        not_material: Union[
+            Type[SerialisedMaterial], Tuple[Type[SerialisedMaterial]], None
+        ] = Void,
+    ):
+        super().__setattr__("material", material)
+        super().__setattr__("not_material", not_material)
+
+    def __call__(self, node: anytree.Node) -> bool:
+        """Filter node based on material include and exclude rules"""
+        return hasattr(node, "material") and self._filterer(node.material)
```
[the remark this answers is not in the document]
je-cook: Improved

bluemira/base/reactor.py
oliverfunk: Haha right, gotta love ye ol python.

hsaunders1904 (approved):

PR #2238, opened 2023-05-11
Port over `RippleConstrainedLengthGOP` in new optimisation feature branch

### LINKED ISSUE #2237: Update `RippleConstrainedLengthGOP` for new optimisation interface

CoronelBuendia:
## Linked Issues

Closes #2237

## Checklist


bluemira/builders/tf_coils.py
```diff
@@ -598,7 +450,9 @@ class RippleConstrainedLengthGOP(GeometryOptimisationProblem):
     def __init__(
         self,
         parameterisation: GeometryParameterisation,
-        optimiser: Optimiser,
+        algorithm: str,
+        opt_conditions: Dict[str, float],
+        opt_parameters: Dict[str, float],
```
CoronelBuendia, line 455: Wee bit clunky methinks, especially as opt_parameters is usually {} which we can't have as a default argument, and would need to separate out to be below other non default arguments.

CoronelBuendia:
This does most of the legwork prior to some `OptimisationProblem`-like interface enabling smoother configuration. A couple of outstanding points:
* Print statements will get called when estimating gradients (easily fixed with extra lines of code for df_constraint)
* Gradient differentiation args cannot be passed
* Have not fixed EU-DEMO yet
Interface likely to change once the ABC-lke stuff is in

CoronelBuendia:
@hsaunders1904 a couple of notes for you, following the inclusion of `GeomOptimisationProblem`. These are mostly missing "hooks":

* I cannot pass a custom `x0` into `.optimise()`
* There are two discretisations for a KOZ constraint as presently implemented (one for the shape, the other for the KOZ). It would be nice to be able to set these independently. The present interface only exposes 1.
* I note that using the KOZ method to implement KOZ constraints enforces a particular approach for discretisation (`byedges=True`, no `dl`). This is probably not a big deal, as I could always write my own custom KOZ constraint. I may need to end up doing this anyway for the TF coil regression test as there I used some dl value.
* I cannot pass the tolerance(s) for (a) KOZ constraint(s) through the `GeomOptimisationProblem`. The default is used automatically.

Once these are done, I can reshuffle the TF coil problem as implemented in EU-DEMO and the json config and do a proper regression test.

bluemira/builders/tf_coils.py
```diff
@@ -641,44 +498,54 @@ def __init__(
...
+    def ineq_constraints(self):
+        """
+        Inequality constraints
+        """
+        return [self._ripple_constraint]
+
     def optimise(self, x0: Optional[np.ndarray] = None) -> GeometryParameterisation:
         """
         Solve the GeometryOptimisationProblem.
         """
```
CoronelBuendia, line 534: I still need to muck around with this.

hsaunders1904:
>     * I cannot pass a custom `x0` into `.optimise()`

My intention was that you pass in the geometry parameterisation, not the raw `x0`. Is there a use-case I'm missing for why you would want to pass in the raw parameterisation array rather than the geometry parameterisation?

I've actually been toying with the idea of removing the `x` from the `GeomOptimiserResult`. Because `x` is normalised using the parameterisation's bounds, as soon as you separate `x` from the bounds, it  becomes pretty meaningless.

>     * There are two discretisations for a KOZ constraint as presently implemented (one for the shape, the other for the KOZ). It would be nice to be able to set these independently. The present interface only exposes 1.

Yep, that's fair. I'll have to think of the best interface for this.

>     * I note that using the KOZ method to implement KOZ constraints enforces a particular approach for discretisation (`byedges=True`, no `dl`). This is probably not a big deal, as I could always write my own custom KOZ constraint. I may need to end up doing this anyway for the TF coil regression test as there I used some dl value.

Again, fair. Maybe a `koz_dl` argument that overrides `koz_discretisation` if given?

```python
def optimise_geometry(
    geom: _GeomT,
    f_objective: GeomOptimiserObjective,
    df_objective: Optional[GeomOptimiserCallable] = None,
    *,
    ...
    koz_discretisation: Union[int, Iterable[int]] = 100,
    koz_dl: Optional[Union[int, Iterable[int]]] = None,
    koz_byedges: Union[bool, Iterable[bool]] = True,
) -> GeomOptimiserResult[_GeomT]:
   """
    Parameters
    ----------
    ...
    koz_dl: 
        The target discretisation length of the given keep-out zones.
        Overrides ``koz_discretisation` if given.
    koz_byedges:
        Whether to discretise the keep-out zones by edges.
    """
    ...

```

>     * I cannot pass the tolerance(s) for (a) KOZ constraint(s) through the `GeomOptimisationProblem`. The default is used automatically.

Ah, yep. `make_keep_out_zone_constraint` does have the `tol` argument, I've just forgotten to wire it up.

hsaunders1904:
@CoronelBuendia 

> hsaunders1904 a couple of notes for you, following the inclusion of `GeomOptimisationProblem`. These are mostly missing "hooks":
> 
>     * I cannot pass a custom `x0` into `.optimise()`
> 
>     * There are two discretisations for a KOZ constraint as presently implemented (one for the shape, the other for the KOZ). It would be nice to be able to set these independently. The present interface only exposes 1.
> 
>     * I note that using the KOZ method to implement KOZ constraints enforces a particular approach for discretisation (`byedges=True`, no `dl`). This is probably not a big deal, as I could always write my own custom KOZ constraint. I may need to end up doing this anyway for the TF coil regression test as there I used some dl value.
> 
>     * I cannot pass the tolerance(s) for (a) KOZ constraint(s) through the `GeomOptimisationProblem`. The default is used automatically.


https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/2331 should deal with these (apart from the first point, which I have mentioned in my previous comment). I see that it might also be useful to be able to customise the tolerance/discretisation from a build config. I'll think about that in https://github.com/Fusion-Power-Plant-Framework/bluemira/issues/2332.

bluemira/builders/tf_coils.py
```diff
@@ -641,44 +508,53 @@ def __init__(
...
-            )
-            constraints.append(koz_constraint)
-
-        super().__init__(parameterisation, optimiser, objective, constraints)
-
     def _make_koz_points(self, keep_out_zone: BluemiraWire) -> Coordinates:
         """
         Make a set of points at which to evaluate the KOZ constraint
         """
+        # TODO: I guess I'm gonna lose this fine control?
```
hsaunders1904, line 520: Is this still an issue? I think the new keep-out zone class has these options.

The keep-out zone stuff is really just for convenience, if finer control is needed, then we can always just write up an inequality constraint.

bluemira/builders/tf_coils.py
hsaunders1904: Is this still an issue? I'm not sure that I see the problem with these having overridable defaults

bluemira/builders/tf_coils.py
```diff
@@ -641,44 +508,53 @@ def __init__(
...
         """
-        parameterisation = super().optimise(x0=x0)
-        self.solver.update_cage(parameterisation.create_shape())
+        result = super().optimise(
+            self.parameterisation,
+            algorithm=self.algorithm,
+            opt_conditions=self.opt_conditions,
+            opt_parameters=self.opt_parameters,
+        )
+        self.parameterisation.variables.set_values_from_norm(result.x)
```
hsaunders1904, line 551: Just so you're aware, there's some debate over whether we want to optimise geometries in-place, or return an optimised copy*. At the moment, we unfortunately can't copy an `OptVariables` instance (therefore we can't copy a geometry parameterisation, but Ollie is doing some work toward fixing this), so we're always editing in place. Which means, for the time being, I don't think this call is doing anything. 

I've also been thinking about whether having the `x` in the geometry optimisation result makes sense. Separated from their bounds, the normalised values don't really mean anything. The key output of the geometry optimisation was really intended to be the `.geom` attribute of the result.

With the above considered, I think the most future-proof way of doing this would be:
```suggestion
        self.parameterisation = super().optimise(
            self.parameterisation,
            algorithm=self.algorithm,
            opt_conditions=self.opt_conditions,
            opt_parameters=self.opt_parameters,
        ).geom
```
\
\
\* My preference is to take a copy. When we pass `x0` into a normal optimisation, we don't expect it to be updated in place, so why should we expect the geometry to be update in place?

bluemira/builders/tf_coils.py
```diff
@@ -203,99 +196,38 @@ def set_wire(self, wire: BluemiraWire):
 
     def make_ripple_constraint(
         self, parameterisation, solver, TF_ripple_limit, rip_con_tol
-    ) -> OptimisationConstraint:
+    ) -> Dict:
```
hsaunders1904, line 199:
```suggestion
    ) -> GeomConstraintT:
```

hsaunders1904 (approved):
Last little thing. I haven't run it, but the code looks good to me. There are a couple of outstanding TODOs (which I think are the same but in two different places?), are they to be resolved in this PR?

CoronelBuendia:
> Last little thing. I haven't run it, but the code looks good to me. There are a couple of outstanding TODOs (which I think are the same but in two different places?), are they to be resolved in this PR?

Yeah I think I'll leave them in, it's just a bit noisy in debug, but it's all flushed anyway, so not that big a deal

PR #2251, opened 2023-05-15
Add OptimisationProblem to optimisation module

hsaunders1904:
## Linked Issues

Nope

## Description

Adds an interface to the new `optimisation` module for defining an optimisation problem.

This just contains the interface for a vanilla optimisation, the interface for a geometry optimisation problem will be written in a different PR, but should be very similar.

## Checklist


hsaunders1904:
Converting to draft as there's a bug in how we're handling `df_objective` 😞

CoronelBuendia (approved):
I don't have time to comment on implementation details, I'm afraid.

I _think_ I'm happy with this from an API perspective. I can imagine sub-classing an `OptimisationProblem` and selecting it by name in a JSON file as we do in the EU-DEMO reactor example. The `OptimisationProblem` can take any __init__ so any parameters or biuld_config or tweaking parameters can go in there and again be specified in the JSON.

The one uhm-and-ahm I had was about the optimiser and its parameters which would (as done here) need to specified separately to the optimisation problem, as opposed to being set in the instantiation. I have decided I am happy with that, because it is a sensible separation. It just means probably a little more config sections in the JSON.

We can handle anything that looks too unseemly as we port over the rest of the problems in the optimisation PR, but crucially this should involve any config repercussions in the EU-DEMO config JSONs.

Of course, one could also override the `optimise` method so that it just takes x0 and the optimiser stuff from instantiation, so this base class is flexible enough to handle both.

bluemira/optimisation/problem.py
```diff
@@ -0,0 +1,123 @@
...
+        """The equality constraints on the optimisation."""
+        return []
+
+    def ineq_constraints(self) -> List[ConstraintT]:
+        """The inequality constraints on the optimisation."""
+        return []
+
+    def bounds(self) -> Optional[Tuple[np.ndarray, np.ndarray]]:
+        """The lower and upper bounds of the optimisation parameters."""
+        return None
```
CoronelBuendia, line 66: Should this not be an abstract method? Or can we mention what None does in terms of bounds? Does it default to -np.infty, np.infty` for example?

bluemira/optimisation/problem.py
```diff
@@ -0,0 +1,123 @@
...
+            opt_conditions=opt_conditions,
+            opt_parameters=opt_parameters,
+            bounds=self.bounds(),
+            eq_constraints=self.eq_constraints(),
+            ineq_constraints=self.ineq_constraints(),
+            keep_history=keep_history,
+        )
+
+    __T1 = TypeVar("__T1", bound=Callable[[Any], Any])
+    __T2 = TypeVar("__T2")
```
CoronelBuendia, line 96: I don't know how this renders in IDEs, but is it not possible to have slightly more descriptive shorthand here?
Also I don't understand what the TypeVar("__T2") does in practice, but in terms of legibility I am none the wiser :D

CoronelBuendia (changes requested):
OK I took a little look at the implementation, minor take it or leave its. Needs a rebase.

bluemira/optimisation/problem.py
hsaunders1904: Yep, better names are a good idea, I'll update them.

`TypeVar` instances basically represent any type (they can also be constrained to specific types, or derivatives of types, by using, e.g., `bound`). So in the case below

```python
T = TypeVar("T")
def f(x: T) -> T:
    ...
```

we're telling the type checker that `f` can take any type as input, and it returns an object of the same type.

bluemira/optimisation/problem.py
hsaunders1904: I'll clarify - just trying to clean up how we handle bounds to make this easier. If we allow scalar bound inputs, we can expand those once we know the dimensionality of the problem, and we can be explicit here and return `(-np.inf, np.inf)`.

je-cook (approved):

Issue #751, opened 2022-02-25
Move from fenics to fenicsX


je-cook:
## Description of issue / requirement to address

At some point in the future we need to move to fenicsX as fenics is unmaintained and had its last update in 2019

## Proposed solution

Update to fenicsX https://fenicsproject.org

## Alternative solutions

N/A

je-cook:
Here is the install script for fenicsx linked from #701 

https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/4c42b1c5b874c29f092c77e4a8dd9169290dc66b/scripts/fenicsx/install-fenicsx-deps.sh

https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/4c42b1c5b874c29f092c77e4a8dd9169290dc66b/scripts/fenicsx/install-fenicsx.sh

fyi @ivanmaione this will overwrite fencis-ufl everything else could be installed side by side

je-cook:
Accomplished in #2872 will go into develop with the python3.10 upgrade

PR #2254, opened 2023-05-16
Docker file with Freecad installed using apt-get

oliverfunk:
## Linked Issues

Closes #{ID}

## Checklist


oliverfunk:
This is partially waiting on #751 and I think should include changes that make would Bluemira packageable & publish'able to pypi, removing dependence on conda etc. and changes to install instructions.
Although maybe another PR would be needed for all those changes.

bluemira/__init__.py
```diff
@@ -23,6 +23,16 @@
...
+import sys
+
 from . import _version  # noqa: ABS101
 
+# this a change
+if os.environ.get("BLUEMIRA_ENV", "") == "CONTAINER":
+    sys.path.append("/usr/lib/freecad/Ext")
+    sys.path.append("/usr/lib/freecad/Mod/Part")
+    sys.path.append("/usr/lib/freecad/Mod/Draft")
+    sys.path.append("/usr/lib/python3/dist-packages")
```
je-cook, line 36: This last line feels bad for when we generalise this. Is it possible to only install python packages in the user venv or is something we need installed here?

oliverfunk: It is pretty terrible and I meant to leave a comment on top the line.

But it's not as bad as it seems because ur appending it last, any packages on the path prior (say numpy) will come from the vevn you're using if they exist on in both i.e. because it's last it has lowest import priority.

Apt-install'ing Freecad puts the packages it needs (PySide and friends) in the global dist-packages. I've tried to make it put them in active venv but it won't work.

I thought it might be possible to copy the packages from the global dist-packages to the venv but I thought it probably wouldn't work because of paths and maybe other files, but I could be wrong.
I didn't try and just did this instead.

I think this is fine for a container but another solution will be needed if we want bluemira to be packageable and link against an installed Freecad.
It might be possible to include only the needed packages in the path (i.e. append PySide etc.) from the global, but also haven't tried.

je-cook: Fair enough! For now a comment will be enough for me. 

It would be nice in future if we only need to add specific packages to the system path, sounds like the best option given other constraints (if its possible).

je-cook:
As mentioned in #2792 Freecad through apt is not a direction we need to explore any more. Docker will probably still be a direction but it will be rewritten most likely

PR #2258, opened 2023-05-16
Reactor Config: add file reading feature

oliverfunk:
## Linked Issues

Based on #2187
Closes #2187

## Description

Allows values in config files to be paths to nested files.

## Checklist

### DIFF
--- a/bluemira/base/reactor_config.py
+++ b/bluemira/base/reactor_config.py
@@ -262,6 +262,7 @@ class ReactorConfig:
         # this routine is designed not to copy any dict's while parsing
 
         current_layer = self.config_data
+
         for next_idx, current_arg_key in enumerate(arg_keys, start=1):
             current_layer = current_layer.get(current_arg_key, {})
             next_arg_key = arg_keys[next_idx] if next_idx < len(arg_keys) else None
--- a/bluemira/base/reactor_config.py
+++ b/bluemira/base/reactor_config.py
@@ -195,7 +195,7 @@ class ReactorConfig:
 
     @staticmethod
     def _read_or_return(config_path: Union[str, Path, dict]) -> Dict:
-        if isinstance(config_path, str) or isinstance(config_path, Path):
+        if isinstance(config_path, (str, Path)):
             return ReactorConfig._read_json_file(config_path)
         elif isinstance(config_path, dict):
             return config_path
--- a/bluemira/base/reactor_config.py
+++ b/bluemira/base/reactor_config.py
@@ -236,20 +236,22 @@ class ReactorConfig:
         # remove _FILEPATH_PREFIX
         f_path = value[len(_FILEPATH_PREFIX) :]
 
-        # replace . with a path relative to the config file
-        if f_path.startswith("."):
+        # check if rel path
+        if not f_path.startswith("/"):
             if not isinstance(self._config_path, str):
                 raise ReactorConfigError(
                     "Can only use relative paths with a path config"
                 )
             # handles removing the first . and the /
             f_path = Path(self._config_path).parent / f_path
+        else:
+            f_path = Path(f_path)
 
         # check if file exists
-        if not Path(f_path).is_file():
+        if not f_path.is_file():
             raise FileNotFoundError(f"Cannot find file {f_path}")
 
-        f_data = self._read_json_file(f_path)
+        f_data = self._read_json_file(f_path.as_posix())
 
         return f_data
 
--- a/bluemira/base/reactor_config.py
+++ b/bluemira/base/reactor_config.py
@@ -101,7 +101,7 @@ class ReactorConfig:
         self.warn_on_empty_config = warn_on_empty_config
 
         self._config_path = config_path
-        self.config_data = ReactorConfig._read_or_return(config_path)
+        self.config_data = self._read_or_return(config_path)
         self.global_params = make_parameter_frame(
             self._extract_file_data_if_needed(self.config_data.get(_PARAMETERS_KEY, {})),
             global_params_type,
--- a/bluemira/base/reactor_config.py
+++ b/bluemira/base/reactor_config.py
@@ -300,7 +300,7 @@ class ReactorConfig:
                 raise ReactorConfigError(
                     f"Arg {current_arg_key} is too specific, "
                     "it must either be another JSON object "
-                    "or a path a JSON file."
+                    "or a path to a JSON file."
                 )
 
             # add all keys not in extracted already
--- a/bluemira/base/reactor_config.py
+++ b/bluemira/base/reactor_config.py
@@ -3,6 +3,7 @@
 import json
 import pprint
 from dataclasses import dataclass
+from pathlib import Path
 from typing import Dict, Iterable, Optional, Tuple, Type, TypeVar, Union
 
 from bluemira.base.error import ReactorConfigError
@@ -98,6 +99,7 @@ class ReactorConfig:
         self.warn_on_empty_local_params = warn_on_empty_local_params
         self.warn_on_empty_config = warn_on_empty_config
 
+        self._config_path = config_path
         self.config_data = ReactorConfig._read_or_return(config_path)
 
         self.global_params = make_parameter_frame(
@@ -229,6 +231,31 @@ class ReactorConfig:
             if not isinstance(a, str):
                 raise ReactorConfigError("args must be strings")
 
+    def _get_nested_filepaths(
+        self,
+        current_layer: str,
+        current_arg_key: str,
+        arg_keys: Tuple[str],
+        next_idx: int,
+    ) -> Dict:
+        if not Path(current_layer).resolve().is_file():
+            if isinstance(self._config_path, str):
+                current_layer = str(Path(self._config_path).parent / current_layer)
+            if not Path(current_layer).resolve().is_file():
+                raise FileNotFoundError(f"Cannot find file {current_layer}")
+
+        new_layer = self._read_json_file(current_layer)
+
+        # Dont need to load files repeatedly so put it back in config_data
+        if next_idx == 1:
+            self.config_data[current_arg_key] = new_layer
+        else:
+            tmp_dict = self.config_data
+            for key in arg_keys[: next_idx - 1]:
+                tmp_dict = tmp_dict[key]
+            tmp_dict[current_arg_key] = new_layer
+        return new_layer
+
     def _extract(self, arg_keys: Tuple[str], is_config: bool) -> dict:
         extracted = {}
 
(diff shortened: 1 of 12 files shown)


bluemira/base/reactor_config.py
```diff
@@ -231,30 +227,31 @@ def _check_args_are_strings(self, args: Iterable[str]):
...
+            if not isinstance(self._config_path, str):
+                raise ReactorConfigError(
+                    "Can only use relative paths with a path config"
+                )
+            # handles removing the first . and the /
+            f_path = Path(self._config_path).parent / f_path
+
+        # check if file exists
+        if not Path(f_path).is_file():
+            raise FileNotFoundError(f"Cannot find file {f_path}")
```
je-cook, line 250: why not just use `Path(path).resolve()`  and then `.is_file()` check if config_path is not set and raise if `is_file()` is false?  This allows for a little bit more flexibility and offloads the relative path checking to `Path`
```suggestion
        f_path = Path(f_path).resolve()
        if not f_path.is_file():
            if not isinstance(self._config_path, str):
                raise ReactorConfigError(
                    "Can only use relative paths with a path config"
                )
            # handles removing the first . and the /
            f_path = Path(self._config_path).parent / f_path
            
            if not f_path.is_file():
                raise FileNotFoundError(f"Cannot find file {f_path}")
```

oliverfunk: You wouldn't want to check if f_path is a file until you've constructed the full (intended) path.
In your suggestion, after the call to .resolve(), that could point to a file in the bluemira dir. which is what's intended by the relative path.

oliverfunk: I agree though that the way I wrote it is a bit weird with f_path being either a `str` or `Path` when it's used again in the `not Path(f_path).is_file()`
So will change

hsaunders1904:
Copying comments on superseded PR here for visibility:

https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/2187/files/312bc0986a0b8c53f1f29b8264cf48e5b417f643#diff-adf0e08775445197a301be6f9e8974e2d9e324634119d05c117f04fc4acd9550
> Thinking about this API a bit, I wonder if this is the best way. Even with the `$` prefix, it doesn't sit very easy.
> 
> The schema I'd pictured looked more like:
> 
> ```json
> {
>   "Tester": {"params": {"$file": "reactor_params.global.json"}}
> }
> ```
> 
> To me, it feels more conceptually straight-forward to have special _keys_, rather than special _values_. It also means we don't have to worry about type checking and then checking the prefix of every single value. Though, I guess we'll still have to type-check every value for if it's a dictionary and contains `"$file"`.
> 
> I think it also makes sense that you can't specify a separate JSON file outside a key-value pair section. What happens if a value is expected to be a non-dictionary and someone gives a file path (even assuming we have the `$` prefix from Oliver's PR)? It doesn't really make sense to load that value from another JSON file, which would, at a minimum, be a single key-value pair.
> 
> Doing this also opens up the possibility of other 'special keys' (can't think what they'd be right now, but you never know 😅).
> 
> This would also allow us to split options between the parent and child file. We could imagine a case like the one below, where we expand/override the local with the global. Maybe this overcomplicates things, but it feels intuitive.
> 
> ```json
> {
>   "Tester": {
>     "params": {
>       "$file": "reactor_params.global.json",
>       "R_0": {
>         "value": 8,
>         "unit": "m"
>       }
>     }
>   }
> }
> ```
> 
> Obviously you two have been working on this, so have a better understanding of the issues than me, but these are my initial thoughts.

See also: https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/2187/files/312bc0986a0b8c53f1f29b8264cf48e5b417f643#diff-bf96bfe35aba6f2a58855b050c91c92e1a8ffdd971f4c660317211ca6d0454e3 on recursive reading.

je-cook (approved):
Looks good to me, @hsaunders1904 I'll leave it in your hands.

bluemira/base/reactor_config.py
```diff
@@ -201,8 +199,7 @@ def _read_or_return(dict_or_str_path: Union[str, dict]) -> Dict:
             return ReactorConfig._read_json_file(dict_or_str_path)
         elif isinstance(dict_or_str_path, dict):
             return dict_or_str_path
-        else:
-            raise ReactorConfigError("Invalid config_path")
+        raise ReactorConfigError("Invalid config_path")
```
hsaunders1904, line 202: It may be nice to support `pathlib.Path` objects here.

Also, this error message is a bit terse. It doesn't really give me any clues about why the path is invalid and what is an acceptable `config_path`.

bluemira/base/reactor_config.py
```diff
@@ -229,12 +226,51 @@ def _check_args_are_strings(self, args: Iterable[str]):
             if not isinstance(a, str):
                 raise ReactorConfigError("args must be strings")
 
+    def _expand_paths_in_dict(self, d: Dict[str, Any], rel_path: Path):
+        """
+        Expand all file paths in the dict
+        by replacing their values with the json file contents.
```
hsaunders1904, line 232: The docstring summary should be on one line. This could be shortened a bit so it fits:

`Expand all file paths by replacing their values with the json file's contents.`

bluemira/base/reactor_config.py
```diff
@@ -229,12 +226,51 @@ def _check_args_are_strings(self, args: Iterable[str]):
...
+        if not f_path.startswith("/"):
+            f_path = rel_path / f_path
+        else:
+            f_path = Path(f_path)
+
+        # check if file exists
+        if not f_path.is_file():
+            raise FileNotFoundError(f"Cannot find file {f_path}")
+
+        f_data = self._read_json_file(f_path.as_posix())
```
hsaunders1904, line 264: Not sure that we need the `as_posix` here, `open` can handle `pathlib.Path` objects - although maybe you were just trying to keep the type-checker happy with what `_read_json_file` expects. Fine either way, but thought it was worth mentioning.

bluemira/base/reactor_config.py
```diff
@@ -245,6 +281,13 @@ def _extract(self, arg_keys: Tuple[str], is_config: bool) -> dict:
                 # get the values from the _PARAMETERS_KEY
                 to_extract = current_layer.get(_PARAMETERS_KEY, {})
 
+            if not isinstance(to_extract, dict):
+                raise ReactorConfigError(
+                    f"Arg ${current_arg_key} is too specific, "
```
hsaunders1904, line 286: I'm not sure what '`too specfic`' means, here. Probably just me not following the code very well, but in what cases would I expect to see this error? Also, there's no test for this error case

bluemira/base/reactor_config.py
```diff
@@ -229,12 +226,51 @@ def _check_args_are_strings(self, args: Iterable[str]):
...
+    ) -> Tuple[Union[Any, dict], str]:
+        if not isinstance(value, str):
+            return value, rel_path
+        if not value.startswith(_FILEPATH_PREFIX):
+            return value, rel_path
+
+        # remove _FILEPATH_PREFIX
+        f_path = value[len(_FILEPATH_PREFIX) :]
+
+        # check if rel path
```
hsaunders1904, line 254: Nitpicky, but it might add some value to this comment to say what the path is relative to. I guess relative to the directory of the config file that the path appears in? Putting myself in the shoes of someone who's trying to work out what the code is doing with no prior knowledge, I guess it's just not 100% clear here what the `rel_path` variable actually is.

hsaunders1904 (approved):
Worked like a charm in my testing 👍. 

A few comments, but all pretty minor

bluemira/base/reactor_config.py
oliverfunk: Ah okay, good catch.

I thought the below few lines explained it more.
It's saying if you do something like config_for('a', 'b', 'c'), the value of 'c' must be either a path or a dict. If it's a literal value, you're probably doing something wrong.

It would like doing config_for('IVC', 'Wall silhouette', 'run_mode')
which is wrong.
It should be config_for('IVC', 'Wall silhouette')['run_mode'] if you wanted just 'run_mode'

bluemira/base/reactor_config.py
oliverfunk: Ya def don't need it!

bluemira/base/reactor_config.py
```diff
@@ -194,15 +194,17 @@ def config_for(self, component_name: str, *args: str) -> dict:
...
-        elif isinstance(dict_or_str_path, dict):
-            return dict_or_str_path
-        raise ReactorConfigError("Invalid config_path")
+    def _read_or_return(config_path: Union[str, Path, dict]) -> Dict:
+        if isinstance(config_path, str) or isinstance(config_path, Path):
+            return ReactorConfig._read_json_file(config_path)
+        elif isinstance(config_path, dict):
+            return config_path
+        raise ReactorConfigError(
+            "config_path must be either a dict, a Path object, or a string"
```
hsaunders1904, line 203: Sorry, another nitpicky one. I always think it makes ID-ing the problem easier if you print out the bad type.

```suggestion
            f"config_path must be either a dict, a Path object, or a string, found {type(config_path)}."
```

bluemira/base/reactor_config.py
```diff
@@ -283,9 +298,9 @@ def _extract(self, arg_keys: Tuple[str], is_config: bool) -> dict:
 
             if not isinstance(to_extract, dict):
                 raise ReactorConfigError(
-                    f"Arg ${current_arg_key} is too specific, "
-                    "it must be bound to a dict (or JSON object), "
-                    "or be a path to one."
+                    f"Arg {current_arg_key} is too specific, "
+                    "it must either be another JSON object "
+                    "or a path a JSON file."
```
hsaunders1904, line 303:
```suggestion
                    "or a path to a JSON file."
```

hsaunders1904 (approved):
Thanks for the changes 😄, a couple more very specific suggestions, happy to merge after that.

A really nice feature 🚀

bluemira/base/reactor_config.py
```diff
@@ -194,15 +194,17 @@ def config_for(self, component_name: str, *args: str) -> dict:
...
 
     @staticmethod
-    def _read_or_return(dict_or_str_path: Union[str, dict]) -> Dict:
-        if isinstance(dict_or_str_path, str):
-            return ReactorConfig._read_json_file(dict_or_str_path)
-        elif isinstance(dict_or_str_path, dict):
-            return dict_or_str_path
-        raise ReactorConfigError("Invalid config_path")
+    def _read_or_return(config_path: Union[str, Path, dict]) -> Dict:
+        if isinstance(config_path, str) or isinstance(config_path, Path):
```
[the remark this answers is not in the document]
oliverfunk: Nice

bluemira/base/reactor_config.py
oliverfunk: Meant to do that, nice catch

PR #2266, opened 2023-05-22
🐛 Fix collect verts and faces for polyscope viewing

je-cook:
## Description

When testing #2167 it would error with the polyscope backed because we were trying to stack empty arrays. 
This fixes that

my test line:
```python
from bluemira.base.reactor import FilterMaterial, Void

reactor.show_cad("xyz", n_sectors=3, filter_=FilterMaterial(keep_material=Void, reject_material=None), backend='polyscope')
```
## Checklist

### DIFF
--- a/bluemira/codes/_freecadapi.py
+++ b/bluemira/codes/_freecadapi.py
@@ -1997,7 +1997,7 @@ def collect_verts_faces(
         # tesselation is likely to be the most expensive part of this
         v, f = face.tessellate(tesselation)
 
-        if v != []:
+        if v != [] and f != []:
             verts.append(np.array(v))
             faces.append(np.array(f) + voffset)
             voffset += len(v)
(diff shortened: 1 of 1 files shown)


bluemira/codes/_freecadapi.py
```diff
@@ -1997,7 +1997,7 @@ def collect_verts_faces(
         # tesselation is likely to be the most expensive part of this
         v, f = face.tessellate(tesselation)
 
-        if v != []:
+        if v != [] and f != []:
             verts.append(np.array(v))
             faces.append(np.array(f) + voffset)
             voffset += len(v)
```
je-cook, line 2003: Another option, but I'm not sure we want faces that have vertices but not faces..
```suggestion
        if v != []:
            verts.append(np.array(v))
            if f != []:
                faces.append(np.array(f) + voffset)
            voffset += len(v)
```

je-cook: this could be more robust though

hsaunders1904: I don't think I know enough about how this works to comment, really. Maybe wires will have vertices but not faces? This looks sensible to me though

hsaunders1904 (approved):

PR #2290, opened 2023-05-31
🎨 Improve freecad saving, examples and tests

je-cook:
## Description

Fixes some FreeCAD saving methods, checks in the test for skip on dependency availability and removes some tests that will never pass.
Adds the new saving to the examples.

~TODO check for other occurances of old saving examples~ cant find any

## Interface Changes

None

## Checklist

### DIFF
--- a/bluemira/codes/_freecadapi.py
+++ b/bluemira/codes/_freecadapi.py
@@ -1415,7 +1415,7 @@ def save_cad(
         elif formatt is CADFileType.AUTOCAD:
             mesg += " FreeCAD requires `LibreDWG` to save in this format."
 
-        raise FileNotFoundError(
+        raise FreeCADError(
             f"{mesg} Not able to save object with format: '{formatt.value}'"
         )
 
--- a/bluemira/codes/_freecadapi.py
+++ b/bluemira/codes/_freecadapi.py
@@ -1193,17 +1193,17 @@ class CADFileType(enum.Enum):
     """
 
     # Commented out currently don't function
-    ACSII_STEREO_MESH = ("ast", "Mesh")
+    ASCII_STEREO_MESH = ("ast", "Mesh")
     ADDITIVE_MANUFACTURING = ("amf", "Mesh")
-    # ASC = ("asc", "Points")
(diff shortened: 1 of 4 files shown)


tests/_helpers.py
```diff
@@ -74,3 +76,16 @@ def new_isfile(path: str) -> bool:
...
+    """Create skipif marker for unimportable modules"""
+    for m in module_name:
+        try:
+            __import__(m)
+            skip = False
+        except ImportError:
+            skip = True
+            break
+
+    return pytest.mark.skipif(skip, reason=f"{module_name} dependency not found")
```
hsaunders1904, line 91: `module_name` is a tuple, is this what you meant to put in the message?

bluemira/codes/_freecadapi.py
```diff
@@ -1402,9 +1402,21 @@ def save_cad(
...
+                " FreeCAD requires `ifcopenshell` and"
+                " IFCJSON module to save in this format."
+            )
+        elif formatt is CADFileType.AUTOCAD:
+            mesg += " FreeCAD requires `LibreDWG` to save in this format."
+
         raise FileNotFoundError(
-            f"{filename} not created, filetype not written by FreeCAD."
-            f"Possibly no object compatible with '{formatt.value}'"
+            f"{mesg} Not able to save object with format: '{formatt.value}'"
```
hsaunders1904, line 1419: Is `FileNotFoundError` the right error to raise here? When writing, I would only really expect a `FileNotFoundError` if we were trying to write a file to a directory that doesn't exist.

Not really sure what the best error to raise would be though, maybe a `CADError`, if we have that or something similar?

hsaunders1904 (approved):
A couple of small things, otherwise looks fine 👍

tests/_helpers.py
je-cook: nope :)

bluemira/codes/_freecadapi.py
je-cook: We have `FreeCADError` just above, I've put that in.

PR #2291, opened 2023-05-31
♻️ Rearchitect plotting options and add dataclass descriptors

je-cook:
## Description

Rearchitects plotting options to use a dataclass. Also adds the user of descriptors for dataclasses instead of the properties hack (see https://docs.python.org/3/library/dataclasses.html#descriptor-typed-fields).

Also fixes quite a few bugs that were in the plotting example.

## Interface Changes

There are some slight API changes due to the dictionary -> dataclass conversion. Also to get the placement as a `BluemiraPlacement` you now need to call the `view_placement` property in the options. I see this as a small change as it isnt a regularly used user facing thing.

## Checklist


CoronelBuendia (approved):
Approved for `geometry` only.

bluemira/display/tools.py
```diff
@@ -0,0 +1,82 @@
...
+    def __init__(self, **kwargs):
+        self.modify(**kwargs)
+
+    def __setattr__(self, attr, val):
+        """
+        Set attributes in options dictionary
+        """
+        if (
+            hasattr(self, "_options")
+            and self._options is not None
```
hsaunders1904, line 45:
```suggestion
            getattr(self, "_options", None) is not None
```

bluemira/utilities/tools.py
```diff
@@ -302,6 +303,28 @@ def cross(
...
+        self._name = "_" + name
+
+    def __get__(self, obj: Any, _) -> str:
+        """Get the hex colour"""
+        if obj is None:
+            return self._default
+
+        return colors.to_hex(getattr(obj, self._name, self._default))
+
+    def __set__(self, obj: Any, value: Union[str, Tuple[float, ...]]):
```
hsaunders1904, line 323: Are we restricted to RGB? In which case, shouldn't this be `Tuple[float, float, float]`?

bluemira/display/plotter.py
```diff
@@ -50,224 +50,130 @@
...
+
+    def __set__(self, obj: Any, value: Union[str, _placement.BluemiraPlacement]):
+        """Set the view"""
+        if isinstance(value, str):
+            if value == "xy":
+                value = _placement.XYZ
+            elif value == "xz":
+                value = _placement.XZY
+            elif value == "yz":
+                value = _placement.YZX
```
hsaunders1904, line 88: Maybe I'm missing something obvious, but why are the strings here different from the keys in the `BluemiraPlacement` enum?

hsaunders1904 (commented):
All basically looks good to me. Only a couple of comments/questions

bluemira/display/plotter.py
```diff
@@ -50,224 +50,130 @@
...
-    """
-
-    _options = None
+    def __set_name__(self, _, name: str):
+        """Set the attribute name from a dataclass"""
+        self._name = "_" + name
+
+    def __get__(
+        self, obj: Any, _
+    ) -> Tuple[Iterable[float], Iterable[float], float, str]:
```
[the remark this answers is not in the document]
je-cook: I could do `Union[List[float], Tuple[float]]`? Its the same typing as here: https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/6071256c2c142551d05b0f7213778db6d7e3075d/bluemira/geometry/placement.py#L57

bluemira/display/plotter.py
je-cook: I'm not sure I follow. `_placement.XYZ` is a `BluemiraPlacement` defined in the `bluemira.geometry.placement` module

je-cook: for some reason the import previously was from `bluemira.geometry import placement as _placement` unsure why but I kept it

je-cook: discussed elsewhere

bluemira/utilities/tools.py
hsaunders1904: Is that potentially worth adding to the docstring?

bluemira/display/plotter.py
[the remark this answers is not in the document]
je-cook: Typing is now more specific for the return type

hsaunders1904 (approved):

Issue #2189, opened 2023-05-04
EPIC: Optimisation feature branch


hsaunders1904:
## Description of issue / requirement to address

List of tasks to address before `feature/new_optimisation_module` can be merged into develop.

- [x] Port existing optimisation routines
  - Normal optimisations:
    - [x] #2231
    - [x] #2383
    - [x] #2225
    - [x] #2221 
  - Geometry optimisations:
    - [x] #2237
    - [x] #2267
    - [x] #2233
- [x] #2190
- [x] #2203
- [x] #2227
- [x] #2256
- [x] #2274
- [x] #2335
- [x] #2252
- [x] #2337
- [x] #2340
- [x] Handle `ad_args` for automatic differentiation [sic] of objective and constraints
   - @hsaunders1904: I would argue if you wanted to customise this automatic differentiation, you should implement `df_objective`/`df_constraint` and use `approx_derivative`.
- [x] ~Tests for equality constraints~ (managed to throw some of these together when testing the optimisation problem class)
- [x] #2414
  - This could be a good opportunity to have a think about the visibility of the optimisation module's submodules, which are a bit of a mess of underscores at the moment.
- [x] Remove geometry constraint registry (also #2420) 
  - The constraints should live on the geometry parametrisations
- [x] #2416
  - This default could be based on which algorithm is used, which is how SciPy handles it.
- [x] #2415
- [x] #2417

## Non-blocking 
Issues that can be resolved after the feature branch has been merged into develop

- [x] #2332
- [x] Do not warn that maximum iterations reached when using ISRES (max iterations are usually expected to be reached) #2499 
  - Maybe just a debug or info message would suffice here
- [x] #3024
- [x] Add further defaults to the `AlgorithmDefaultTolerances` covered in #2447

PR #2331, opened 2023-06-12
Make keep-out zones more customisable

hsaunders1904:
## Linked Issues

Refs #2189 
Addresses issues flagged in https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/2238#issuecomment-1564046300

## Description

Expose more keep-out zone features through the `optimise_geometry` and `GeomOptimisationProblem` API.

This is done using a new class `KeepOutZone` which is constructed using a `BluemiraWire` and, optionally, discretization configuration and constraint tolerance.

## Checklist

### DIFF
--- a/bluemira/optimisation/_geometry/_tools.py
+++ b/bluemira/optimisation/_geometry/_tools.py
@@ -19,7 +19,8 @@
 # You should have received a copy of the GNU Lesser General Public
 # License along with bluemira; if not, see <https://www.gnu.org/licenses/>.
 import copy
-from typing import List
+from dataclasses import dataclass
+from typing import List, Optional
 
 import numpy as np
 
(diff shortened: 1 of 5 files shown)


bluemira/optimisation/_geometry/_tools.py
```diff
@@ -102,17 +116,13 @@ def calculate_signed_distance(
     return signed_distance_2D_polygon(s.T, zone_points.T).T
```
CoronelBuendia, line 116: Might be worth exposing the dl, byedges, ndiscr in the shape discretisation as well... Just thinking about a "stable" API here; with this discretisation control on the KOZ, we might also want the same level of control on the shape. If we can do that in the same way, API-wise, it would be good. Not sure if this means a different interface, but it could well do.

CoronelBuendia: Ah yes, that's true... It's likely only to be the case for very low ndiscr, or polygons with many many sides. But it's a potential problem yes. OK, anyway, this PR looks good but I'd like to have a play with it tomorrow and get back to you.

hsaunders1904: I've added the ability to set `n_shape_discr`

hsaunders1904:
Rebased

PR #2336, opened 2023-06-12
Fix returning initial value and not previous iteration on round-off error

### LINKED ISSUE #2335: Fix returning initial value and not previous iteration on round-off error

hsaunders1904:
## Linked Issues

Closes #2335 

## Description

On a round-off error, we were actually returning the _initial_ parameterisation, not the parameterisation from the previous iteration 😱.

I'm a little dubious of the way this was being handled in the old optimiser. I found it necessary to copy the NLOpt array to get the documented behaviour (at least what I thought the documented behaviour meant). This was not being done in the old optimiser classes.

I thought we were intending to return the parameterisation from the iteration _before_ we got the round-off error (this is what I've implemented). But it occurs to me now that maybe we wanted the iteration that actually caused the round-off error (I believe this is the behaviour the old optimiser has. Couldn't this mean we get a bunch of `NaN`s?, though).

@CoronelBuendia any thoughts on what behaviour we're actually after?

Also, thoughts on this [TODO](https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/2336/files#diff-7c6c17eb9e2800ef1b3be1e40569f1c8c9ab22c63f789c02eeda0e1744d48b72R234) would be appreciated.

## Checklist

### DIFF
--- a/bluemira/optimisation/_nlopt/optimiser.py
+++ b/bluemira/optimisation/_nlopt/optimiser.py
@@ -229,16 +229,22 @@ class NloptOptimiser(Optimiser):
             f_x = self._objective.f(x_star)
         except nlopt.RoundoffLimited:
             # It's likely that the last call was still a reasonably good solution.
-            bluemira_warn(
-                "optimisation: round-off error occurred, returning last optimisation "
-                "parameterisation."
-            )
+            round_off_msg = "optimisation: round-off error occurred"
             if self._objective.history:
+                # TODO(hsaunders1904): Is this really what we want to
+                # return? Can't the minimum have been something that
+                # wildly violates constraints?
+                bluemira_warn(
+                    f"{round_off_msg}. Returning best parameterisation found so far."
+                )
                 fx_values = np.array(self._objective.history).T[1]
-                f_x = np.min(fx_values)
                 arg_min_fx = np.argmin(fx_values)
+                f_x = fx_values[arg_min_fx]
                 x_star = self._objective.history[arg_min_fx][0]
             else:
+                bluemira_warn(
+                    f"{round_off_msg}. Returning last optimisation parameterisation."
+                )
                 x_star = self._objective.prev_iter
                 f_x = np.infty
 
--- a/bluemira/optimisation/_nlopt/optimiser.py
+++ b/bluemira/optimisation/_nlopt/optimiser.py
@@ -20,7 +20,7 @@
 # License along with bluemira; if not, see <https://www.gnu.org/licenses/>.
 from __future__ import annotations
 
-from typing import Any, Dict, List, Mapping, Optional, Union
+from typing import Any, Dict, List, Mapping, Optional, Tuple, Union
 
 import nlopt
 import numpy as np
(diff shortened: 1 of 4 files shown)


je-cook:
I dunno whether this is related but https://github.com/stevengj/nlopt/issues 513

hsaunders1904:
> I dunno whether this is related but https://github.com/stevengj/nlopt/issues 513

Ha! Looks like the exact same problem. We're having to take a copy on every iteration as well

bluemira/optimisation/_nlopt/optimiser.py
```diff
@@ -281,6 +270,33 @@ def set_upper_bounds(self, bounds: Union[np.ndarray, float]) -> None:
...
+        Handle a round-off error occurring in an optimisation.
+
+        It's likely the last call was a decent solution, so return that
+        (with a warning).
+        """
+        round_off_msg = "optimisation: round-off error occurred"
+        if self._objective.history:
+            # TODO(hsaunders1904): Is this really what we want to
+            # return? Can't the minimum have been something that
+            # wildly violates constraints?
```
CoronelBuendia, line 284: It's possible, I think, depending on the algorithm. But normally round-off errors return reasonable points.

Quoting the NLOpt docs here https://nlopt.readthedocs.io/en/latest/NLopt_Reference/#error-codes-negative-return-values:
```
NLOPT_ROUNDOFF_LIMITED = -4

Halted because roundoff errors limited progress. (In this case, the optimization still typically returns a useful result.)

```

CoronelBuendia: Sorry, that was not a good answer. Upon round-off, we return best point found so far, rather than the last. This is what we were doing before, correct? I guess I got paranoid... In my experience, round-off errors tend to occur if:
* you haven't normalised the input vector and have either very large or very small (or a mixture of both)
* you are more or less stepping around an optimum under certain conditions (e.g. near one or more constraints).
Perhaps it would be wiser to follow whatever NLopt does here, or check the best vector against the constraints.

hsaunders1904: OK, that makes sense. Is what we're doing okay here? 

The old optimiser stuff was just returning a reference to the parameterisation array, not the best result so far. I'm not convinced it was returning the parameterisation from the previous iteration, but rather returning the parameterisation that caused the round-off error (but maybe that's OK?). I'll need to double-check. At the moment, we're copying the parametrisation at every iteration, which is not ideal.

Busy with STEP stuff today and tomorrow, so probably won't get to this until next week.

hsaunders1904: It was found that the old optimiser was returning uninitialised memory on a round-off error.

It seems that we cannot get around making a copy of the parameterisation at every iteration if we want to return the last result on a round-off error. So that's what we do in the new optimiser.

<details>
  <summary>Round-off error snippet (thanks @CoronelBuendia)</summary>

  ```python
    class RoundOffErr:
      def __init__(self) -> None:
          self.it = 0
  
      def __call__(self, x, grad):
          """Objective function to create a round-off error."""
          self.it += 1
          value = -x[0] + x[1]
  
          if grad.size > 0:
              grad[0] = -1
              grad[1] = 1
  
          print(f"{self.it:02}: {x=}, {value=}")
          return value
  

  f_round_off = RoundOffErr()
  optimiser = Optimiser("COBYLA", 2, opt_conditions={"ftol_abs": 0.01})
  optimiser.set_objective_function(f_round_off)
  optimiser.set_lower_bounds([-3, -3])
  optimiser.set_upper_bounds([3, 3])
  result = optimiser.optimise([0.5, 2])
  print(f"{result=}")
  print(f"{optimiser.optimum_value=}")
  ```
</details>

hsaunders1904:
This in from an above comment thread, but I'm putting it here as well for posterity:

> It was found that the old optimiser was returning uninitialised memory on a round-off error.
> 
> It seems that we cannot get around making a copy of the parameterisation at every iteration if we want to return the last result on a round-off error. So that's what we do in the new optimiser.
> 
> ```python
> from bluemira.utilities.optimiser import Optimiser
> 
> 
> class RoundOffErr:
>     def __init__(self) -> None:
>         self.it = 0
> 
>     def __call__(self, x, grad):
>         """Objective function to create a round-off error."""
>         self.it += 1
>         value = -x[0] + x[1]
> 
>         if grad.size > 0:
>             grad[0] = -1
>             grad[1] = 1
> 
>         print(f"{self.it:02}: {x=}, {value=}")
>         return value
> 
> 
> f_round_off = RoundOffErr()
> optimiser = Optimiser("COBYLA", 2, opt_conditions={"ftol_abs": 0.01})
> optimiser.set_objective_function(f_round_off)
> optimiser.set_lower_bounds([-3, -3])
> optimiser.set_upper_bounds([3, 3])
> result = optimiser.optimise([0.5, 2])
> print(f"{result=}")
> print(f"{optimiser.optimum_value=}")
> ```

CoronelBuendia (approved):

PR #2366, opened 2023-06-21
Add `force_wire_to_spline`

CoronelBuendia:
## Linked Issues

Closes #{ID}

It's often the case that because of failures in primitive geometry operations we end up with large polygons which we subsequently have to deal with. These polygons are slow (particularly on boolean operations) and look hideous.

## Description

This PR adds a function to force these polygons to be splines.

## Interface Changes

Adds to `force_wire_to_spline` to `geometry/tools.py`
Removes an unnecessary warning from `interpolate_bspline` when `closed=True`

## Checklist


bluemira/geometry/tools.py
```diff
@@ -392,6 +392,68 @@ def interpolate_bspline(
...
+        bluemira_debug(
+            f"Wire already has {original_n_edges} < {n_edges_max=}. No point forcing to a spline."
+        )
+        return wire
+
+    original_points = wire.discretize(ndiscr=2 * original_n_edges, byedges=False)
+
+    n_discrs = np.array(original_n_edges * np.linspace(0.8, 0.1, 8), dtype=int)
+
+    for n_discr in n_discrs:
```
je-cook, line 435:
```suggestion
    for n_discr in np.array(original_n_edges * np.linspace(0.8, 0.1, 8), dtype=int):
```

tests/geometry/test_tools.py
```diff
@@ -846,3 +847,49 @@ def get_shared_fragments(group_1, group_2):
...
+        wires = []
+        with open(filename, "r") as file:
+            data = json.load(file)
+
+        for wire in data["BluemiraFace"]["boundary"]:
+            wires.append(deserialize_shape(wire))
+        cls.wires = wires
+
+    @mock.patch("bluemira.geometry.tools.bluemira_warn")
+    def test_force_spline(self, bm_warn):
```
je-cook, line 868: Its overkill here but you can also use the pytest caplog fixture for checking logging output instead of using the mock.  
`def test_force_spline(self, caplog):`

je-cook (approved):
Minors but nothing important. Looks good!

PR #2388, opened 2023-06-30
EU-DEMO lower port integration

CoronelBuendia:
## Linked Issues

Closes #{ID}

## Description

image

Shame about the construction lines, but it does the job.

## Checklist


CoronelBuendia:
@oliverfunk I'm about 80% of the way there with this, with most of the restructuring done. Feel free to take a look, I'm out all next week so no rush.

eudemo/eudemo/maintenance/lower_port/duct_designer.py
```diff
@@ -246,29 +249,20 @@ def _calc_y_point(x_point):
...
+        x = [ib_div_pt_x, ib_div_pt_x, ob_div_pt_x, ob_div_pt_x]
+        y = [ib_inner_y, -ib_inner_y, -ob_inner_y, ob_inner_y]
+        z = [ib_div_pt_z, ib_div_pt_z, ob_div_pt_z, ob_div_pt_z]
+        duct_inner_xy = make_polygon({"x": x, "y": y, "z": z}, closed=True)
+
+        # Translate a little inwards and upwards to ensure penetration to
+        # main body
+        angle = np.deg2rad(self.params.lower_port_angle.value)
+        direction = np.array([np.cos(angle), 0, np.sin(angle)])
+        duct_inner_xy.translate(-1 * direction)
```
oliverfunk, line 264: Not sure why you need to do this as this boundary is constructed around the divertor, which is definatly within the main body.
The angled part of the duct made from this boundary needs to be cut/merged into the blanket, which it defintly goes through (as it's started around the divertor)

oliverfunk (commented):
Just have one comment

eudemo/eudemo/maintenance/lower_port/duct_designer.py
CoronelBuendia: It just needs to overshoot a bit, otherwise boolean operations from faces which are offset from the original do not fully go through.

oliverfunk (approved):
Looks great!

There are some debug things left but they can be removed later.

PF coil placements working at extremes of -89 and 0 degrees.
However, it breaks at line 535 of duct_connections.py for a port angle of -90
I'm not sure if you want to fix that before merging (I couldn't comment on the line because it's not one that changed in the PR)

oliverfunk:
> > Looks great!
> > There are some debug things left but they can be removed later.
> > PF coil placements working at extremes of -89 and 0 degrees. However, it breaks at line 535 of duct_connections.py for a port angle of -90 I'm not sure if you want to fix that before merging (I couldn't comment on the line because it's not one that changed in the PR)
> 
> Oh right! OK good check. I will take a look. I was struggling with the -90 case to be honest, I fixed it in the test mostly by changing the test a little. It was working for -89.99 though. Headscratching ensued. I will take a look at that problem then.
> 
> How are you producing it? Setting lower_port_angle = -90 and running reactor.py?

Yea exactly.
It's probably a matter of checking if the list is empty and skipping some steps if it is, but yea handling the -90 case was generally quite a pain.

oliverfunk:
@CoronelBuendia fails for me on 0 degrees now. Working great for -90!

CoronelBuendia:
> @CoronelBuendia fails for me on 0 degrees now. Working great for -90!

xD shit, OK will have a look.

CoronelBuendia:
Alright so it should work now, but it ain't pretty. There's a couple of wider issues with that build function that could cause issues. I'd say we're at a stage where it's OK for now though.

oliverfunk:
@CoronelBuendia wd 🚀

PR #2410, opened 2023-07-12
EU-DEMO port plug integration

CoronelBuendia:
## Linked Issues

Closes #{ID}

## Description

Finally gitignores JSONs in EUDEMO config folder.

image
image
image

Lower port now included

## Checklist


eudemo/config/FirstWallDesign.json
je-cook, line 1: Need to do `git rm --cached FirstWallDesign.json` on this, PF and TF to keep the file but remove it from git tracking.

je-cook (approved):
Just one minor otherwise I'm (very) happy.

Finally back to a complete CAD!

bluemira/equilibria/run.py
```diff
@@ -719,7 +719,7 @@ def _consolidate_coilset(
         max_pf_currents = np.max(pf_currents, axis=0)
         # Relax the max currents a bit to avoid oscillation
         max_pf_current = self._eq_settings.peak_PF_current_factor * self.params.I_p.value
-        max_pf_currents = np.clip(1.05 * max_pf_currents, 0, max_pf_current)
+        max_pf_currents = np.clip(1.1 * max_pf_currents, 0, max_pf_current)
```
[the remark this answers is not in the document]
CoronelBuendia: Agreed... My thought was basically when I have some headspace I will find a better to specify this such that we dont have to do this.

The problem (for our future selves) is as follows:
* We are fixing the coil sizes and discretising simultaneously from an initial solution
* The discretisation affects the behaviour of the current for a given coil slightly
* Hard-prescribing the current for a new-size and discretised coil is "unfair"
* So we give it a bit of a current bound buffer (which we subsequently do not resize for)
This needs a better approach in general, but the problem will likely remain: How to solve two equilibria for a new coilset and correctly apply relaxation such that the coils are big enough in all cases.

Technically should include the breakdown too. To be continued.

eudemo/config/FirstWallDesign.json
CoronelBuendia: OK done

je-cook (approved):

Issue #2383, opened 2023-06-27
Update `CoilsetOptimisationProblem` for new optimisation interface


hsaunders1904:

hsaunders1904:
Closed by #2412

PR #2412, opened 2023-07-12
Retrofit coilset optimisations to use new optimiser interface

hsaunders1904:
## Linked Issues

Partially addresses #2383 

## Description

This needs a lot of cleaning up. I've been rushing through it a bit, trying to get things in (at least) working order. I'm expecting flake8 to tell me off.

The idea has basically been to copy across all the old classes in `equlilibria.opt_[problems|constraints|constraint_funcs|objectives]` into a new module: `equilibria.optimisation`. `CoilsetOptimisation` no longer inherits from `OptimisationProblem`, it's now just an abstract class where you need to define `optimise` which returns a `CoilsetOptimiserResult`. The approach I've taken is to make a call to `bluemira.optimisation.optimise` in the `CoilsetOptimisation.optimise` implementations.

The old coilset optimisation objectives have been re-written to match a new `ObjectiveFunction` interface. Which requires definition of `f_objective`, and optionally, `df_objective`. The coilset constraint function have been re-written to satisfy a new interface defined by `ConstraintFunction` (you must define `f_constraint` and optionally `df_constraint`). The `UpdateableConstraint` interface has been largely unchanged, but a `f_constraint` abstract method has been added, which must return a `ConstraintFunction`.

The old interface had a few inconsistencies that have been carried through here, like one of the base `CoilsetOptimisationProblem` methods assumes a `self.eq` exists, but the interface does not guarantee that, similar for `self.coilset`. We also need to clear up the constraint interface a little, so it's clear where `constraint_type` and `tolerance` should be defined.

I've been using `examples/equilibria/single_null.ex.py` and `examples/equilibria/eudemo_2017.ex.py` as test beds and get pretty much the exact same results. The `single_null` example seems to just do one extra iteration, could be the handling of the opt conditions, but I haven't had a chance to dig into that yet.

There are still some places using `equlilibria.opt_[problems|constraints|constraint_funcs|objectives]` that need to be dealt with, but it should all work similarly.

It could be worth splitting this branch into several PRs, one for each opt problem - might make things easier to review. Or just review one optimisation problem at a time - I've split them into separate files, so hopefully they're a bit more readable.

There's a real lack of tests in equilibria - could be worth fashioning one or two regression tests from the examples I mentioned above. I've run out of time a bit, I'm afraid. 

Note that the branch that contains the more ambitious refactor, with a more polished interface, is: [hsaunders1904/2383_update_coilsetoptimisationproblem_for_new_optimisation_interface](https://github.com/Fusion-Power-Plant-Framework/bluemira/tree/hsaunders1904/2383_update_coilsetoptimisationproblem_for_new_optimisation_interface). I was quickly running out of time, so abandoned this for the quicker refactor that is this PR.

## Checklist

### DIFF
--- a/bluemira/equilibria/solve.py
+++ b/bluemira/equilibria/solve.py
@@ -475,7 +475,7 @@ class PicardIterator:
 
     def _optimise_coilset(self):
         try:
-            coilset = self.opt_prob.optimise(fixed_coils=self.fixed_coils)
+            coilset = self.opt_prob.optimise(fixed_coils=self.fixed_coils).coilset
             self.store.append(coilset)
         except ExternalOptError:
             coilset = self.store[-1]
--- a/bluemira/equilibria/solve.py
+++ b/bluemira/equilibria/solve.py
@@ -475,7 +475,11 @@ class PicardIterator:
 
     def _optimise_coilset(self):
         try:
-            coilset = self.opt_prob.optimise(fixed_coils=self.fixed_coils).coilset
+            result = self.opt_prob.optimise(fixed_coils=self.fixed_coils)
+            try:
+                coilset = result.coilset
+            except AttributeError:
+                coilset = result
             self.store.append(coilset)
         except ExternalOptError:
             coilset = self.store[-1]
--- a/bluemira/equilibria/opt_problems.py
+++ b/bluemira/equilibria/opt_problems.py
@@ -296,25 +296,25 @@ class CoilsetPositionCOP(CoilsetOptimisationProblem):
 
     Parameters
     ----------
-    coilset: CoilSet
+    coilset:
         Coilset to optimise.
     eq: Equilibrium
         Equilibrium object used to update magnetic field targets.
-    targets: MagneticConstraintSet
+    targets:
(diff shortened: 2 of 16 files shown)


bluemira/equilibria/solve.py
```diff
@@ -475,7 +475,11 @@ def __init__(
 
     def _optimise_coilset(self):
         try:
-            coilset = self.opt_prob.optimise(fixed_coils=self.fixed_coils)
+            result = self.opt_prob.optimise(fixed_coils=self.fixed_coils)
+            try:
+                coilset = result.coilset
+            except AttributeError:
+                coilset = result
```
je-cook, line 482: when would result just be a coilset?

Is this layover from the leftover from the old stuff you mentioned?

je-cook (approved):
I'm happy to merge this as is into the feature branch and then do the flake cleanups in a separate PR.

Just the one quick question

bluemira/equilibria/solve.py
hsaunders1904: Yep, sorry, should have left a comment - this just keeps it compatible with the old stuff

PR #2418, opened 2023-07-13
Je cook/fix flake8

je-cook:
## Description

Fixes the flake8 issues on #2412 

## Checklist


bluemira/equilibria/optimisation/objectives.py
```diff
@@ -48,6 +48,8 @@ def f_objective(x, grad, args):
 
 
 class ObjectiveFunction(abc.ABC):
+    """Base class for ObjectiveFunctions"""
+
```
hsaunders1904, line 52: I think I made a little bit of a mistake here. We should remove `df_objective`, and specify that it can, optionally, be implemented. If it's not implemented, we use the numerical approximation.

You'll notice in the optimisation problems I do:

```python
optimise(
    f_objective=objective.f_objective,
    df_objective=getattr(objective, "df_objective", None),
    ...
)
```

If the base class implements `df_objective`, then we'll never end up using the approximate gradient.

While you're here, would you remove `df_objective` from the base class, and add to the docstring that it can be implemented in derived classes? If you can think of a better way to do this, feel free to change things.

je-cook: For now I've done as you suggested. I can't think of a nicer way right now either way :)

hsaunders1904 (approved):
Thanks for doing this, all looks good to me. Feel free to merge when you're ready

PR #2442, opened 2023-07-17
Convert SH constraint to new optimiser interface

je-cook:
## Linked Issues

Closes #{ID}

## Description

Converts the existing SH constraint so its using the new constraint mechanics

## Checklist

### DIFF
--- a/bluemira/equilibria/optimisation/constraints.py
+++ b/bluemira/equilibria/optimisation/constraints.py
@@ -32,6 +32,8 @@ if TYPE_CHECKING:
     from bluemira.equilibria.coils import CoilSet
     from bluemira.equilibria.equilibrium import Equilibrium
 
+import matplotlib.patches as patch
+import matplotlib.pyplot as plt
 import numpy as np
 import numpy.typing as npt
 
@@ -502,12 +504,14 @@ class SphericalHarmonicsConstraint(MagneticConstraint):
         """
         return len(self.ref_harmonics)
 
-    def plot(self, ax):
+    def plot(self, ax=None):
         """
         Plot the constraint onto an Axes.
         """
-        # Possibly the r_t semi circle?
-        raise NotImplementedError("TODO")
+        if ax is None:
+            _, ax = plt.subplots()
+
+        ax.add_patch(patch.Circle((0, 0), self.r_t, ec="orange", fill=True, fc="orange"))
 
 
 class AbsoluteMagneticConstraint(MagneticConstraint):
--- a/bluemira/equilibria/optimisation/constraints.py
+++ b/bluemira/equilibria/optimisation/constraints.py
@@ -458,7 +458,7 @@ class SphericalHarmonicsConstraint(MagneticConstraint):
     ----------
     r_t:
         Typical length scale of the problem (e.g. radius at outer midplane)
-    ref_harmonics:
+    ref_coil_psi_amplitudes:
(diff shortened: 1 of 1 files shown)


bluemira/equilibria/optimisation/constraints.py
```diff
@@ -445,6 +447,69 @@ def f_constraint(self) -> ConstraintFunction:
...
+        super().__init__(
+            target_value,
+            weights,
+            tolerance=tolerance,
+            f_constraint=f_constraint,
+            constraint_type=constraint_type,
+        )
+        self.max_degree = max_degree
+        self.r_t = r_t
+        self.ref_harmonics = ref_harmonics
```
oliverfunk, line 485: After to talking with Georgi, I think "harmonics" is a slightly misleading label. They are more amplitudes, or perhaps `ref_coil_psi_amplitudes``

As they are the SH amplitudes of psi w.r.t the current in the coils

bluemira/equilibria/optimisation/constraints.py
```diff
@@ -445,6 +447,69 @@ def f_constraint(self) -> ConstraintFunction:
...
+        The mathematical size of the constraint.
+        """
+        return len(self.ref_harmonics)
+
+    def plot(self, ax):
+        """
+        Plot the constraint onto an Axes.
+        """
+        # Possibly the r_t semi circle?
+        raise NotImplementedError("TODO")
```
oliverfunk, line 510: The only thing I could think of that would be possibly useful for this would be if you could plot the collocations points with their ref_amplitudes on the grid and maybe the coilset coil positions

oliverfunk (approved):
LGTM! I had to understand how it worked before I could review properly. Two small comments.

bluemira/equilibria/optimisation/constraints.py
oliverfunk: Just chatted with Georgi, it would be useful to plot the semi-circle (i.e. r_t) and the collocation points

je-cook: In its currrent form we don't have any information on the collocation points. Thats comes in before the max_degree is specified IIRC. For that information to be available I think the harmonic amplitudes and the max degree would need to be calculated in the constraint init. 
If thats a good direction and whats wanted it can be an improvement PR, probably outside the scope of this straight conversion.

I've added the r_t semi-circle

bluemira/equilibria/optimisation/constraints.py
```diff
@@ -445,6 +449,71 @@ def f_constraint(self) -> ConstraintFunction:
...
+        self.r_t = r_t
+        self.ref_coil_psi_amplitudes = ref_coil_psi_amplitudes
+
+    def control_response(self, coilset: CoilSet) -> npt.NDArray:
+        """
+        Calculate control response of a CoilSet to the constraint.
+        """
+        return coil_harmonic_amplitude_matrix(coilset, self.max_degree, self.r_t)
+
+    def evaluate(self, _: Equilibrium) -> npt.NDArray:
```
oliverfunk, line 495: Note: Check if the prepare method has a bug in it where it's setting b_vec = self.target_value - self.evaluate()

je-cook: for clarity this could be a bug that has existed for a while (before the optimisation refactor). Just needs checking

PR #2484, opened 2023-08-03
Improve robustness of FEM solver

je-cook:
## Description

While checking on the fem rework with the new optimiser I ran into the 'Jackpot' message. This is an attempt to avoid that as dividing by 0 and sqrt -ves gives Nans which will always hit the end of the loop.
Secondly there was a weird array masking deletion thing that I dont think worked as it was intended

## Checklist

### DIFF
--- a/bluemira/codes/plasmod/equilibrium_2d_coupling.py
+++ b/bluemira/codes/plasmod/equilibrium_2d_coupling.py
@@ -41,7 +41,7 @@ from scipy.interpolate import interp1d
 from tabulate import tabulate
 
 from bluemira.base.components import PhysicalComponent
-from bluemira.base.constants import EPS, MU_0
+from bluemira.base.constants import MU_0
 from bluemira.base.file import get_bluemira_path, try_get_bluemira_path
 from bluemira.base.look_and_feel import bluemira_debug, bluemira_print, bluemira_warn
 from bluemira.base.parameter_frame import Parameter, ParameterFrame
@@ -667,9 +667,7 @@ def calc_curr_dens_profiles(
         dum3 = np.gradient(dum2, psi_1D)
         betahat = dum3 / q3 / AA
         chat = -4 * np.pi**2 * MU_0 * pprime / AA
-        #  or np.where(y >= 0, np.sqrt(2.0 * y), -np.sqrt(2.0 * abs(y)))
-        #  or np.sqrt(2.0 * abs(y))
-        FF = np.sqrt(2.0 * np.clip(y, EPS, None))  # noqa: N806
+        FF = np.sqrt(2.0 * np.clip(y, 0, None))  # noqa: N806
 
         ff_prime = 4 * np.pi**2 * (chat - betahat * FF**2)
 
--- a/bluemira/codes/plasmod/equilibrium_2d_coupling.py
+++ b/bluemira/codes/plasmod/equilibrium_2d_coupling.py
@@ -41,7 +41,7 @@ from scipy.interpolate import interp1d
 from tabulate import tabulate
 
 from bluemira.base.components import PhysicalComponent
-from bluemira.base.constants import MU_0
+from bluemira.base.constants import EPS, MU_0
 from bluemira.base.file import get_bluemira_path, try_get_bluemira_path
 from bluemira.base.look_and_feel import bluemira_debug, bluemira_print, bluemira_warn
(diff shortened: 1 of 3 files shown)


CoronelBuendia (approved):
So the changes make sense, and should be more robust. But I am perplexed as we've never seen lack of convergence on this procedure before. It's possible it relates to changes in output coming from PROCESS. In all honesty though, this approach was always particularly fragile, it's just concerning that the one part that seemed to always work was this one!

Don't want to merge just yet. I need to understand better, and this PR might be a good place to do that.

CoronelBuendia:
> So the changes make sense, and should be more robust. But I am perplexed as we've never seen lack of convergence on this procedure before. It's possible it relates to changes in output coming from PROCESS. In all honesty though, this approach was always particularly fragile, it's just concerning that the one part that seemed to always work was this one!
> 
> Don't want to merge just yet. I need to understand better, and this PR might be a good place to do that.

OK I've gone through and checked the example as well. It is good to go. I can pick up the remaining failure in the EU-DEMO procedure separately. Requires some piping of new stuff.

bluemira/codes/plasmod/equilibrium_2d_coupling.py
```diff
@@ -666,7 +667,10 @@ def calc_curr_dens_profiles(
         dum3 = np.gradient(dum2, psi_1D)
         betahat = dum3 / q3 / AA
         chat = -4 * np.pi**2 * MU_0 * pprime / AA
-        FF = np.sqrt(2.0 * y)  # noqa: N806
+        #  or np.where(y >= 0, np.sqrt(2.0 * y), -np.sqrt(2.0 * abs(y)))
+        #  or np.sqrt(2.0 * abs(y))
+        FF = np.sqrt(2.0 * np.clip(y, EPS, None))  # noqa: N806
```
CoronelBuendia, line 672: Actually, why clip at `EPS` and not at `0.0`?

je-cook: no reason really just habit not using 0. I'll change, see what happens

je-cook: still didnt converge on the equilibria example put probably more correct so pushed

bluemira/equilibria/fem_fixed_boundary/utilities.py
```diff
@@ -350,11 +351,9 @@ def get_flux_surfaces_from_mesh(
...
                 index.append(i)
 
-    n = len(index)
-    for xi in range(n):
-        x_1d = np.delete(x_1d, index[xi])
-
-    return x_1d, flux_surfaces
+    mask = np.ones_like(x_1d, dtype=bool)
+    mask[index] = False
+    return x_1d[mask], flux_surfaces
```
CoronelBuendia, line 356: This was definitely a bug. Rather confusingly, fixing it causes the example to no longer converge.

I think I can explain this behaviour, and potentially fix it, but I require time for further investigation.

@je-cook, separate PR or not?

je-cook: in that case a separate issue and PR is probably the way to go. This now just makes the problem slightly more robust without changing behaviour

bluemira/equilibria/fem_fixed_boundary/utilities.py
```diff
@@ -350,11 +351,10 @@ def get_flux_surfaces_from_mesh(
...
                 index.append(i)
 
-    n = len(index)
-    for xi in range(n):
-        x_1d = np.delete(x_1d, index[xi])
-
-    return x_1d, flux_surfaces
+    mask = np.ones_like(x_1d, dtype=bool)
+    mask_index = index + np.arange(len(index))
+    mask[mask_index[mask_index < len(x_1d)]] = False
```
CoronelBuendia, line 356: Right but, we're confident this was a bug, so why leave it in? Can we at least put a comment or TODO relating to this which needs to be fixed.

je-cook: fair enough I'll take it out (and open an issue)

je-cook: #2574

CoronelBuendia (approved):

CoronelBuendia (approved):

hsaunders1904 (approved):

PR #2516, opened 2023-08-14
Replace codeqa dependencies with Ruff

je-cook:
## Description

Ruff is faster, has more plugins that can be enabled and catch quite a few code smells

## Interface Changes

None

## Checklist


je-cook:
List of PRs that need to go in first because they are larger fixes or bugs

- [ ] #2518
- [ ] #2519
- [ ] #2523 
- [ ] #2524

scripts/copyright_update.py
```diff
@@ -70,7 +70,7 @@ def edit_files(files: List[str], copyright_line: str):
 
 def main():
     """Run the copyright updater"""
-    edited = list(edit_files(arguments(), r"2021-{}".format(date.today().year)))
+    edited = list(edit_files(arguments(), rf"2021-{date.today().year}"))  # noqa: DTZ011
```
oliverfunk, line 73: What is the r here for?

oliverfunk (approved):
Went through almost all of the files (got difficult towards the end where GH stops showing the diff automatically).

Looks really good. Use of much more conscience syntax, use of Path and random is improved.

scripts/copyright_update.py
je-cook: The raw text encoding here (I'm guessing you knew that bit) is because edit files reads all files without encoding/decoding the strings so you need to pass a non escaped version of the string. 

Sorry I can't seem to find a good explanation link

bluemira/geometry/optimisation/problem.py
```diff
@@ -61,7 +61,8 @@ def df_objective(self, geom: _GeomT) -> np.ndarray:
         """
         raise NotImplementedError
 
-    def eq_constraints(self) -> List[GeomConstraintT]:
+    @staticmethod
+    def eq_constraints() -> List[GeomConstraintT]:
         """
```
CoronelBuendia, line 66: I don't understand; doesn't that mean that all derived classes should follow the same signature? I.e. staticmethod?

bluemira/optimisation/problem.py
```diff
@@ -94,15 +95,18 @@ def objective(self, x: np.ndarray) -> float:
...
 
-    def eq_constraints(self) -> List[ConstraintT]:
+    @staticmethod
+    def eq_constraints() -> List[ConstraintT]:
         """The equality constraints on the optimisation."""
         return []
 
-    def ineq_constraints(self) -> List[ConstraintT]:
+    @staticmethod
+    def ineq_constraints() -> List[ConstraintT]:
```
CoronelBuendia, line 104: Again with the staticmethods... When you subclass, won't you get warnings for using self thereafter?

CoronelBuendia (commented):
Minor, mostly questions, otherwise looks good. A moment of silence for 88 characters vs. 89 :'(

bluemira/geometry/optimisation/problem.py
je-cook: I take it as the minimum number of inputs to the method is none but any child method can be a normal method (therefore have a self arg) or have any other number of arguments. The base class implements the minimum interface.

Maybe @oliverfunk has some further thoughts?
I'll see if I can find some guidance somewhere.

oliverfunk: I see it as, marking a method as static means it can (should) only be static for itself and all inheritors of the class.

The shape of the input to the function (i.e. the number and type of the parameters) can change, but the way it's called should stay consistent.
(GeomOptimisationProblem.eq_constraints() vs some_insta.eq_constraints())

A method shouldn't sometimes be static and sometimes be dynamic based on the type imo.

je-cook: Ok, after a bit of followup discussion, I'll take away the staticmethod's for ABCs

je-cook: Done

CoronelBuendia (approved):

PR #2541, opened 2023-08-25
Changes to fix J. Roff issues

oliverfunk:
## Linked Issues

None, the issues are described here.

## Description

Some issues pointed out by a user James Roff can fixed with the changes in this PR.
Below are slight adaptions to the minimal reproductions supplied by him.

**A. Vertex returned from value_at throws when used with parameter_at and split_wire**:

```python
p = PictureFrame()
p.adjust_variable("x1", 2, lower_bound=0, upper_bound=5)
p.adjust_variable("x2", 13, lower_bound=10, upper_bound=20)
p.adjust_variable("z1", 13, lower_bound=8, upper_bound=15)
p.adjust_variable("z2", -11, lower_bound=-15, upper_bound=-8)
p.adjust_variable("ri", 2)
p.adjust_variable("ro", 5)

wire = p.create_shape()

dist = (13 + 11 - 2 * 2) + 0.25 * np.pi * 2
vertex = wire.value_at(distance=dist)

wire.parameter_at(vertex) # throws
first_half, second_half = split_wire(wire, vertex) # throws
```

This is due to the split_wire works internally needing to discretize the wire and FPE's around that.

The specific distance `dist` calculated and used isn't important
One can change it to use any overly precise value, if you increase the number of decimal points it fails eventually.

This not really an issue and is easily solved my slightly increasing the EPS tolerance by a factor of 10, making it a value in the order of E-15 which will be fine for every usecase I can think of for Bluemira.

**B. Coordinates class behaviour for 3x3 matrices** -? see #2598 

**C. Small bit added to wire after split_wire on a vertex**

```python
def _split_vertex(wire: BluemiraWire) -> BluemiraWire:
    vertexes = wire.vertexes.points

    v1 = vertexes[-2]
    v2 = vertexes[-1]
    v3 = vertexes[0]
    v4 = vertexes[1]

    _, wire_mid = split_wire(wire, v4)  # This adds a 'tiny' wire to the start
    return wire_mid

# Create new PictureFrame with Tapered Inner Leg
inner_rad = 1
min_x = 4
min_z = -10

p = PictureFrame(
    inner="TAPERED_INNER",
    var_dict={
        "x1": {"value": 3, "lower_bound": 0, "upper_bound": 5},
        "x4": {"value": min_x, "lower_bound": 0, "upper_bound": 5},
        "x2": {"value": 16, "lower_bound": 10, "upper_bound": 20},
        "z1": {"value": 10},
        "z2": {"value": min_z},
        "z3": {"value": 5},
        "ri": {"value": inner_rad},
        "ro": {"value": 2},
    },
)

test_wire = p.create_shape()

inner_radius_offset = (1 - np.sqrt(2) / 2) * inner_rad
bot_left_loc = (min_x + inner_radius_offset, 0, min_z + inner_radius_offset)

new_wire = _split_vertex(test_wire)

try:
    display.plot_2d(new_wire)
except Exception as e:
    print(e)

cadapi.fix_wire(new_wire.shape)  # This fixes the issue
display.plot_2d(new_wire)
```

This one is most likely a real issue, however, it may be with FreeCAD, not Bluemira.
Splitting on `v4` specifically causes the issue, other vertices do not.

As suggested, adding the fix_wire call to the init of a BluemiraWire does fix the issue and causes no noticeable downside.
So I added the change to this PR.

## Checklist

### DIFF
--- a/bluemira/geometry/wire.py
+++ b/bluemira/geometry/wire.py
@@ -61,6 +61,8 @@ class BluemiraWire(BluemiraGeo):
         self, boundary: List[Union[cadapi.apiWire, BluemiraWire]], label: str = ""
     ):
         boundary_classes = [self.__class__, cadapi.apiWire]
+        if isinstance(boundary, cadapi.apiWire):
+            cadapi.fix_wire(boundary)
         super().__init__(boundary, label, boundary_classes)
         self._check_orientations()
 
--- a/bluemira/geometry/wire.py
+++ b/bluemira/geometry/wire.py
@@ -207,7 +207,9 @@ class BluemiraWire(BluemiraGeo):
 
         return cadapi.wire_value_at(self.shape, distance)
 
-    def parameter_at(self, vertex: Iterable[float], tolerance: float = EPS) -> float:
+    def parameter_at(
+        self, vertex: Iterable[float], tolerance: float = EPS * 10
+    ) -> float:
         """
         Get the parameter value at a vertex along a wire.
 
--- a/bluemira/geometry/wire.py
+++ b/bluemira/geometry/wire.py
@@ -61,8 +61,6 @@ class BluemiraWire(BluemiraGeo):
         self, boundary: List[Union[cadapi.apiWire, BluemiraWire]], label: str = ""
     ):
         boundary_classes = [self.__class__, cadapi.apiWire]
-        if isinstance(boundary, cadapi.apiWire):
-            cadapi.fix_wire(boundary)
         super().__init__(boundary, label, boundary_classes)
         self._check_orientations()
 
--- a/bluemira/codes/_freecadapi.py
+++ b/bluemira/codes/_freecadapi.py
@@ -983,7 +983,7 @@ def wire_value_at(wire: apiWire, distance: float) -> np.ndarray:
 
 
 def wire_parameter_at(
-    wire: apiWire, vertex: Iterable[float], tolerance: float = EPS
+    wire: apiWire, vertex: Iterable[float], tolerance: float = EPS * 10
 ) -> float:
     """
     Get the parameter value at a vertex along a wire.
--- a/bluemira/geometry/tools.py
+++ b/bluemira/geometry/tools.py
@@ -940,7 +940,7 @@ def distance_to(
 
 
 def split_wire(
-    wire: BluemiraWire, vertex: Iterable[float], tolerance: float = EPS
+    wire: BluemiraWire, vertex: Iterable[float], tolerance: float = EPS * 10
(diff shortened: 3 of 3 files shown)


je-cook:
I think its worth untransposing Coordinates. It is weird to do that.

bluemira/geometry/wire.py
```diff
@@ -61,6 +61,8 @@ def __init__(
         self, boundary: List[Union[cadapi.apiWire, BluemiraWire]], label: str = ""
     ):
         boundary_classes = [self.__class__, cadapi.apiWire]
+        if isinstance(boundary, cadapi.apiWire):
+            cadapi.fix_wire(boundary)
```
CoronelBuendia, line 65: Please could you do a rough time comparison on the EU-DEMO CAD build sequence with this fixing on and off?

I'd like to get an understanding of the overhead, and whether it's worth applying to all `BluemiraWire`s rather than addressing the root cause of the particular issue in this case.

CoronelBuendia (commented):
LGTM, all makes sense. Worried about the overhead of wire-fixing. Don't have a feel on it - plus would be good to check more complicated build sequences (EU-DEMO example) with this on.

bluemira/geometry/wire.py
oliverfunk: I've run EUDEMO a couple of time with and without these changes.

The full reactor build takes around 96 seconds to finish with tf and pf designers set to read run_mode

It takes in the region of 1.5-2 seconds longer to build with these changes, which is essentially negligible compared to the total time it takes to build.
It seems like very little overhead is added and it doesn't "feel" any slower when running.

I used `time.perf_counter()` to read the time before and after all the build_x functions are run in `reactor.py`

je-cook (approved):
Looks good to me.

Might be good to put a comment in the code for the extra `fix_wire` as it may not be needed once #1347 is solved by a future FreeCAD version.

Other than that if @CoronelBuendia is happy with the slight performance hit, I am.

oliverfunk:
After more tests were run by @je-cook, an unacceptably large performance hit was observed due to the `fix_wire` call, especially during equilibria optimisation and thus will be removed.

The fix to EPS made little to no difference to the runtime performance of the eu-demo reactor build.

The results from the tests below (plotting was turned off):

| No fix  | Fix with `fix_wire` | Fix with EPS only |
| ------------- | ------------- | ------------- |
| 188.5 | 247 | 191 |
| 187 | 246  | 189 |
| 188 | 247 | 189 |

The slight increase with the Fix with EPS only could have come from the fact that Jame's laptop was pretty warm at that point (after running eu demo 6 six times prior).

PR #2576, opened 2023-09-01
FEM fixed boundary equilibrium solver - flux surface cut-off

### LINKED ISSUE #2574: FEM equilibrium example does not converge
## Describe the bug

In #2484 it was discovered that the FEM loop does not converge with a fixed `get_flux_surfaces_from_mesh` function 

## Steps to reproduce

Run `python examples/codes/equilibrium_plasmod_example.ex.py`

The example does not converge in under 15 iterations (roughly 20 mins).

## Expected behaviour

The example to converge, previously this took about 8 iterations and a 5 mins.

- OS: 
- Version: 
- Branch:

CoronelBuendia:
## Linked Issues

Closes #2574 

The FEM fixed boundary equilibrium solver does not do a great job converging in many cases. Right now, since #2484, the example of PLASMOD <-> fixed boundary equilibrium no longer converges.

I recently discovered that this is partly due to poor quality flux surfaces near the axis being extracted from the mesh and used to calculate profile quantities. 

Very commonly, the first one or two flux surfaces as extracted from the mesh (even with refinement) can be diamond- or square-shaped (with the number of points on a flux surface ranging from ~13-50 or so). They can oscillate between the two from one iteration to another (modifying V, g2 and g3), causing poor convergence.

## Description

This PR puts a crude number of points cut-off below which the flux surfaces are not used.

Really, we need a mesh that is adaptive or at least with some knowledge of the problem, a la SPIDER or similar.

Alternatively one could return values of V, g2, and g3 near the axis based on reasonable geometrical extrapolations of the LCFS and the last available flux surface, similar perhaps to what is done in METIS - at least this is done from the LCFS there I believe.

I don't have time for either right now, and frankly would rather invest said time in writing a proper 1.5-D solver with a fixed boundary that can be specified from points. Regardless, this PR fixes convergence on the example, and adds a few settings into EU-DEMO for good measure (although of course it does not presently work for EU-DEMO).

## Checklist

### DIFF
--- a/bluemira/codes/plasmod/equilibrium_2d_coupling.py
+++ b/bluemira/codes/plasmod/equilibrium_2d_coupling.py
@@ -233,6 +233,7 @@ def solve_transport_fixed_boundary(
     refine: bool = False,
     num_levels: int = 2,
     distance: float = 1.0,
+    ny_fs_min: int = 40,
 ):
     """
     Solve the plasma fixed boundary problem using delta95 and kappa95 as target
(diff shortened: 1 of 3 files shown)


ivanmaione (approved):
Totally fine with this hack. Currently, I don't see a better solution that can be quickly implemented.

PR #2594, opened 2023-09-06
`TrapezoidalPrismCurrentSource` bugfixes and API changed to degrees, not radians

### LINKED ISSUE #2592: Trapezoidal prism current source doesn't provide a warning when an invalid shape is created
## Describe the bug

1)

When using the trapezoidal current source certain combinations of upper angle, lower angle and side lengths results in invalid shapes such as in the figure below. When this happens no warning is published in the terminal.

invalid_shape

2)

Additionally the calculation of the magnetic fields when the angles exceed pi/2 deviate from what would be expected by the prism shape. For example with upper and lower angles of 3pi/4 the shape looks identical to pi/4 however the magnetic fields are different and this should also raise a warning. Below is figure showing discrepency between two identical shapes (angles of 0 or pi) where contour is of difference between the two (ie B1 - B2)

discrepency

## Steps to reproduce

To reproduce can run the following code as a script. Will produce the various plots without any warnings being thrown.

```
import numpy as np
from bluemira.magnetostatics.trapezoidal_prism import TrapezoidalPrismCurrentSource
from bluemira.utilities.plot_tools import Plot3D
source = TrapezoidalPrismCurrentSource(
    np.array([0, 0, 0]),
    np.array([0, 0, 2]),
    np.array([1, 0, 0]),
    np.array([0, 1, 0]),
    0.5,
    0.5,
    np.pi * 2 / 4,
    np.pi * 2 / 4,
    1e6,
)
ax = Plot3D()
source.plot(ax=ax)
plt.show()
source2 = TrapezoidalPrismCurrentSource(
    np.array([0, 0, 0]),
    np.array([0, 0, 2]),
    np.array([1, 0, 0]),
    np.array([0, 1, 0]),
    0.5,
    0.5,
    np.pi * 0 / 4,
    np.pi * 0 / 4,
    1e6,
)
source3 = TrapezoidalPrismCurrentSource(
    np.array([0, 0, 0]),
    np.array([0, 0, 2]),
    np.array([1, 0, 0]),
    np.array([0, 1, 0]),
    0.5,
    0.5,
    np.pi * 4 / 4,
    np.pi * 4 / 4,
    1e6,
)
ax = Plot3D()
x = np.linspace(-2, 2, 30)
y = np.linspace(-2, 2, 30)
xx, yy = np.meshgrid(x, y)
zz = np.zeros_like(xx)
bx, by, bz = source2.field(xx, yy, zz)
bx2, by2, bz2 = source3.field(xx, yy, zz)
B = np.sqrt(bx**2 + by**2 + bz**2)
B2 = np.sqrt(bx2**2 + by2**2 + bz2**2)
source2.plot(ax=ax)
source3.plot(ax=ax)
ax.contourf(xx, yy, B-B2, zdir="z", offset=0.0, cmap="magma")
plt.show()
```

## Resolution

Potential solution to 1) would be to test whether the planes of the upper and lower rectangles intercept after the shape is created and if they do raise an error. For 2) either raise warnings when other angles are used (maybe with tests for these cases) or alter how the angles are used in the magnetostatics calculations so that when it goes outside the valid range it is moved (ie pi -> 0).

CoronelBuendia:
## Linked Issues

Closes #2592 

## Description
Fixes the issues in the aforementioned issue, but also:
* Moves the API of `TrapezoidalPrismCurrentSource` to use degrees and not radians
* Caps the inputs of valid angles to the necessary [0, 180)
* Deals with negative angles (allowed) and protects against mixed angles (not allowed)

## Checklist

### DIFF
--- a/bluemira/magnetostatics/circuits.py
+++ b/bluemira/magnetostatics/circuits.py
@@ -141,7 +141,7 @@ class ArbitraryPlanarRectangularXSCircuit(SourceGroup):
             beta = alpha
             betas.append(beta)
 
-        return betas, alphas
+        return np.rad2deg(betas), np.rad2deg(alphas)
 
     def _transform_to_xz(self, shape):
         normal_vector = shape.normal_vector
--- a/bluemira/magnetostatics/trapezoidal_prism.py
+++ b/bluemira/magnetostatics/trapezoidal_prism.py
@@ -379,13 +379,13 @@ class TrapezoidalPrismCurrentSource(RectangularCrossSectionCurrentSource):
             )
         if not (0 <= abs(alpha) < 0.5 * np.pi):
             raise MagnetostaticsError(
-                f"{self.__class__.__name__} instantiation error: {alpha=} is outside "
-                "bounds of [0, pi/2)"
+                f"{self.__class__.__name__} instantiation error: {alpha=:.3f} is outside "
+                "bounds of [0, pi/2)."
             )
         if not (0 <= abs(beta) < 0.5 * np.pi):
             raise MagnetostaticsError(
-                f"{self.__class__.__name__} instantiation error: {beta=} is outside "
-                "bounds of [0, pi/2)"
+                f"{self.__class__.__name__} instantiation error: {beta=:.3f} is outside "
+                "bounds of [0, pi/2)."
             )
 
     def _check_raise_self_intersection(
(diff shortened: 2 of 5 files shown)


CoronelBuendia:
@kj5428, FYI

je-cook (approved):

Issue #1320, opened 2022-08-10
Plotting test helper functions


je-cook:
## Description of issue / requirement to address

Add a helper for plotting during tests to deal with creating and closing a figure so the tests work without having an interactive backend.

## Proposed solution

Create some helper classes

## Alternative solutions

Do nothing and make people close their plots

## Additional context

#740

CoronelBuendia:
@je-cook I believe this works now, no?

je-cook:
What we currently do works but we havent made a helper function so you dont have to remember to write a load of things in setup and teardown methods

PR #2597, opened 2023-09-06
`BiotSavartFilament` plotting fixed

### LINKED ISSUE #2593: BiotSavartFilament behaviour for open wires need clarification
When using the BiotSavartFilament current source for finite straight wires the source removes the final coordinate point as it assumes it is a closed wire, This can result in discrepancy between the expected and actual results. If you set up a line with 3 points from -1 to 1 in z and plot the source, only from -1 to 0 is plotted. 

```
coords1 = np.array([
        [0,0,-1],
        [0,0,0],
        [0,0,1],
])
source1 = BiotSavartFilament(coords1, 0.05, 1e6)
source1.plot()
plt.show()
```

incomplete_wire

This then results in the magnetic field plots being inaccurate. As seen below in the difference in actual and expected value for the magnetic field as x position is varied.

incomplete_wire_field

Additionally different weird behaviour occurs if you have wires where the final point (the one that gets removed) is different but all other points the same then the magnetic field also varies. Coordinates given below with the magnetic result afterwards.

```
coords1 = np.array([
        [0,0,-1],
        [0,0,0],
        [0,0,1]
])
coords2 = np.array([
        [0,0,-1],
        [0,0,0],
        [0,0,2]
])
```

ident_wire

This is certainly not expected with how it behaved before. Hopefully this problem is tied in with the other problem so is fixed at the same time.

CoronelBuendia:
## Linked Issues

Closes #2593 

## Description
At present all I have done is fix the plotting, which was not showing the last point of a single filament. 

This sounds quite innocuous, but to avoid a whole bunch of length checking and closed/open checking for each individual array, I've simplified things down, at potentially some non-negligible additional computational cost. If Biot-Savart ever shows up in profiling, I can take another look at this.

## Checklist

### DIFF
--- a/tests/magnetostatics/test_biot_savart.py
+++ b/tests/magnetostatics/test_biot_savart.py
@@ -56,6 +56,7 @@ def test_biot_savart_loop():
     circle = make_circle(radius=x_coil, center=(0, z_coil, 0), axis=(0, 0, 1))
     filament = circle.discretize(ndiscr=2000)
     bsf = BiotSavartFilament(filament, radius)
+    bsf.plot()
 
     Bx2, _, Bz2 = bsf.field(x_2d, np.zeros_like(x_2d), z_2d)
 
--- a/tests/magnetostatics/test_biot_savart.py
+++ b/tests/magnetostatics/test_biot_savart.py
@@ -57,6 +57,8 @@ def test_biot_savart_loop():
     filament = circle.discretize(ndiscr=2000)
     bsf = BiotSavartFilament(filament, radius)
     bsf.plot()
+    plt.show()
+    plt.close()
 
     Bx2, _, Bz2 = bsf.field(x_2d, np.zeros_like(x_2d), z_2d)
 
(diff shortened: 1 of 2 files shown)


tests/magnetostatics/test_biot_savart.py
```diff
@@ -56,6 +56,7 @@ def test_biot_savart_loop():
     circle = make_circle(radius=x_coil, center=(0, z_coil, 0), axis=(0, 0, 1))
     filament = circle.discretize(ndiscr=2000)
     bsf = BiotSavartFilament(filament, radius)
+    bsf.plot()
```
je-cook, line 59: Although this is closed in the `plot_errors` function we should really close this one here or at the end of this function. I should really get around to #1320...
```suggestion
    bsf.plot()
    plt.show()
    plt.close()
```

je-cook (approved):
Other than the one test change looks good. The only slow down would be on rotations because of the extra calculation as I see it, maybe I've missed something. At any rate it the plotting is easier to read.

Issue #2239, opened 2023-05-12
If `Equilibrium` is inited with an eqdsk without a coilset it errors


je-cook:
## Describe the bug

If you try to initiailse an `Equilibrium` object with an eqdsk that doesnt have a coilset it fails with the new `Coilset`. In previous this means that all the coilset methods returned 0 but now a `Coilset` can't be created without a coil.

## Steps to reproduce

```python
Equilibrium.from_eqdsk('file.eqdsk') # an error is raise if there is no coilset
```

## Expected behaviour

Adding coils to an eqdsk is something we have added to the standard and is not in most implementations so this should be allowed. It worked before coils was rewritten (eg v0.0.0.dev1)

## Screenshots / Tracebacks

N/A

## Possible solution

Create an `EmptyCoilset` class that mirrors the methods of a coilset but returns 0 or other sensible results for the methods

## Environment (please complete the following information)

- OS: Ubuntu 22.04
- Version: 1.3+
- Branch:develop

je-cook:
fixed in #2617

PR #2617, opened 2023-09-15
Temporary for Issue #2239 -  eqdsk without a coilset

geograham:
## Linked Issues

#2239

Closes #{ID}

## Description

If you try to initiailse an Equilibrium object with an eqdsk that doesnt have a coilset it fails with the new Coilset.

Added a user input option for a coilset when initiailse an Equilibrium and a dummy coilset that is created if no coilset is present.  Dummy coils are placed at the corners of the grid, and have current, dx, dz, j_max and b_max set to zero. A warning is printed if the dummy coilset is used. This should allow us to initiailse an eqdsk with no coilset until we have a permanent solution. 

## Checklist

### DIFF
--- a/bluemira/equilibria/coils/_grouping.py
+++ b/bluemira/equilibria/coils/_grouping.py
@@ -46,7 +46,6 @@ from bluemira.equilibria.coils._tools import (
 )
 from bluemira.equilibria.constants import I_MIN
 from bluemira.equilibria.error import EquilibriaError
-from bluemira.equilibria.grid import Grid
 from bluemira.equilibria.plotting import CoilGroupPlotter
 from bluemira.utilities.tools import flatten_iterable, yintercept
 
@@ -167,13 +166,9 @@ class CoilGroup(CoilGroupFieldsMixin):
         kwargs:
             passed to matplotlib's Axes.plot
         """
-        if self.ctype == CoilType.DUM:
-            # Do not plot if it is a dummy coil
-            pass
-        else:
-            return CoilGroupPlotter(
-                self, ax=ax, subcoil=subcoil, label=label, force=force, **kwargs
-            )
+        return CoilGroupPlotter(
+            self, ax=ax, subcoil=subcoil, label=label, force=force, **kwargs
+        )
 
     def fix_sizes(self):
         """
@@ -291,26 +286,6 @@ class CoilGroup(CoilGroupFieldsMixin):
         pfcoils = []
         cscoils = []
         passivecoils = []
-        if eqdsk.ncoil < 1:
-            grid = Grid.from_eqdsk(eqdsk)
-            dum_xc = [np.min(grid.x), np.max(grid.x), np.max(grid.x), np.min(grid.x)]
-            dum_zc = [np.min(grid.z), np.min(grid.z), np.max(grid.z), np.max(grid.z)]
-            for i in range(4):
-                coil = Coil(
-                    dum_xc[i],
-                    dum_zc[i],
-                    current=0,
-                    dx=0,
-                    dz=0,
-                    ctype="DUM",
-                    j_max=0,
-                    b_max=0,
-                )
-                coil.fix_size()  # Oh ja
-                pfcoils.append(coil)
-            coils = pfcoils
-            bluemira_warn("EQDSK coilset empty - dummy coilset in use.")
-            return cls(*coils)
         for i in range(eqdsk.ncoil):
             dx = eqdsk.dxc[i]
             dz = eqdsk.dzc[i]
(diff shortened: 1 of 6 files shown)


bluemira/equilibria/coils/_grouping.py
```diff
@@ -283,6 +284,26 @@ def from_group_vecs(cls, eqdsk: EQDSKInterface):
...
+            for i in range(4):
+                coil = Coil(
+                    dum_xc[i],
+                    dum_zc[i],
+                    current=0,
+                    dx=0,
+                    dz=0,
+                    ctype="PF",
+                    j_max=0,
+                    b_max=0,
```
CoronelBuendia, line 300: maybe also `control=False`

bluemira/equilibria/coils/_grouping.py
```diff
@@ -283,6 +284,26 @@ def from_group_vecs(cls, eqdsk: EQDSKInterface):
         pfcoils = []
         cscoils = []
         passivecoils = []
+        if eqdsk.ncoil < 1:
```
CoronelBuendia, line 287: Does a coilset not work with one coil?

bluemira/equilibria/coils/_grouping.py
```diff
@@ -283,6 +284,26 @@ def from_group_vecs(cls, eqdsk: EQDSKInterface):
...
+            grid = Grid.from_eqdsk(eqdsk)
+            dum_xc = [np.min(grid.x), np.max(grid.x), np.max(grid.x), np.min(grid.x)]
+            dum_zc = [np.min(grid.z), np.min(grid.z), np.max(grid.z), np.max(grid.z)]
+            for i in range(4):
```
CoronelBuendia, line 291: What's the rationale for 4 dummy coils here?

bluemira/equilibria/coils/_grouping.py
```diff
@@ -283,6 +284,26 @@ def from_group_vecs(cls, eqdsk: EQDSKInterface):
...
+                coil = Coil(
+                    dum_xc[i],
+                    dum_zc[i],
+                    current=0,
+                    dx=0,
+                    dz=0,
+                    ctype="PF",
```
CoronelBuendia, line 298: Perhaps we could have a dummy type? And make sure they don't get plotted?

CoronelBuendia (commented):
Hi @geograham, thanks for this. A few quick comments. Do you know if the use case is for loading an equilibrium from a fixed boundary solver? If so, you may find there are issues with reconstruction of the psi field / separatrix finding.

I was always in two minds about using a FixedBoundaryEquilibrium object for theses cases. Unfortunately, the one available in bluemira at the moment is not fit for this purpose.

Some tests would be nice! You can add a test in the private tests (not visible OS), too, to check we can load this EQDSK you are trying to read in our tests.

bluemira/equilibria/coils/_grouping.py
je-cook: it does, it just doesnt work with 0 coils atm

bluemira/equilibria/coils/_grouping.py
geograham: New ctype=DUM added and coils do not plot.

bluemira/equilibria/optimisation/problem/base.py
```diff
@@ -140,8 +140,8 @@ def read_coilset_state(
         Number of substates (blocks) in the state vector.
         """
         substates = 3
-        x, z = coilset.position
-        currents = coilset.current / current_scale
+        x, z = coilset.get_control_coils().position
```
oliverfunk, line 143: I'm slightly concerned that if this is being called when reading the coilset state, the set_coilset_state method below doesn't set the state on the control coils only, just sets the coilset currents to the state vector passed in.

But wouldn't that break? Wouldn't they be different array lengths?

bluemira/equilibria/coils/_grouping.py
geograham: I agree, control=False added.

bluemira/equilibria/coils/_grouping.py
geograham: A coil in each corner of the grid seemed ok as they should not be used. But I could change this.

bluemira/equilibria/optimisation/problem/base.py
CoronelBuendia: Is this related to loading an EQDSK with no coils? This changes this function to reading the coilset state, to reading the control coilset state. Separate function? Indexing after the fact?

geograham: Sorry, this should not be on this branch. I am removing it.

CoronelBuendia (approved):

je-cook (approved):

PR #2619, opened 2023-09-15
Polyhedral prism current source

kj5248:
Addition of polyhedral current source. The basis of issue 1816 but doesn't achieve it in a analytic way as desired. Simply a first step to that by using a series of trapezoidal current sources to calculate the magnetic field.

Tests added using trapezoidal tests as a base.


CoronelBuendia:
CLosed in favour of #2703

PR #2641, opened 2023-09-21
Add `FixedPlasmaEquilibrium` for loading fixed boundary EQDSKs.

### LINKED ISSUE #2639: Ability to handle EQDSK files from fixed boundary equilibrium solvers

Technically, the G-EQDSK file format is exclusively for fixed boundary equilibria, and yet we cannot load an `Equilibrium` from one.

Such that the following will error (on branch #2617 - which fixes the issue of coilsets being required):

```python
from bluemira.equilibria.equilibrium import Equilibrium
eq = Equilibrium.from_eqdsk("my_fixed_boundary_eq.eqdsk")
```

with:

```python
>>> eq = Equilibrium.from_eqdsk(path)
+-------------------------------------------------------------------------+
| WARNING: EQDSK coilset empty - dummy coilset in use.                    |
+-------------------------------------------------------------------------+
  File "/home/matti/code/bluemira/bluemira/equilibria/equilibrium.py", line 675, in from_eqdsk
    jtor = profiles.jtor(grid.x, grid.z, psi, o_points=o_points, x_points=x_points)
  File "/home/matti/code/bluemira/bluemira/equilibria/profiles.py", line 772, in jtor
    psisep, psiax, mask = self._jtor(x, z, psi, o_points, x_points)
  File "/home/matti/code/bluemira/bluemira/equilibria/profiles.py", line 469, in _jtor
    mask = in_plasma(x, z, psi, o_points, x_points)
  File "/home/matti/code/bluemira/bluemira/equilibria/find.py", line 982, in in_plasma
    lcfs, _ = find_LCFS_separatrix(x, z, psi, o_points=o_points, x_points=x_points)
  File "/home/matti/code/bluemira/bluemira/equilibria/find.py", line 729, in find_LCFS_separatrix
    flux_surface = get_flux_loop(middle)
  File "/home/matti/code/bluemira/bluemira/equilibria/find.py", line 715, in get_flux_loop
    f_s = find_flux_surf(x, z, psi, psi_norm, o_points=o_points, x_points=x_points)
  File "/home/matti/code/bluemira/bluemira/equilibria/find.py", line 566, in find_flux_surf
    raise EquilibriaError(f"No flux surface found for psi_norm = {psinorm:.4f}")
bluemira.equilibria.error.EquilibriaError: No flux surface found for psi_norm = 1.0000
>>> 
```

This is simply because the `Equilibrium` object implicitly assumes a free boundary equilibrium, and many of the methods and operations within require the definition of either some normalised psi value, or some plasma boundary (LCFS), X-point, or even a separatrix.

## Proposed solution

I think it is fair to say that we want to be able to:
1. Load a fixed boundary equilibrium from an EQDSK
2. Have some representation of a fixed boundary equilibrium from which we can do a few operations (e.g. plot, field at some point, etc.)
3. Be able to reach an `Equilibrium` object from some fixed boundary EQDSK, but adding in some coils etc. There are so many ways of doing that, that it may never an actual function that we provide, but we could offer an example.

I'm not 100% on this, because it gets messy either way, but because:
* `Equilibrium` at the moment is already pretty heavy, and contains a little too many mechanics for my liking. It is also not a proper data structure, per se.
* Fixed boundary equilibria (no coil, no separatrix) are in any case appreciably different from free boundary equilibria.
* There are a great deal of methods on `Equilibrium` that make no sense for a fixed boundary equilibrium.

I propose to introduce a third `MHDState`, working name `FixedPlasmaEquilibrium` which effectively act as a finite difference representation of a fixed boundary equilibrium (plasma, no coils), much like `Breakdown` is an [sic] equilibrium without a plasma. This would tackle 1 and 2.

I think for 3, we could start with an example and perhaps leave it there.
## Alternative solutions

I have briefly considered `Equilibrium.from_fixed_eqdsk(path)` as an approach, but really there are relatively few things that you would get out of this (grid, psi, and some boundary but no separatrix or X-point). Then you'd be left with a flag saying that it is fixed boundary and have to check it every time one of the methods requires something that isn't possible with a fixed boundary equilibrium.

## Additional context

We already have a `FixedBoundaryEquilibrium` object in `equilibria/fem_fixed_boundary/fem_magnetostatic_2d.py` which is not really an appropriate formulation (at present) for loading a fixed boundary EQDSK, in particular as it relies on a FE formulation, whereas the EQDSK file format lends itself better to FD.

CoronelBuendia:
## Linked Issues

Closes #2639

## Description

Adds `FixedPlasmaEquilibrium` to handle loading EQDSKs that come from fixed boundary solvers. I'm not sure what the convention is regarding extrapolation of poloidal magnetic flux outside of the plasma boundary.

It appears JETTO does not do this for you, so the resulting `FixedPlasmaEquilibrium` looks like this:

image
image

It would be easy enough to include extrapolation, but a little fiddly if other solvers actually do provide non-0 poloidal magnetic flux outside the plasma.

There is the outstanding issue of #318, which is not addressed here.

## Checklist

### DIFF
--- a/tests/equilibria/test_equilibrium.py
+++ b/tests/equilibria/test_equilibrium.py
@@ -418,7 +418,9 @@ class TestFixedPlasmaEquilibrium:
     @classmethod
     def setup_class(cls):
         root = try_get_bluemira_private_data_root()
-        cls.path = Path(root, "equilibria", "STEP_SPR_08", "jetto.eqdsk_out")
+        path = Path(root, "equilibria", "STEP_SPR_08", "jetto.eqdsk_out")
+        cls.eq = FixedPlasmaEquilibrium.from_eqdsk(path)
 
-    def test_init(self):
-        eq = FixedPlasmaEquilibrium.from_eqdsk(self.path)
+    def test_plotting(self):
+        self.eq.plot()
+        plt.show()
--- a/tests/equilibria/test_equilibrium.py
+++ b/tests/equilibria/test_equilibrium.py
@@ -421,6 +421,7 @@ class TestFixedPlasmaEquilibrium:
         path = Path(root, "equilibria", "STEP_SPR_08", "jetto.eqdsk_out")
         cls.eq = FixedPlasmaEquilibrium.from_eqdsk(path)
 
-    def test_plotting(self):
-        self.eq.plot()
+    @pytest.mark.parametrize("field", [False, True])
+    def test_plotting(self, field):
+        self.eq.plot(field=field)
         plt.show()
--- a/tests/equilibria/test_equilibrium.py
+++ b/tests/equilibria/test_equilibrium.py
@@ -425,3 +425,4 @@ class TestFixedPlasmaEquilibrium:
     def test_plotting(self, field):
         self.eq.plot(field=field)
         plt.show()
+        plt.close()
--- a/tests/equilibria/test_equilibrium.py
+++ b/tests/equilibria/test_equilibrium.py
@@ -28,7 +28,7 @@ import pytest
 from matplotlib import pyplot as plt
 
 from bluemira.base.file import get_bluemira_path, try_get_bluemira_private_data_root
-from bluemira.equilibria.equilibrium import Equilibrium
+from bluemira.equilibria.equilibrium import Equilibrium, FixedPlasmaEquilibrium
 from bluemira.equilibria.file import EQDSKInterface
 from bluemira.equilibria.grid import Grid
 from bluemira.equilibria.opt_constraints import (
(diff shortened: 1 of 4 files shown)


CoronelBuendia:
So I am relatively happy with this in terms of loading a fixed boundary equilibrium, and have also got something for extrapolation of the poloidal flux, which I have left out of this PR. Really I think we need to treat #318 before going any further with this. I think our present EQDSK loading cannot handle JETTO output properly.

tests/equilibria/test_equilibrium.py
```diff
@@ -411,3 +411,17 @@ def test_q_benchmark(self):
...
+    @classmethod
+    def setup_class(cls):
+        root = try_get_bluemira_private_data_root()
+        path = Path(root, "equilibria", "STEP_SPR_08", "jetto.eqdsk_out")
+        cls.eq = FixedPlasmaEquilibrium.from_eqdsk(path)
+
+    @pytest.mark.parametrize("field", [False, True])
+    def test_plotting(self, field):
+        self.eq.plot(field=field)
+        plt.show()
```
je-cook, line 427:
```suggestion
        plt.show()
        plt.close()
```

je-cook (approved):
Looks good just the one plot close thing

CoronelBuendia:
> So I am relatively happy with this in terms of loading a fixed boundary equilibrium, and have also got something for extrapolation of the poloidal flux, which I have left out of this PR. Really I think we need to treat #318 before going any further with this. I think our present EQDSK loading cannot handle JETTO output properly.

I found an ASTRA eqdsk which I'm pretty sure is fixed boundary. That does have psi outside of the LCFS, so I'm not super comfortable providing default extrapolation.

PR #2690, opened 2023-10-06
🚑️ Fixing PROCESS InDat writer

je-cook:
## Description

This aims to fix our PROCESS InDat writer which would only allow values from the default `ProcessInputs` to be written in the `template_in_dat`.

I think this still needs a bit of cleanup but it does, as far as I can tell, work, I will also add a test.

@CoronelBuendia, @oliverfunk 
## Interface Changes

We now setup the defaults for the `ParameterFrame` directly from the provided template whereas before the default template would be used as a base. 

In previous versions if any variable was set to send, not updated in the input params to the solver and a non default `template_in_dat` was used, the variables default from the default template would overwrite the value in the `template_in_dat`

## Checklist


tests/codes/process/test_solver.py
```diff
@@ -96,70 +99,134 @@ def test_get_species_fraction_retrieves_parameter_value(self):
...
-        assert solver.params.r_ts_ib_in.value == pytest.approx(3.47836)
-        assert solver.params.r_vv_ib_in.value == pytest.approx(4.09836)
-        assert solver.params.r_fw_ib_in.value == pytest.approx(4.89136)
-        assert solver.params.r_fw_ob_in.value == pytest.approx(12.67696)
-        assert solver.params.r_vv_ob_in.value == pytest.approx(13.69696)
+        assert solver.params.r_ts_ib_in.value == pytest.approx(3.52836)
+        assert solver.params.r_vv_ib_in.value == pytest.approx(4.14836)
+        assert solver.params.r_fw_ib_in.value == pytest.approx(4.94136)
+        assert solver.params.r_fw_ob_in.value == pytest.approx(12.72696)
+        assert solver.params.r_vv_ob_in.value == pytest.approx(13.74696)
```
CoronelBuendia, line 144: Where does the additional +0.05 come from? We've had radial build mismatches with PROCESS before, and these changes are reminiscent of those issues, if not symptomatic.

CoronelBuendia (commented):
OK a lot going on, but this certainly looks like it will be a better starting basis. Just one comment on something that fills me with dread.

tests/codes/process/test_solver.py
je-cook: It comes from this line: 
https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/9befa589f939fd631014a5ba69f2e1c3c894f630/bluemira/codes/process/_teardown.py#L235
which came from here
https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/1489/commits/d0ad188afaa18dc124d3e3fbb71b8009a1b2cfd1

CoronelBuendia: That line looks wrong. Almost certainly. The thermal shield thickness is (inexplicably) "split" in PROCESS, in the sense that they have three or even four parameters for it (inboard, outboard, up, down). We only have one, because a thermal shield is just a thin sheet of metal with some pipes through which it is cryogenically cooled.

So out thermal shield thickness I think should just be the inboard thermal shield thickness, because `tk_ts` = `thshield_ib` = `thshield_ob`, etc. Adding them up just means doubling the thermal shield thickness in practice.

je-cook: Ok, I've just changed it to `shield_th = data["thshield_ib"]`

PR #2703, opened 2023-10-12
Feature: polyhedral cross-section current sources and a general VIM formulation of 3-D magnetostatics

### LINKED ISSUE #1816: Feature: arbitrary cross section current sources

Currently we only have analytical forms of rectangular cross section magnets following an arbitrary path. 

It would be nice to be able to support analytical forms of other cross sections.

## Proposed solution

This paper even though it is for permanent magnets may help with the generalisation:
doi.org/10.1016/j.jmmm.2020.166894

EDIT (Matti): I have found a tractable formulation of a volumetric integral approach for vector potential and magnetic field here: https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=4407584. Of the dozen or so approaches I ended up looking at, this was by far the most understandable and easy to implement. I have written a PR for this, but a lot remains to be done for me to  be happy with this:

* [ ] Fix tests
  * [ ] Babic and Aykel verification test from the `TrapezoidalPrism` is failing and I don't know who is right, but I suspect we have some singularity issues
  * [ ] Benchmarking with the `TrapezoidalPrism` we see differences when `alpha` and `beta` are not equal. Again not sure who is right.
* [ ] Add tests:
  * [ ] Test that two triangle sources make up a rectangular source
  * [ ] Test continuity of subsequent sources
  * [ ] Test field and potential symmetry of a symmetric polyhedral current source
  * [ ] Verification test of field and potential from a circular rectangular cross-section make with a polyhedral circuit with analytical cylindrical equivalents
  * [ ] Test rotations 
  * [ ] etc..
* [ ] Make some assembly utilities (see e.g. #2704)
* [ ] Write documentation
* [ ] Improve performance (e.g. finish JITting)
* [ ] Add examples:
  * [ ] Add a simple `PolyhedralPrismCurrentSource` example
  * [ ] Add a simple `AbitraryPlanarPolyhedralCircuit` example
* [ ] etc..

@kj5248 
## Alternative solutions

It is possible to just use an arrangement of `BiotSavartFilament`s to have a very similar result.

## Additional context

This is useful for STEP, in that arbitrary cross-section conductors can be used, but I was interested in this stuff because I needed a general way to do volume integrals to calculate things like vector potential, and then self-inductance, etc. It's also potentially useful for stellarators, in that the conductors no longer need to be planar.

CoronelBuendia:
## Linked Issues

Closes #1816 (almost)

This is useful for STEP, in that arbitrary cross-section conductors can be used, but I was interested in this stuff because I needed a general way to do volume integrals to calculate things like vector potential, and then self-inductance, etc. It's also potentially useful for stellarators, in that the conductors no longer need to be planar.

UPDATE (matti): As discussed 11/12 with the STEP team, at present we have only been successful in implementing a VIM method for arbitrary cross-sections with div J = 0 (equal end-cap angles). As it is possible to model their desired geometry using such a source term, I will update this feature branch to bring in an equal-angle VIM formulation with the necessary caveats. At a later date, and if a suitable implementation can be developed, we can replace it with a more general formulation.

## Description
Adds `PolyhedralPrismCurrentSource` where you specify a set of x-z `Coordinates` rather than rectangular dimensions.

image
image

I've benchmarked results against the `RectangularPrismCurrentSource`, and it is a mixed bag:
* For a simple source modelled with both approaches, I get the same results
* For the Babic and Aykel paper verification test it fails spectacularly on both points at the moment, and I can't work out why.

More generally, the formulation does allow for arbitrary construction of conducting geometry, and we may want to provide a mesh interface for this volume integral approach

NOTE: `RectangularPrismCurrentSource` will stay alive for the foreseeable future. It's considerably faster. I have yet to fully JIT the polyhedral one because of reflected list issues, but I did performance benchmark a fully JIT-ed version of the polyhedral stuff for n=4 (dodging list reflection) and it is still much, much slower.

## Checklist


bluemira/magnetostatics/polyhedral_prism.py
```diff
@@ -0,0 +1,462 @@
...
+    d = (
+        r1r * r2r * r3r
+        + r3r * np.dot(r1_r, r2_r)
+        + r2r * np.dot(r1_r, r3_r)
+        + r1r * np.dot(r2_r, r3_r)
+    )
+    a = np.dot(r1_r, np.cross(r2_r, r3_r))
+    # Not sure this is an issue...
+    # if abs(a) < ZERO_GUARD_EPS and (-ZERO_GUARD_EPS < d < 0):
+    #     return 0 # and not pi as per IEEE
```
je-cook, line 124: I hope its not an issue...

bluemira/magnetostatics/trapezoidal_prism.py
```diff
@@ -367,46 +371,10 @@ def __init__(
         self.alpha = alpha
         self.beta = beta
         # Current density
-        self.rho = current / (4 * breadth * depth)
+        self.area = 4 * breadth * depth
```
je-cook, line 374: remove the breadth and depth lines above too if you take my earlier suggestion
```suggestion
        self._breadth_depth_area(breadth, depth)
```

je-cook (commented):
First run through, looks good gonna run the tests etc now

bluemira/magnetostatics/baseclass.py
```diff
@@ -192,9 +192,84 @@ def plot(self, ax: Optional[Axes] = None, show_coord_sys: bool = False):
...
+
+class RectangularCrossSectionCurrentSource(CrossSectionCurrentSource):
+    """
+    Abstract base class for a current source with a rectangular cross-section.
+    """
+
+    breadth: float
+    depth: float
+    length: float
+
```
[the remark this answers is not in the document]
CoronelBuendia: I'd go a step further and maybe say this baseclass isn't necessary... I don't anticipate another for it, and `TrapezoidalPrism` already handles the area stuff seld-consistently...

What about removing this base class, and making breadth and depth private attributes of `TrapezoidalPrism` to discourage on-the-fly re-assignment?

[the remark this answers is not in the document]
je-cook: Yeah I agree, sounds good

bluemira/magnetostatics/baseclass.py
```diff
@@ -192,9 +192,84 @@ def plot(self, ax: Optional[Axes] = None, show_coord_sys: bool = False):
...
+        angle:
+            The rotation degree [degree]
+        axis:
+            The axis of rotation
+        """
+        super().rotate(angle, axis)
+        r = rotation_matrix(np.deg2rad(angle), axis).T
+        self.face_normals = np.array([n @ r for n in self.face_normals])
+        self.face_points = np.array([p @ r for p in self.face_points])
+        self.mid_points = np.array([p @ r for p in self.mid_points])
```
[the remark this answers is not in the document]
CoronelBuendia: I think this whole functionality needs a profile and then I'd like to do all the speed-ups in a single PR. Thing is, it's quite possible it would be better to move to a different datastructure to do the maths faster, and that might be a good candidate for some hardcore einsumming. The rotations are not concerning at the moment. If they are, there might actually be cute tricks to rotate the field points rather than the sources in the `HelmholtzCage` but that's another story.

je-cook (commented):
Other than the tests I'm happy

je-cook:
Just fyi this warning needs to be fixed before merge:
image

bluemira/magnetostatics/polyhedral_prism.py
```diff
@@ -0,0 +1,796 @@
...
+        # Normalised direction cosine matrix
+        self._dcm = np.array([t_vec, ds / length, normal])
+        self._set_cross_section(xs_coordinates)
+
+        self._alpha = alpha
+        self._beta = beta
+
+        # Current density
+        self.set_current(current)
+        self._points = self._calculate_points()
```
je-cook, line 435: just for my awareness where is this used? I can't find where it is.

tests/magnetostatics/test_polyhedral_prism.py
```diff
@@ -0,0 +1,357 @@
...
+
+        f = plt.figure()
+        ax = f.add_subplot(1, 1, 1, projection="3d")
+        self.hexagon.plot(ax)
+        ax.set_title("HexagonPrism")
+        Bx, By, Bz = self.hexagon.field(xx, yy, zz)
+        B_new = np.sqrt(Bx**2 + By**2 + Bz**2)
+        args_new = [xx, yy, zz, B_new]
+        cm = ax.contourf(args_new[i], args_new[j], args_new[k], zdir=plane, offset=0)
+        f.colorbar(cm)
```
je-cook, line 262: there is no assert in this test. Could we add one if possible?

tests/magnetostatics/test_polyhedral_prism.py
```diff
@@ -0,0 +1,357 @@
...
+
+        f = plt.figure()
+        ax = f.add_subplot(1, 1, 1, projection="3d")
+        self.triangle.plot(ax)
+        ax.set_title("TrianglePrism")
+        Bx, By, Bz = self.triangle.field(xx, yy, zz)
+        B_new = np.sqrt(Bx**2 + By**2 + Bz**2)
+        args_new = [xx, yy, zz, B_new]
+        cm = ax.contourf(args_new[i], args_new[j], args_new[k], zdir=plane, offset=0)
+        f.colorbar(cm)
```
je-cook, line 277: Same here no assert

je-cook (changes requested):
On the whole looks good. A couple of questions. I'll add the test plotting stuff as its minor

je-cook:
also there is still an extra warning lying about: /home/runner/work/bluemira/bluemira/bluemira/magnetostatics/polyhedral_prism.py:323: NumbaPerformanceWarning: np.dot() is faster on contiguous arrays, called on (Array(float64, 1, 'C', False, aligned=True), Array(float64, 1, 'A', False, aligned=True)

je-cook:
and possibly a rebase to bring in the new ruff config

bluemira/magnetostatics/polyhedral_prism.py
CoronelBuendia: Nothing turned up so far. I'm of the opinion that this comment deserves to exist, because it remains true, and is a potential starting point if something weird does ever crop up

bluemira/magnetostatics/polyhedral_prism.py
```diff
@@ -0,0 +1,796 @@
...
+    d = (
+        r1r * r2r * r3r
+        + r3r * np.dot(r1_r, r2_r)
+        + r2r * np.dot(r1_r, r3_r)
+        + r1r * np.dot(r2_r, r3_r)
+    )
+    a = np.dot(r1_r, np.cross(r2_r, r3_r))
+    # Not sure this is an issue...
+    # if abs(a) < ZERO_GUARD_EPS and (-ZERO_GUARD_EPS < d < 0):
+    #     return 0 # and not pi as per IEEE
```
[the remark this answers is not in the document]
CoronelBuendia: Nothing turned up so far. I'm of the opinion that this comment deserves to exist, because it remains true, and is a potential starting point if something weird does ever crop up

bluemira/magnetostatics/polyhedral_prism.py
CoronelBuendia: Yeah it is a necessary attribute coming from the baseclass (for plotting IIRC)

je-cook (approved):

PR #2721, opened 2023-10-18
`Coil` remove control=False call

CoronelBuendia:
## Linked Issues

Closes #{ID}

## Description

:/ removing the control flag from Coil internals means that now when loading an EQDSK with passive coils, one has to manually assign those coils as non-controlled coils after initialisation.

## Checklist

### DIFF
--- a/bluemira/equilibria/coils/_grouping.py
+++ b/bluemira/equilibria/coils/_grouping.py
@@ -333,7 +333,6 @@ class CoilGroup(CoilGroupFieldsMixin):
                         dx=dx,
                         dz=dz,
                         ctype="NONE",
-                        control=False,
                     )
                 )
             elif dx != dz:  # Rough and ready


je-cook:
I'm a bit confused, passive coils are meant to be uncontrolled no?

CoronelBuendia:
> I'm a bit confused, passive coils are meant to be uncontrolled no?

Yep, but when loading an EQDSK with coils that are intended to be passive, we can't set them as anything other than controlled.

je-cook:
Oh yes sorry, this fixes the Coil init but it doesn't set the control status for those coils which would need some override in `CoilSet` and possibly a slight modification of the `to_group_vecs` function

CoronelBuendia:
> Oh yes sorry, this fixes the Coil init but it doesn't set the control status for those coils which would need some override in `CoilSet` and possibly a slight modification of the `to_group_vecs` function

Yeah.. that can be a "feature" down the line. It's not a huge loss of functionality compared to before, it's just that some codes (e.g. CREATE) also put passive structures in the EQDSK file as coils with almost 0 current, which I used to set as control=False. Now one just has to do:
```python
coilset.control = [c for c in coilset.control if "NONE" not in c]
```
after the fact

je-cook:
Something like this would fix the above I havent looked up how to do super with classmethods so that could be broken:

```python
class CoilSet(CoilGroup):
    ...
   @classmethod
    def from_group_vecs(cls, ...):
        self = super().from_group_vecs(...)
        self.control = self.get_coiltype(CoilType.NONE).name
        return self
```

je-cook (approved):

PR #2726, opened 2023-10-18
Fixed the passive coils so they no longer require manual setting

### LINKED ISSUE #2724: Re-add auto control=False for passive coils from eqdsk

Since the move of control to a coilset concept instead of a coil concept it has not been possible to automatically read in passive coils as uncontrolled from an EQDSK (and actually broken until #2721)

## Proposed solution

Something like this:

```python
class CoilSet(CoilGroup):
    ...
   @classmethod
    def from_group_vecs(cls, ...):
        self = super().from_group_vecs(...)
        self.control = self.get_coiltype(CoilType.NONE).name
        return self
```

_Originally posted by @je-cook in https://github.com/Fusion-Power-Plant-Framework/bluemira/issues/2721#issuecomment-1767906102_

## Alternative solutions

None considered

OceanNuclear:
## Linked Issues

Closes #2724, which stems from #2721.

## Description

Passive coils no longer require manually setting.

## Interface Changes

None, no API changes.

## Checklist

### DIFF
--- a/bluemira/equilibria/coils/_coil.py
+++ b/bluemira/equilibria/coils/_coil.py
@@ -65,6 +65,13 @@ class CoilType(Enum, metaclass=CoilTypeEnumMeta):
     DUM = auto()
     NONE = auto()
 
+    @classmethod
+    def get_actively_controlled_types(cls):
+        """Get the list of types which are actively controlled
+        (as opposed to passive coils).
+        """
+        return [cls.PF, cls.CS]
+
 
 class CoilNumber:
     """
--- a/bluemira/equilibria/coils/_coil.py
+++ b/bluemira/equilibria/coils/_coil.py
@@ -65,13 +65,6 @@ class CoilType(Enum, metaclass=CoilTypeEnumMeta):
     DUM = auto()
     NONE = auto()
 
-    @classmethod
-    def get_actively_controlled_types(cls):
-        """Get the list of types which are actively controlled
-        (as opposed to passive coils).
-        """
-        return [cls.PF, cls.CS]
-
 
 class CoilNumber:
     """
--- a/bluemira/equilibria/coils/_grouping.py
+++ b/bluemira/equilibria/coils/_grouping.py
@@ -945,11 +945,11 @@ class CoilSet(CoilSetFieldsMixin, CoilGroup):
                     - a boolean, which denotes all controlled vs none controlled.
         """
         names = self.name
-        if isinstance(control_names, List):  # List[str] case
+        if isinstance(control_names, List):
             self._control_ind = [names.index(c) for c in control_names]
-        elif control_names or control_names is None:  # bool case or None case
+        elif control_names or control_names is None:
             self._control_ind = np.arange(len(names)).tolist()
-        else:  # bool case
+        else:
             self._control_ind = []
         self._control = [names[c] for c in self._control_ind]
 
@@ -1012,16 +1012,18 @@ class CoilSet(CoilSetFieldsMixin, CoilGroup):
         return None
 
     @classmethod
-    def from_group_vecs(cls, eqdsk: EQDSKInterface):
+    def from_group_vecs(
+        cls, eqdsk: EQDSKInterface, control_coiltypes=(CoilType.PF, CoilType.CS)
+    ):
         """Create CoilSet from eqdsk group vectors.
-        
+
         Automatically sets all coils that are not implicitly passive to control coils
         """
         self = super().from_group_vecs(eqdsk)
 
         self.control = [
             coil.name
-            for active_coiltype in (CoilType.CS, CoilType.PF)
-            for coil in self._get_coiltype(active_coiltype)
+            for ctype in control_coiltypes
+            for coil in self._get_coiltype(ctype)
         ]
         return self
--- a/bluemira/equilibria/coils/_grouping.py
+++ b/bluemira/equilibria/coils/_grouping.py
@@ -935,7 +935,7 @@ class CoilSet(CoilSetFieldsMixin, CoilGroup):
 
     @control.setter
     def control(self, control_names: Optional[Union[List, bool]] = None):
-        """Set control coils"""
+        """Set which coils are actively controlled"""
         names = self.name
         if isinstance(control_names, List):
             self._control_ind = [names.index(c) for c in control_names]
@@ -946,7 +946,7 @@ class CoilSet(CoilSetFieldsMixin, CoilGroup):
         self._control = [names[c] for c in self._control_ind]
 
     def get_control_coils(self):
-        """Get Control coils"""
+        """Get control coils"""
         coils = []
         for c in self._coils:
             names = c.name
@@ -1014,7 +1014,11 @@ class CoilSet(CoilSetFieldsMixin, CoilGroup):
         """
         self = super().from_group_vecs(eqdsk)
 
-        if coil_name := self.get_coiltype(CoilType.NONE):
+        if (none_coil := self.get_coiltype(CoilType.NONE)) or (
+            dum_coil := self.get_coiltype(CoilType.DUM)
+        ):
             # if passive coils are present: give them the correct names.
-            self.control = coil_name.name
-        return self
+            self.control = [
+                *([] if none_coil is None else none_coil.name),
+                *([] if dum_coil is None else dum_coil.name),
+            ]
--- a/bluemira/equilibria/coils/_grouping.py
+++ b/bluemira/equilibria/coils/_grouping.py
@@ -934,14 +934,19 @@ class CoilSet(CoilSetFieldsMixin, CoilGroup):
         return self._control
 
     @control.setter
-    def control(self, control_names: Optional[Union[List, bool]] = None):
-        """Set which coils are actively controlled"""
+    def control(self, control_names: Optional[Union[List[str], bool]] = None):
+        """Set which coils are actively controlled
+        control names: can be one of the following:
+                    - list of str, each one being the name of each control coil.
+                    - None, for when ALL coils are control coils.
+                    - a boolean, which denotes all controlled vs none controlled.
+        """
         names = self.name
-        if isinstance(control_names, List):
+        if isinstance(control_names, List):  # List[str] case
             self._control_ind = [names.index(c) for c in control_names]
-        elif control_names or control_names is None:
+        elif control_names or control_names is None:  # bool case or None case
             self._control_ind = np.arange(len(names)).tolist()
-        else:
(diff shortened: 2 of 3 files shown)


bluemira/equilibria/coils/_grouping.py
```diff
@@ -515,7 +515,9 @@ def _get_coiltype(self, ctype):
 
     def get_coiltype(self, ctype: Union[str, CoilType]):
         """Get coil by coil type"""
-        return CoilGroup(*self._get_coiltype(ctype))
+        if coiltype := self._get_coiltype(ctype):
+            return type(self(*coiltype))
```
je-cook, line 519: Bracket in the wrong place it looks like
```suggestion
            return type(self)(*coiltype)
```

je-cook: Thinking about this (I know I suggested this) I think it may be better to add an override to `CoilSet` that looks like this:

```python
    # Coilset
    def get_coiltype(self, ctype): 
        """Get coils by coils type"""
        if coiltype := self._get_coiltype(ctype):
            return CoilSet(*coiltype) 
        return None
```
and this change
```suggestion
            return CoilGroup(*coiltype)
```
I reckon they're could be some hidden bugs when `Circuit` is used otherwise.

bluemira/equilibria/coils/_grouping.py
```diff
@@ -998,3 +996,18 @@ def _sum(
...
+    @classmethod
+    def from_group_vecs(cls, eqdsk: EQDSKInterface):
+        """Same as its parent class,
+        initialize an instance of CoilSet from group vectors,
+        but it also auto-populate the control attribute.
+
+        This is done so that the user doesn't have to
+        manually set passive coils as non-controlled.
+        """
+        self = super().from_group_vecs(eqdsk)
```
je-cook, line 1009: Can you double check for me that the type of self here is `Coilset`?  Possibly add a test to that effect

OceanNuclear: Change applied, test written.

OceanNuclear:
Brief note on the error: the following lines works when
```python
testfile = Path(get_bluemira_path("eqdsk", subfolder="data"), "jetto.eqdsk_out")
e = EQDSKInterface.from_file(testfile)
coil = grouping.from_group_vecs(e)
```
`grouping = CoilSet`; but fails with the error when `grouping = CoilGroup`.
This is because CoilSet has argument `control_names` but CoilGroup has no argument `control_names`.

tests/equilibria/test_equilibrium.py
```diff
@@ -374,16 +374,17 @@ def test_qpsi_calculation_modes(self):
...
-        assert coilset.b_max.any() == 0
-        assert coilset.n_coils(ctype="DUM") == 4
-        assert len(coilset.control) == 0
+        coil = grouping.from_group_vecs(e)
+        assert isinstance(coil, grouping), "Check classmethod is making the right class"
+        assert coil.current.any() == 0
+        assert coil.j_max.any() == 0
+        assert coil.b_max.any() == 0
+        assert coil.n_coils(ctype="DUM") == 4
+        assert len(coil.control) == 0
```
je-cook, line 387:
```suggestion
        if grouping is CoilSet:
            assert len(coil.control) == 0
```

je-cook (approved):
Here's the fix to the test, otherwise looks good!

tests/equilibria/test_equilibrium.py
```diff
@@ -329,7 +329,7 @@ def test_betapliip_profile(self, shape):
             self.B_0,
             shape=shape,
             li_min_iter=0,
-            li_rel_tol=rel_tol,
+            li_reltol=rel_tol,
```
je-cook, line 332: oh this should be changed back
```suggestion
            li_rel_tol=rel_tol,
```

bluemira/equilibria/coils/_grouping.py
```diff
@@ -998,3 +996,25 @@ def _sum(
...
+
+        if coil_name := self.get_coiltype(CoilType.NONE):
+            # if passive coils are present: give them the correct names.
+            self.control = coil_name.name
```
je-cook, line 1019: We need another line for dummy coils
```suggestion
        if (none_coil := self.get_coiltype(CoilType.NONE)) or (
            dum_coil := self.get_coiltype(CoilType.DUM)
        ):
            # if passive coils are present: give them the correct names.
            self.control = [
                *([] if none_coil is None else none_coil.name),
                *([] if dum_coil is None else dum_coil.name),
            ]
            import
```

OceanNuclear: Should be the negation of this. Will implement accordingly.

OceanNuclear:
Just wondering @je-cook, why do we want to [return None](https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/ocean/control-fix/bluemira/equilibria/coils/_grouping.py#L520) instead of an empty list though?

OceanNuclear:
I hid the list of active coils away into get_actively_controlled_types() so that it only appears explicitly in ._coils.py (inside class CoilType).

je-cook:
> Just wondering @je-cook, why do we want to [return None](https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/ocean/control-fix/bluemira/equilibria/coils/_grouping.py#L520) instead of an empty list though?

Because a list is not the original return type (which is a coil group) and you cannot have an empty coilgroup. Therefore if there aren't any coils to select we dont return anything which is the same as returning None.

bluemira/equilibria/coils/_coil.py
```diff
@@ -65,6 +65,13 @@ class CoilType(Enum, metaclass=CoilTypeEnumMeta):
     DUM = auto()
     NONE = auto()
 
+    @classmethod
+    def get_actively_controlled_types(cls):
+        """Get the list of types which are actively controlled
+        (as opposed to passive coils).
+        """
+        return [cls.PF, cls.CS]
+
```
je-cook, line 74: I know where you're coming from but I don't think this is a good idea. This implicitly restricts the user and will only be used in eqdsk read in. 

Remove please

bluemira/equilibria/coils/_grouping.py
```diff
@@ -514,8 +514,10 @@ def _get_coiltype(self, ctype):
         return coils
 
     def get_coiltype(self, ctype: Union[str, CoilType]):
-        """Get coil by coil type"""
-        return CoilGroup(*self._get_coiltype(ctype))
+        """Get coils matching coil type"""
+        if coiltype := self._get_coiltype(ctype):  # if matched coils list is not empty
+            return CoilGroup(*coiltype)
+        return None  # Don't return anything.
```
je-cook, line 520: The comments are not needed here
```suggestion
        if coiltype := self._get_coiltype(ctype):
            return CoilGroup(*coiltype)
        return None
```

bluemira/equilibria/coils/_grouping.py
```diff
@@ -932,19 +934,24 @@ def control(self) -> List:
...
+        """
         names = self.name
-        if isinstance(control_names, List):
+        if isinstance(control_names, List):  # List[str] case
             self._control_ind = [names.index(c) for c in control_names]
-        elif control_names or control_names is None:
+        elif control_names or control_names is None:  # bool case or None case
             self._control_ind = np.arange(len(names)).tolist()
-        else:
+        else:  # bool case
```
je-cook, line 949: Or here
```python
        if isinstance(control_names, List): 
            self._control_ind = [names.index(c) for c in control_names]
        elif control_names or control_names is None:
            self._control_ind = np.arange(len(names)).tolist()
        else:
```

bluemira/equilibria/coils/_grouping.py
```diff
@@ -998,3 +1001,26 @@ def _sum(
...
+            return CoilSet(*coiltype)
+        return None
+
+    @classmethod
+    def from_group_vecs(cls, eqdsk: EQDSKInterface):
+        """Same as its parent class,
+        initialize an instance of CoilSet from group vectors,
+        but it also automatically decide which coils are control coils
+        according to CoilType.get_actively_controlled_types.
+        """
```
je-cook, line 1017: This is too verbose. See suggestion, remember this should be able to be read in isolation.
```suggestion
        """Create CoilSet from eqdsk group vectors.
        
        Automatically sets all coils that are not implicitly passive to control coils
        """
```

bluemira/equilibria/coils/_grouping.py
```diff
@@ -998,3 +1001,26 @@ def _sum(
...
+        self = super().from_group_vecs(eqdsk)
+
+        controlled_coils = []
+        for active_coiltype in CoilType.get_actively_controlled_types():
+            controlled_coils.extend(
+                [coil.name for coil in self._get_coiltype(active_coiltype)]
+            )
+        self.control = controlled_coils
```
je-cook, line 1025: Its only setting the one variable really so I prefer inlining the assignment (list comprehension reverses the order of the loops)
```suggestion
        self.control = [
            coil.name
            for active_coiltype in (CoilType.CS, CoilType.PF)
            for coil in self._get_coiltype(active_coiltype)
        ]
```

bluemira/equilibria/coils/_grouping.py
```diff
@@ -932,19 +934,24 @@ def control(self) -> List:
...
     @control.setter
-    def control(self, control_names: Optional[Union[List, bool]] = None):
-        """Set control coils"""
+    def control(self, control_names: Optional[Union[List[str], bool]] = None):
+        """Set which coils are actively controlled
+        control names: can be one of the following:
+                    - list of str, each one being the name of each control coil.
+                    - None, for when ALL coils are control coils.
+                    - a boolean, which denotes all controlled vs none controlled.
+        """
```
je-cook, line 943: slight layout adjustment
```suggestion
        """Set which coils are actively controlled

        Parameters
        ----------
        control_names:
                    - list of str, each one being the name of each control coil.
                    - None, for when ALL coils are control coils.
                    - a boolean, which denotes all controlled vs none controlled.
        """
```

je-cook (commented):
Well done on getting the tests to pass. These are all stylistic.

bluemira/equilibria/coils/_coil.py
OceanNuclear: Yep, understood, and will delete.

OceanNuclear:
Thank you, the double list-comprehension was useful, I didn't know it can work that way. I've deleted all the unwanted comments and the classmethod inside CoilType(Enum), and all equilibria tests should pass now.

PR #2730, opened 2023-10-19
🧪 Add closed equilibria shape CAD tests

je-cook:
## Description

We aren't consistent in our Johner variable naming (psi or phi), I've picked phi because it seemed to be the original.

Our Johner CAD parameterisations are not closed (from the earlier tests). It seems to be that this has been the case for a while? 

## Checklist


CoronelBuendia:
I agree `phi` is a better name; does it not break API though?

je-cook:
yeah it does, that and the CAD not being closed are why its still draft

PR #2740, opened 2023-10-23
Add PROCESS API example

CoronelBuendia:
## Linked Issues

Closes #{ID}

## Checklist

### DIFF
--- a/examples/codes/run_process_example.ex.py
+++ b/examples/codes/run_process_example.ex.py
@@ -1,3 +1,19 @@
+# ---
+# jupyter:
+#   jupytext:
+#     cell_metadata_filter: tags,-all
+#     notebook_metadata_filter: -jupytext.text_representation.jupytext_version
+#     text_representation:
+#       extension: .py
+#       format_name: percent
+#       format_version: '1.3'
+#   kernelspec:
+#     display_name: Python 3 (ipykernel)
+#     language: python
+#     name: python3
+# ---
+
+# %% tags=["remove-cell"]
 # bluemira is an integrated inter-disciplinary design tool for future fusion
 # reactors. It incorporates several modules, some of which rely on other
 # codes, to carry out a range of typical conceptual fusion reactor design
--- a/examples/codes/run_process_example.ex.py
+++ b/examples/codes/run_process_example.ex.py
@@ -345,10 +345,17 @@ template_builder.add_input_value("n_pancake", 20)
 
 
 # %%[markdown]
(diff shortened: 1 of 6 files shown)


examples/codes/run_process_example.ex.py
```diff
@@ -0,0 +1,376 @@
...
+
+# %%
+template_builder.add_input_value("qnuc", 1.3e4)
+template_builder.add_input_value("n_layer", 20)
+template_builder.add_input_value("n_pancake", 20)
+
+
+# %%[markdown]
+# Finally, let us run PROCESS with our inputs:
+
```
je-cook, line 352: Could we add a sentence here about why the first argument is `{}` and link to eg:

```python
#  [External code example](../external_code.ex.py)
```

examples/codes/run_process_example.ex.py
```diff
@@ -0,0 +1,376 @@
...
+# one of these as an iteration variable.
+# Note that the PROCESS term for the impurity concentrations is the `fimp`
+# array, for which sadly one must still know the integer in order to specify
+# it as an iteration variable. 13 is Xenon.
+
+# %%
+template_builder.add_impurity(Impurities.H, 1.0)
+template_builder.add_impurity(Impurities.He, 0.1)
+template_builder.add_impurity(Impurities.W, 5.0e-5)
+template_builder.add_variable("fimp(13)", 3.573e-04)
```
je-cook, line 200: Need to check it works but this, If not we could fix the string produced to add the close bracket or something. There must be a reason I didnt close the bracket originally...: 

```suggestion
template_builder.add_variable(Impurities.Xe.id(), 3.573e-04)
```

je-cook (commented):
On the whole looks good. Could you also add the below to the examples CI as we don't install process atm eg `-e 'codes/run_process'`:

https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/bdde3af24fb31f215d9588947c273c044cc28042/.github/workflows/examples.yml#L40

examples/codes/run_process_example.ex.py
je-cook: ahh looks like in old MFILEs they hadn't added the closing bracket and in new ones they have so if the id function is fixed all of the above will work just need to close the bracket here: 
https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/bdde3af24fb31f215d9588947c273c044cc28042/bluemira/codes/process/api.py#L157

PR #2750, opened 2023-10-23
✨ Allow different in and out names for external codes

### LINKED ISSUE #1242: Enable external codes with in and out variables to map back to a singular variable

`P_el_net` and `P_el_net_process` are essentially the same variable but mapped to an in and out variable for process. It would be nice if we could combine these as our `send/recv` mechanics can manage where the variable is stored. I'm pretty sure that if a variable isn't received back from a solver it is still stored on the solver.

I can see in future more codes having different variables for in and out and it would be nice to contain these 

## Proposed solution

A `send_name` mapping and `recv_name` mapping. By default they would be the same thing.

## Alternative solutions

Have two variables as is the current solution

## Additional context

There may be contexts where the out variable is not as accurate as required but that should be managed by `send/recv`

je-cook:
## Linked Issues

Closes #1242 

## Description

The added optional `out_name` for mappings allows different names to be in and out variables for an external code. 

## Interface Changes

`out_name` optional extra input for `ParameterMapping`

## Checklist


bluemira/base/parameter_frame/_frame.py
```diff
@@ -406,8 +406,18 @@ def _validate_units(param_data: Dict, value_type: Iterable[Type]):
...
-        quantity = pint.Quantity(f'{param_data["value"]}*{param_data["unit"]}')
-        param_data["value"] = quantity.magnitude
+        try:
+            quantity = pint.Quantity(f'{param_data["value"]}*{param_data["unit"]}')
+        except pint.errors.PintError as pe:
+            if param_data["value"] is None:
+                quantity = pint.Quantity(
+                    1 if param_data["unit"] in (None, "") else param_data["unit"]
+                )
+                param_data["source"] = f"{param_data.get('source', '')}\nMAD UNIT 🤯 😭:"
```
CoronelBuendia, line 416: I worry the emoticons will cause `ParameterFrame` `__repr__` and printing to break...

je-cook: they dont, they will just print as undefined squares if you dont have utf8 support, we already have one somewhere else in the code

CoronelBuendia (approved):

PR #2770, opened 2023-10-30
Correct type of names for ParameterFrame.get_values

bielsnohr:
## Linked Issues

Very minor fix. Didn't think it needed an issue, but can create one if you like.

## Description

The type annotation for the `names` argument of `ParameterFrame.get_values` should be `List[str]` but is currently just `str`.

## Interface Changes

N/A

## Checklist

- [ ] ~~Tests run locally and pass `pytest tests --reactor`~~
- [ ] ~~Documentation built locally and checked `sphinx-build -W documentation/source documentation/build`~~

### DIFF
--- a/bluemira/base/parameter_frame/_frame.py
+++ b/bluemira/base/parameter_frame/_frame.py
@@ -157,7 +157,7 @@ class ParameterFrame:
             except TypeError:
                 self.update_values(new_values)
 
-    def get_values(self, *names: List[str]) -> Tuple[ParameterValueType, ...]:
+    def get_values(self, *names: Tuple[str]) -> Tuple[ParameterValueType, ...]:
         """Get values of a set of Parameters"""
         try:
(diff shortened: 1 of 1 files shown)


bielsnohr:
Ah, might more correctly be `Tuple[str]`, I'll change now.

je-cook:
I just checked and I think its right according to PEP484:
 https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

bielsnohr:
Ah, this is my error. I was passing a tuple when it should just be a bare argument list. Therefore, was getting mypy complaints, which I erroneously thought were because of the type hint. I'll close this PR as it is not needed.

je-cook:
Thanks for the PR though. There are many broken typings elsewhere and any PR fixing them is greatly appreciated!

PR #2792, opened 2023-11-10
✨ Use Freecad from apt

je-cook:
## Description

This only works with the python version apt uses which currently is 3.10. It is quite a lot of hacking of paths

## Checklist


je-cook:
I'm tempted to merge this (after review) as all the tests pass because we're using the conda freecad. Thoughts?

je-cook:
As we're adding the option of CADQuery soon in #4298 this is no longer a direction we will explore. FreeCAD doesnt seem to have the resources to put effort into the conda direction understandably prefering the GUI

PR #2873, opened 2023-12-12
Powercycle layout base

je-cook:
## Description

This creates a powercycle configuration reader and constructs a phase and loads object on demand.
The json file contains the below top level keys;

```  
  "scenario"
  "pulse_library"
  "phase_library"
  "subphase_library"
  "system_library"
  "sub_system_library"
  "load_library"
```
the high level structure is as follows: 
- a `scenario` contains `pulses` which is a list of pulse names that are keys in `pulse_library`
- a `pulse` contains `phases` which is a list of phase names that are keys in `phase_library`
- a `phase` contains `subphase` which is a list of phase names which are keys in `subphase_library`
- `system_library` contains `systems` which contain subsystems which is keys in `sub_system_library`
- `sub_system_library` contains `loads` 
- `subphases` contain the link to the `load_library

image

For other mechanics please look at the example. The configuration object allows you to expand the load list and you can replace `duration`s that are strings in the config with values using `import_subphase_duration`.

In essence the three interesting objects are the following classes: `Loads`,`Phase`, `PowerCycleLibraryConfig` and `PowerCycleLoadConfig` which should be the main entry points a user should need. In its simplest for they will probably only ever directly use `Phase` and `PowerCycleLibraryConfig` with the current setup.

In future the config could be split into multiple json files instead of one monolith. I would suggest that the split should only be at the top level keys because otherwise overwriting of various things would be hard to catch. Either way for a future PR.

@tplobo, @oliverfunk please comment as you see fit!

## Checklist


bluemira/power_cycle/net.py
```diff
@@ -0,0 +1,660 @@
...
+                for k, v in data["sub_system_library"].items()
+            },
+            loads={
+                k: PowerCycleLoadConfig(name=k, **v)
+                for k, v in data["load_library"].items()
+            },
+        )
+
+    def make_phase(self, phase: str, *, check=True) -> Phase:
+        """Create a single phase object"""
```
oliverfunk, line 629:
```suggestion
    def get_phase(self, phase: str, *, check=True) -> Phase:
```

bluemira/power_cycle/net.py
```diff
@@ -0,0 +1,660 @@
...
+            subphases,
+            Loads({
+                ld: self.loads[ld]
+                for subphase in subphases.values()
+                for ld in subphase.loads
+            }),
+        )
+
+    def make_pulse(self, pulse: str, *, check=True) -> Dict[str, Phase]:
+        """Create a pulse dictionary"""
```
oliverfunk, line 646:
```suggestion
    def get_pulse(self, pulse: str, *, check=True) -> Dict[str, Phase]:
```

bluemira/power_cycle/net.py
```diff
@@ -0,0 +1,660 @@
...
+        if check:
+            self.check_config()
+        return {
+            phase: self.make_phase(phase, check=False)
+            for phase in self.pulse[pulse].phases
+        }
+
+    def make_scenario(self) -> Dict[str, Dict[str, Union[float, Dict[str, Phase]]]]:
+        """Create a scenario dictionary"""
```
oliverfunk, line 655:
```suggestion
    def get_scenario(self) -> Dict[str, Dict[str, Union[float, Dict[str, Phase]]]]:
```

Also maybe break the return type up into type alliases?
It's pretty difficult to see where the float type comes from, i.e. either float or dict[str, phase]

bluemira/power_cycle/net.py
oliverfunk: And ofc, refactor all subsequent calls.

bluemira/power_cycle/net.py
```diff
@@ -0,0 +1,660 @@
...
+    phase: LibraryConfigDescriptor = LibraryConfigDescriptor(config=PhaseConfig)
+    subphase: LibraryConfigDescriptor = LibraryConfigDescriptor(
+        config=PowerCycleSubPhase
+    )
+    system: LibraryConfigDescriptor = LibraryConfigDescriptor(config=PowerCycleSystem)
+    subsystem: LibraryConfigDescriptor = LibraryConfigDescriptor(
+        config=PowerCycleSubSystem
+    )
+    loads: LibraryConfigDescriptor = LibraryConfigDescriptor(config=PowerCycleLoadConfig)
+
```
oliverfunk, line 539:
```suggestion
class PowerCycleLibraryConfig:
    """Power Cycle Configuration"""

    def __init__(
        self,
        scenario: ScenarioConfig,
        pulse: Dict[str, PulseConfig],
        phase: Dict[str, PhaseConfig],
        subphase: Dict[str, PowerCycleSubPhase],
        system: Dict[str, PowerCycleSystem],
        subsystem: Dict[str, PowerCycleSubSystem],
        loads: Dict[str, PowerCycleLoadConfig],
    ):
        self.scenario = scenario
        self.pulse = pulse
        self.phase = phase
        self.subphase = subphase
        self.system = system
        self.subsystem = subsystem
        self.loads = loads

        self.check_config()
```

bluemira/power_cycle/net.py
```diff
@@ -0,0 +1,660 @@
...
+                raise ValueError(f"Unknown loads {unknown_load}")
+
+    def import_subphase_duration(self, subphase_duration_params):
+        """Import subphase data"""
+        for s_ph in self.subphase.values():
+            if isinstance(s_ph.duration, str):
+                s_ph.duration = getattr(
+                    subphase_duration_params, s_ph.duration.replace("$", "")
+                )
+
```
oliverfunk, line 578:
```suggestion
    def set_subphase_durations(self, duration_values: Dict[str, float]):
        """Import subphase data"""
        for s_ph in self.subphase.values():
            if isinstance(s_ph.duration, str):
                # force the lookup as it should fail if not found
                # i.e. all durations should be imported
                s_ph.duration = duration_values[s_ph.duration.replace("$", "")]
```

This could be made private and called during the __init__ where duration_values could be optionally provided.

Could make it ensure that all str durations are set with values during the init of the config

bluemira/power_cycle/net.py
```diff
@@ -0,0 +1,660 @@
...
+    @property
+    def duration(self):
+        """Duration of phase"""
+        data = [s_ph.duration for s_ph in self.subphases.values()]
+        try:
+            return getattr(np, self.config.operation)(data)
+        except np.core._exceptions.UFuncTypeError as e:
+            raise TypeError(f"duration variables have not been imported {data}") from e
+
```
oliverfunk, line 522: This kinda becomes unnec. if the durations check (see comment) is done during init of PowerCycleLibraryConfig

bluemira/power_cycle/net.py
```diff
@@ -0,0 +1,660 @@
...
+class PhaseConfig(Config):
+    """Power cycle phase config"""
+
+    operation: str
+    subphases: list[str]
+    description: str = ""
+
+
+@dataclass
+class PowerCycleSubPhase(Config):
```
oliverfunk, line 221: Add Config to name, remove PowerCycle

bluemira/power_cycle/net.py
```diff
@@ -0,0 +1,660 @@
...
+    """Data class property descriptor
+
+    See https://docs.python.org/3/library/dataclasses.html#descriptor-typed-fields
+    """
+
+    def __set_name__(self, _, name: str):
+        """Set the attribute name from a dataclass"""
+        self._name = f"_{name}"
+
+
```
oliverfunk, line 103: Can delete these class Desc. classes

bluemira/power_cycle/net.py
je-cook: I haven't done the check config at the end of the init because I can see a situation where eg the load hasnt been added yet. The check is still done on the get_* functions

bluemira/power_cycle/net.py
```diff
@@ -0,0 +1,644 @@
...
+            self.active_data = np.zeros_like(self.time)
+
+        for var_name in ("time", "reactive_data", "active_data"):
+            var = getattr(self, var_name)
+            if not isinstance(var, np.ndarray):
+                setattr(self, var_name, np.asarray(var))
+        if isinstance(self.model, str):
+            self.model = LoadModel[self.model.upper()]
+        for data in (self.reactive_data, self.active_data):
+            if data.size != self.time.size:
```
oliverfunk, line 243: Should len() be used here instead of .size?

bluemira/power_cycle/net.py
```diff
@@ -0,0 +1,644 @@
...
+    ):
+        self.scenario = scenario
+        self.pulse = pulse
+        self.phase = phase
+        self.subphase = subphase
+        self.system = system
+        self.subsystem = subsystem
+        self.loads = loads
+        if durations is not None:
+            self._import_subphase_duration(durations)
```
oliverfunk, line 512: The only problem with this is if durations aren't given and there are str's in the duration params, they won't be checked and will error eventaully when called.

IMO it should be something like, check if any duration is a str value, if so get the value from the durations dict (error if not there or durations dict isn't given) and set it

oliverfunk (commented):
Think it's good now, two small comments

bluemira/power_cycle/net.py
je-cook: no because that wont catch if for whatever reason data is a 2d array, size will tell me the total number of elements not the length of the first axis.

oliverfunk (approved):

examples/power_cycle/simple_example.ex.py
```diff
@@ -0,0 +1,91 @@
...
+
+phase = config.get_phase("dwl")
+
+normalised_time = interpolate_extra(phase.loads.build_timeseries(), 3)
+timeseries = normalised_time * phase.duration
+
+active_loads = phase.loads.get_load_data_with_efficiencies(
+    normalised_time, "active", "MW"
+)
+active_load_total = phase.loads.load_total(normalised_time, "active", "MW")
```
tplobo, line 86: It is currently a bit complicated to extract/compute/plot only consumption/production loads and their individual sums. The basic feature I expect of the `power_cycle` module is to get me those. We should think about generic `get_contributions` and `get_total` methods that provide collections or sums of collections of loads that contain certain characteristics/specifiers/traits.

We can later develop that into those "Tags" that we had discussed, to be more flexible (_e.g._ when one wants to list and sum all loads that have the `consumption`, `active` and `heating &current drive` tags; note that the `active` tag,for example, could be automatically added depending on associated `bool` in the `LoadConfig` instance).

```suggestion
active_loads = phase.loads.get_load_data_with_efficiencies(
    normalised_time, "active", "MW"
)
active_load_total = phase.loads.load_total(normalised_time, "active", "MW")

reactive_loads = phase.loads.get_load_data_with_efficiencies(
    normalised_time, "reactive", "MW"
)
reactive_load_total = phase.loads.load_total(normalised_time, "reactive", "MW")

all_loads = phase.loads
active_interpolated_loads = all_loads.get_interpolated_loads(
    normalised_time, "active", "MW"
)
pp(active_interpolated_loads)
active_consumption_loads = {
    ld.name: active_interpolated_loads[ld.name] if ld.consumption else None
    for ld in all_loads.loads.values()
}
pp(active_consumption_loads)
for key, load in active_consumption_loads.items():
    pp(key)
    pp(load)
    if load is not None:
        plt.plot(normalised_time, load, label=key)
plt.plot(normalised_time, active_load_total, "--", label="total")
plt.legend()
plt.grid()
plt.show()

```

tplobo (commented):
Here are some of the comments I had added yesterday before our talk today. You can use them as reminders for the points we discussed.

examples/power_cycle/scenario_config.json
```diff
@@ -0,0 +1,422 @@
...
+          "site",
+          "sf6_air"
+        ],
+        "efficiencies": {
+          "hcpb_bb_peak": [
+            {
+              "value": 5,
+              "description": "coolant flow to 20%",
+              "reactive": true
+            },
```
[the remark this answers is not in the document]
je-cook: Discussed externally, current setup is easier to follow

examples/power_cycle/scenario_config.json
```diff
@@ -0,0 +1,422 @@
...
+            {
+              "value": 5,
+              "description": "coolant flow to 20%",
+              "reactive": false
+            }
+          ],
```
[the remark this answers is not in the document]
je-cook: Discussed elsewhere. This opens us up to some very confused users. Even though technically the same. 

Specifically here this is for testsing that this works

bluemira/power_cycle/net.py
```diff
@@ -0,0 +1,648 @@
...
+        *(
+            np.linspace(vector[s], vector[s + 1], n_points + 1, endpoint=False)
+            for s in range(len(vector) - 1)
+        ),
+        np.atleast_1d(vector[-1]),
+    ])
+
+
+class Loads:
+    """Loads of a phase"""
```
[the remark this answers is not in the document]
je-cook: Discussed elsewhere summary:

If a user wants to add a load to the config they are interfacing with the config. They are not at the same time dealing with the resultant phase and loads on that phase.

If in future we do generalise its best to do it while creating that usecase instead of trying to make it too general too soon.

On the whole we shouldnt need to implement basic mathematical operations on classes. There needs to be a very good reason for this and adding simple arrays together on the surface doesnt seem to be one of them

examples/power_cycle/simple_example.ex.py
```diff
@@ -0,0 +1,91 @@
...
+
+reactive_loads = phase.loads.get_load_data_with_efficiencies(
+    normalised_time, "reactive", "MW"
```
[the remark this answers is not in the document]
je-cook: technically maybe but for numeric purposes it makes no odds. I've added 'var' (lower case is the standard according to wikipedia) to the units in our pint registry and I will specify some text in the example to that effect.

As a side note VA already exists and is also numerically equivalent to W

examples/power_cycle/simple_example.ex.py
je-cook: there is now a consumption boolean option on most functions to explicitly get one or the other if specified

bluemira/power_cycle/net.py
```diff
@@ -0,0 +1,648 @@
...
+
+    def __init__(
+        self,
+        loads: Dict[str, LoadConfig],
```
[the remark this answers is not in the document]
je-cook: loads is now a private variable and the class is now called LoadSet

examples/power_cycle/scenario_config.json
```diff
@@ -0,0 +1,422 @@
...
+        "time": [0, 1],
+        "reactive_data": [4.7, 4.7],
+        "active_data": [9.7, 9.7],
+        "unit": "MW",
+        "model": "RAMP",
+        "normalised": true,
+        "consumption": true,
+        "efficiencies": [],
+        "description": "Vacuum Vessel (VV)"
+      },
```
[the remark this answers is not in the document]
je-cook: done

examples/power_cycle/simple_example.ex.py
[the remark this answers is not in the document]
je-cook: they weren't but I've simplified use

examples/power_cycle/scenario_config.json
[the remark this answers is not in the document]
tplobo: I still think we should make efficiency values independent from `consumption` / `production`. Even though it makes sense from a physics point of view, the `consumption` aspect is not directly visible within the `efficiencies` part of the JSON input file and is cumbersome to remember.

tplobo:
> @tplobo I havent included the unique domains additions as I think that warrants its own PR after this has gone in because of all the possible pitfalls.

That's OK, but against your [last argument here](https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/2873#discussion_r1469237929), just remember that ensuring a Load/Timeseries has a function domain (_i.e._ unique elements in X for the (X,Y) vector pair) is one of the (non-trivial) aspects that motivate us to create such a "prototype" class, in addition to addition and subtraction methods.

examples/power_cycle/simple_example.ex.py
tplobo: This is a good quick solution, but I suggest the following modification for this option:
- `consumption = None`: uses/computes/adds both consumption and production loads
- `consumption = True`: only uses/computes/adds consumption loads
- `consumption = False`: only uses/computes/adds production loads

otherwise we can only separately gather consumption loads, but not production loads.

examples/power_cycle/simple_example.ex.py
```diff
@@ -0,0 +1,90 @@
...
+
+config.add_load_config(
+    LoadConfig(
+        "cs_power",
+        data={"active": [1, 2], "reactive": [10, 20]},
+        efficiencies=[Efficiency(0.1)],
+        description="something made up",
+    ),
+    ["cru", "bri"],
+    [Efficiency({"reactive": 0.2})],
```
tplobo, line 68: Just small modification: I reiterate my comment on adding parameter names explicitly to make example easier to read.

examples/power_cycle/simple_example.ex.py
je-cook: exactly whats implemented :)

bluemira/power_cycle/net.py
```diff
@@ -0,0 +1,870 @@
...
+        load_check = LoadType.check(load_type)
+        for ld_name in data:
+            load_conf = self._loads[ld_name]
+            for eff in load_conf.efficiencies:
+                for eff_type, eff_val in eff.value.items():
+                    if eff_type in load_check:
+                        c_eff = (
+                            1 / eff_val
+                            if load_conf.consumption and eff.value > 0
+                            else eff_val
```
tplobo, line 427: I come back to the point I mentioned in the general comments and during our discussion: perhaps an implementation independent of the `consumption` input is easier to understand/remember. Production values being lowered and consumption values being raised by efficiencies makes sense from the physical point of view, but the input file will become confusing.

examples/power_cycle/simple_example.ex.py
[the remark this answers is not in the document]
tplobo: Much better now!

tplobo: Ah! I had not understood the set output in `_consumption_flag`:
```
return {True, False} if consumption is None else {consumption}
```
I didn't know a flag could work with multiple values like that, that's neat!

bluemira/power_cycle/net.py
je-cook: Done

examples/power_cycle/simple_example.ex.py
je-cook: Done

tplobo (approved):
I think all's good now!

Issue #2890, opened 2023-12-20
Remove the only place where eV is used


OceanNuclear:
## Describe the bug

Following up on issue #2857:
bluemira.codes.process._solver.get_species_data is the only piece of code where temperature is given in units of eV is instead of keV or K.

## Expected behaviour

get_species_data returns `t_ref` in keV instead of eV.

## Suggested fix

```python
t_ref_keV = raw_uc(np.array(next(t_ref).data, dtype=float), "eV", "keV")
return t_ref_keV, np.array(next(lz_ref).data, dtype=float), np.array(next(z_av_ref).data, dtype=float)
```

## Additional context

This change would be API-breaking, thus must be approved by all users before merging into the develop branch. @DarioV86

PR #2891, opened 2023-12-20
Fix units and unit-conversion factors

### LINKED ISSUE #2853: Papercut: Remove *_TO_* constants from bluemira.base.constants
## Describe the papercut

There are 5 variables:
1. `EV_TO_J`
2. `J_TO_EV`
3. `AMU_TO_KG`
4. `YR_TO_S`
5. `S_TO_YR`

1-3 are used in bluemira.plasma_physics while 4-5 are used in bluemira.fuel_cycle.
Where these constants are used multiplicatively, e.g. `(e_dt * EV_TO_J)`, they can be replaced by `raw_uc(e_dt, 'eV', 'J')` for easier comprehension.

## Suggested fix/Desired result

No more `*_TO_*` constants left in bluemira.base.constants. `raw_uc`; Manually find-and-replace all instances of the variables A_TO_B above and replace with raw_uc(..., 'A', 'B'), so that they become more readable/ easier to debug.

## Alternative fix
N/A

## Estimated difficulty

Easy

Variable 1-3 were only used a maximum of twice per file, so only minimal change would be required, and should result in reduction in number of lines of code (fewer imports).

Variable 4-5 were used many more times in fuel cycle, but it should also be an easy find-and-replace job.

### LINKED ISSUE #2857: Temperature given in eV instead of keV or K in 3 files.
## Describe the papercut

When plasma temperature is specified in non-SI unit, it's almost always given as [keV]; except for 3 files:

- bluemira.codes.process._solver.get_species_data outputs temperature `tref` in eV.
- bluemira.plasma_physics.rules_of_thumb.estimate_loop_voltage takes in electron temperature on axis `T_e` in eV.
- bluemira.plasma_physics.collisions.spitzer_conductivity takes in electron temperature on axis `T_e` in eV.

## Suggested fix
Convert all of the above from eV to keV instead, using bluemira.base.constants.raw_uc when appropriate.

## Other modules affected/ dependencies
No other modules affected.
bluemira.plasma_physics.rules_of_thumb.estimate_loop_voltage uses bluemira.plasma_physics.collisions.spitzer_conductivity.

OceanNuclear:
## Linked Issues

Closes #2853 with caveats: YR_TO_S and S_TO_YR cannot be removed completely as bluemira/plasma_physics/tools.py uses them, and cannot be converted to raw_uc there due to numba.jit.

Closes #2857 with caveats: get_species_data needs to be overhauled if this change were to be made, so it has been left unchanged, and may be ultimately solved in issue #2890.

## Description

Solves issue #2853 and #2857 simultaneously, so all temperature units are in keV or K (as opposed to a mix of eV/keV/K); and no more conversion multiplies (`*_TO_*`) are required.

## Interface Changes

bluemira.plasma_physics.rules_of_thumb.estimate_loop_voltage now takes in electron temperature in keV instead of eV.
bluemira.plasma_physics.collisions.spitzer_conductivity now takes in electron temperature in keV instead of eV.

## Checklist

### DIFF
--- a/bluemira/base/constants.py
+++ b/bluemira/base/constants.py
@@ -62,7 +62,11 @@ class BMUnitRegistry(UnitRegistry):
         Add new contexts to registry
         """
         if not self._contexts_added:
-            self.contexts = [self._energy_temperature_context(), self._flow_context()]
+            self.contexts = [
+                self._energy_temperature_context(),
+                self._mass_energy_context(),
+                self._flow_context(),
+            ]
 
             for c in self.contexts:
                 self.add_context(c)
@@ -103,6 +107,31 @@ class BMUnitRegistry(UnitRegistry):
             e_to_t,
             t_units,
             ev_units,
+            lambda _, x: x * conversion,
+            lambda _, x: x / conversion,
+        )
+
+    def _mass_energy_context(self):
+        """
+        Converter between mass and energy
+
+        energy = mass * speed-of-light^2
+
+        Returns
+        -------
+        pint context
+        """
+        m_to_e = Context("Mass_to_Energy")
+
+        m_units = "[mass]"
+        e_units = "[energy]"
+
+        conversion = self.Quantity("c^2")
+
+        return self._transform(
+            m_to_e,
+            m_units,
+            e_units,
             lambda _, x: x * conversion,
             lambda _, x: x / conversion,
         )
@@ -248,9 +277,6 @@ ANGLE_UNITS = [
 # Physical constants
 # =============================================================================
 
-# Speed of light
-C_LIGHT = ureg.Quantity("c").to_base_units().magnitude  # [m/s]
-
 # Vacuum permeability
 MU_0 = ureg.Quantity("mu_0").to_base_units().magnitude  # [T.m/A] or [V.s/(A.m)]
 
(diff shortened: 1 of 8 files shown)


OceanNuclear:
(Manually checked and confirmed that this change does not lead to conflict in any other branches)

je-cook:
I'm only starting going through this one. My first thought is that the reduction in the use of the S_TO_YR and YR_TO_S will increase the number of calculations that have to be undertaken. In one instance as the conversion happens within a for loop its a many times increase.

Most of the other conversions are fairly sensible but I think there is no need to reduce so drastically the time conversions especially as its so widely used

OceanNuclear:
I thought the number of calculations would be the same across 
```python
raw_uc(delta_t, "yr", "s")
```
and 
```python
delta_t * YR_TO_S
```
? As raw_uc does the multiplication under the hood, plus a few other unit lookup calls inside pint.

The only place where this might be a significant a problem is in the for-loop inside EUDEMOFuelCycleModel.tbreed, but I thought that's not a function needing optimization because it re-calculated the temporary variable `(YR_TO_S * dt)` four times in a row, so I thought my method of initializing the local variable `dts` would still be faster as it removes the need for 3 multiplications.

je-cook:
I agree with your second point but your first point `raw_uc` has at least 2 extra operations; decoding of unit strings to numbers. Therefore if we only have to do that once for some speed relevant code thats a good thing especially as we can't remove YR_TO_S completely anyway. 

Its only in that usecase no need to apply the logic everywhere

je-cook (approved):
Great, thanks!

PR #2967, opened 2024-01-26
Classes to compute active/reactive power loads of the Coil Power Supply System

### LINKED ISSUE #3017: Magnetics

tplobo:
## Linked Issues
Closes #3017.

## Checklist


bluemira/power_cycle/coilsupply.py
```diff
@@ -0,0 +1,705 @@
...
+                args[one_field.name] = argument
+            return cls(**args)
+        raise ValueError(
+            "A 'CoilSupplyParameter' instance must be initialized "
+            f"with 'None' for an empty instance, a '{cls.__name__}' "
+            "instance for no alteration, a 'dict' for a distributed "
+            "instantiation or any of the following types for a "
+            f"single-value instantiation: {cls.single_value_types}. "
+            f"Argument was '{type(argument)}' instead."
+        )
```
je-cook, line 115: None of this is needed please remove, its over complicating things again
```suggestion
```

bluemira/power_cycle/coilsupply.py
```diff
@@ -0,0 +1,705 @@
...
+        The value stored in each attribute of the 'other' instance is
+        added to its respective 'dict' in the returned 'self'. If it
+        was not a dictionary, its value is stored in the 'other_key'
+        key of that attribute's dictionary.
+        """
+        all_fields = fields(self)
+        for one_field in all_fields:
+            self_value = getattr(self, one_field.name)
+            if self.validate_parameter(other):
+                other_value = getattr(other, one_field.name)
```
je-cook, line 149: stop validating
```suggestion
```

je-cook (changes requested):

bluemira/power_cycle/coilsupply.py
tplobo: Hi @je-cook,

this one is a bit more complicated than that... I have made a "dataclass factory" to create dataclasses to store vectors for a `CoilSupplySystem` object. This ensures the dataclass has one vector for each coil defined in the `CoilSupplySystem`. These methods are not so much for validation, but rather automatized creation of these containers. I call the related methods "validation" but they are used to instantiate the container objects independent of the input so maybe the term is not the most appropriate.

In short, the workflow is as follows:
1) a `CoilSupplyInputs` object is initialized, with a certain list of coils in `config.coil_names`
2) this automatically specifies what fields the container dataclass must have, so a `CoilSupplyParameter` class is defined as child of `CoilSupplyParameterABC`, and stored in the `parameter` attribute of `CoilSupplyInputs`
3) a `CoilSupplySystem` object is initialized with the `CoilSupplyInputs` object, so all methods that need to create a container can do it easily, either empty (full of `None` values) or passing an object to `validate_parameter`... dictionaries are distributed in the dataclass, other objects are just copied.

Perhaps that's not the best way to do this, but this was meant to be discussed with you and @oliverfunk 

Let me know what you think,

tplobo:
@je-cook, I assume this one should also be rebased on `feature/powercycle`, instead of develop... what is the argument for `--onto` in this case? I don't understand what should be the equivalent one for `powercycle_coils`

je-cook:
@tplobo can this be rebase on top of feature/powercycle please? It a bit impossible to review at the moment

tplobo:
> @tplobo can this be rebase on top of feature/powercycle please? It a bit impossible to review at the moment

@je-cook, yes, but as I mentioned above,I am unsure how to rebase this because I believe I need the `--onto` argument but I'd like some support in identifying the proper hash

je-cook:
> > @je-cook, I assume this one should also be rebased on feature/powercycle, instead of develop... what is the argument for --onto in this case? I don't understand what should be the equivalent one for powercycle_coils
> 
> > @tplobo can this be rebase on top of feature/powercycle please? It a bit impossible to review at the moment
> 
> @je-cook, yes, but I am unsure how to rebase this because I believe I need the `--onto` argument but I'd like some support in identifying the proper hash

once you've pulled the most recent commit on `feature/powercycle` this is command. You branched at e929c15c
`git rebase --onto feature/powercycle e929c15c`

tplobo:
> once you've pulled the most recent commit on `feature/powercycle` this is command. You branched at [1db075f](https://github.com/Fusion-Power-Plant-Framework/bluemira/commit/1db075f6b1ecd858895c9ed7ff62a6cea707c6ca) `git rebase --onto feature/powercycle 1db075f6`
> 
> EDIT looks like there is a modified commit compared to the feature branch so 2 things before the above:
> 
> `git rebase --interactive HEAD~59`
> 
> edit line 2 save and exit
> 
> before `pick e929c15c Powercycle layout base (#2873)` after `pick 1db075f6 Powercycle layout base (#2873)`

@je-cook, I have followed your instructions, but as I mentioned I had to make some corrections while interactively merging due to some conflicts; I've run my tests but you might encounter a bug or another. I have also added the IDM files used in verification as ignored files in a local `.gitignore`; let me know if we should change approach that to something else. Also, is it necessary to run a `filter-repo` to remove the previous files, or can we leave it as is?
Finally, to complete the tests I have also had to modify `match_domains` because one of the last `numpy` updates broke it. Thanks for the help and let me know if you need any support during review.

PR #2980, opened 2024-02-01
Fixed broken conditional statements in ComponentPlotter so now chooses correct plotting option

### LINKED ISSUE #2293: Plotting example shows some functionality that doesnt work as described
## Describe the bug

https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/46e1d550be02ab74f52928de587b83acae6eb0f5/bluemira/display/plotter.py#L612

In the above line `comp.plot_options` can never be `None` because by default all Components get an initialised plot options object when created. This means that the cascade approach defined in the plotting example can never work. See this cell where the red face is never created: 

https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/46e1d550be02ab74f52928de587b83acae6eb0f5/examples/geometry/plotting_tutorial.ex.py#L420

Some fixes to the overall example I've added in #2291 but this probably requires a bit of discussion about whether we want cascading config for any node in a tree. Secondly if we do want it how it should be implemented because in most situations only some overrides would be wanted. This feels quite a complicated feature.

- OS: 
- Version: 
- Branch:

kj5248:
## Linked Issues

Closes #2293 

## Description

Before the conditional statements relied on plotting options being None, however since default options exist this could never be true. This meant that when providing custom plotting options to a component post creation it was ignored and the default was used instead.

Wanted to change this so the following priority ordering existed:

1. component with non-default plotting options at creation
2. custom plotting options post-creation
3. default component plotting option

This was done by checking the various plotting options against their default values. Was not possible to check entire options dictionary against default but could however check the face colour, which is what was done until the full dictionary can be compared in a concise manner.

## Checklist

### DIFF
--- a/bluemira/display/plotter.py
+++ b/bluemira/display/plotter.py
@@ -41,6 +41,7 @@ from bluemira.geometry.coordinates import (
     get_centroid_3d,
     rotation_matrix_v1v2,
 )
+from bluemira.utilities.tools import flatten_iterable
 
 if TYPE_CHECKING:
     from bluemira.geometry.base import BluemiraGeo
@@ -552,9 +553,15 @@ class ComponentPlotter(BasePlotter):
 
         def _populate_plotters(comp):
             if comp.is_leaf and getattr(comp, "shape", None) is not None:
-                options = (
-                    self.options if comp.plot_options is None else comp.plot_options
-                )
+                if comp.plot_options.face_options["color"] in flatten_iterable(
+                    BLUE_PALETTE.as_hex()
+                ):
+                    if self.options.face_options["color"] == "blue":
+                        options = comp.plot_options
+                    else:
+                        options = self.options
+                else:
+                    options = comp.plot_options
                 plotter = _get_plotter_class(comp.shape)(options)
                 plotter._populate_data(comp.shape)
                 self._cplotters.append(plotter)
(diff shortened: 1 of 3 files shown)


kj5248:
In addition to the problem of this fix only checking face colours not the entire plot_options dictionary certain parts of the conditional statements don't work for the ComponentPlotter class but only when plot_2d is used. Without calling plot_2d self.options remains default even if a value has been given to it. This can be seen in the test.plotter.py file where to test the functionality of line 562 in plotter.py display.plot_2d had to be used in place of the ComponentPlotter. Both of these may need to be fixed at a later date.

kj5248:
For the proper checking of the settings against the default wanted something along the lines of: 

```
copy_options = comp.plot_options
copy_options.face_options["color"] = "blue"
if (copy_options == get_default_options()) and (
    comp.plot_options.face_options["color"] in flatten_iterable(
    BLUE_PALETTE.as_hex())):
    options = (
        comp.plot_options if self.options == get_default_options() else self.options
    )
else:
    options = comp.plot_options
```

This doesn't work but gives an idea of what is needed.

je-cook (approved):
Thanks! I think what needs to happen is an `__eq__` method on plot_options. However for now it does solve the issue. I will open another one for the points you suggested.

PR #2981, opened 2024-02-02
Add lower port tests

### LINKED ISSUE #2253: Improve tests for lower port

More and better tests are needed for the lower port duct.
The designer and builder were thoroughly tested during development, currently, however, only the angle of the duct is tested is unit tested.

It's a complex challenge writing tests that capture validly between in the input params and the output, hence this issue.

Some tests that could be implemented:

- Does the xz KOZ match an xz slice of the output xyz cad?
- Is the width and height of the port correct?
- Is the tf_offset correctly applied? Does the angled duct end at the correct place?
- Is the tf_thickness correctly taken into account when designing the angled extrude face?

je-cook:
## Linked Issues

Closes #2253 

## Description

Adds some lower port tests and does some small rearrangement.

There is a choice here which limit should be imposed on the straight section size.

1) scale to the maximum size but the minimum size is the user input (alternative 1)
2) limit to the user input (alternative 2)  **taken forward**

I've currently got 2) set up but up to the reviewer. I can't decide which is best.

The approximation of the TF coil being semi circular also leads to some failures in y point selection. For now I have just added a TODO (and will open an issue if agreed). This only affects extremes of input.

## Checklist

### DIFF
--- a/bluemira/utilities/opt_variables.py
+++ b/bluemira/utilities/opt_variables.py
@@ -17,6 +17,7 @@ from typing import Dict, Generator, Optional, TextIO, TypedDict, Union
 
 import numpy as np
 from tabulate import tabulate
+from typing_extensions import NotRequired
 
 from bluemira.base.look_and_feel import bluemira_warn
 from bluemira.utilities.error import OptVariablesError
@@ -27,9 +28,9 @@ class OptVarVarDictValueT(TypedDict, total=False):
     """Typed dictionary for a the values of an OptVariable from a var_dict."""
 
     value: float
-    lower_bound: float
-    upper_bound: float
-    fixed: bool
+    lower_bound: NotRequired[float]
+    upper_bound: NotRequired[float]
+    fixed: NotRequired[bool]
 
 
 VarDictT = Dict[str, OptVarVarDictValueT]
--- a/eudemo/eudemo/maintenance/lower_port/duct_designer.py
+++ b/eudemo/eudemo/maintenance/lower_port/duct_designer.py
@@ -115,7 +115,9 @@ class LowerPortKOZDesigner(Designer):
         )
 
         duct_straight_inner_extrude_boundary = self._straight_duct_inner_yz_boundary(
-            straight_top_inner_pt, straight_bot_inner_pt
+            straight_top_inner_pt,
+            straight_bot_inner_pt,
+            duct_angled_inner_extrude_boundary.bounding_box.y_max,
         )
 
         return (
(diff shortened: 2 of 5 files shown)


bluemira/utilities/opt_variables.py
```diff
@@ -27,9 +28,9 @@ class OptVarVarDictValueT(TypedDict, total=False):
     """Typed dictionary for a the values of an OptVariable from a var_dict."""
 
     value: float
-    lower_bound: float
-    upper_bound: float
-    fixed: bool
+    lower_bound: NotRequired[float]
+    upper_bound: NotRequired[float]
+    fixed: NotRequired[bool]
```
oliverfunk, line 34: These are fine beucase this type is used in the `adjust_variables` function, i.e. lower, upper etc can be none for that, no adjustment

eudemo/eudemo/maintenance/lower_port/duct_designer.py
```diff
@@ -176,6 +178,7 @@ def _straight_duct_inner_yz_boundary(
         self,
         straight_top_inner_pt: Tuple,
         straight_bot_inner_pt: Tuple,
+        # angled_duct_ymax: float,
```
oliverfunk, line 181: The comment etc. can be removed here

eudemo/eudemo/maintenance/lower_port/duct_designer.py
```diff
@@ -215,8 +223,12 @@ def _calc_y_point(x_point):
 
             return x_len * np.tan(self._half_beta)
 
-        ib_inner_y = _calc_y_point(ib_div_pt_padded[0]) - self.wall_tk
-        ob_inner_y = _calc_y_point(ob_div_pt_padded[0]) - self.wall_tk
+        # TODO alternative limit
+        flat_y_max = self.port_width / 2
+        ib_inner_y = min(flat_y_max, _calc_y_point(ib_div_pt_padded[0]) - self.wall_tk)
+        ob_inner_y = min(flat_y_max, _calc_y_point(ob_div_pt_padded[0]) - self.wall_tk)
+        # ib_inner_y = _calc_y_point(ib_div_pt_padded[0]) - self.wall_tk
```
oliverfunk, line 230: Yeah okay this makes sense. It won't fill up all the space, only up to what it needs to (i.e. port_width/2)

oliverfunk (commented):
Looks good, just remove comments

oliverfunk (approved):

PR #3000, opened 2024-02-07
Fix incorrect application of EPS (Papercut/ issue #1084)

### LINKED ISSUE #1084: Add a `GEOMETRY_TOLERANCE` or similar and use for treatment of geometry

FreeCAD's tolerance is of 1e-7 and we often use EPS in geometry in bluemira, which is much smaller than this and designed really for machine precision.

## Proposed solution

We should use a tolerance for geometry more consistent with that of FreeCAD. This has come in the context of checking for closed wires and coordinates, but is likely to be relevant elsewhere too.

OceanNuclear:
## Linked Issues

Closes #1084

## Description

Since floating point numbers isn't a dense representation of the real number line (it's sparse, but "denser" toward the center:
image
)
We have to consider that the "resolution" varies with the value of the float, i.e. the step size increases wrt. abs value of float.

## Interface Changes

Added a variable called "EPS_FREECAD" which would be used in place of EPS whenever geometry is involved.

## Checklist

### DIFF
--- a/bluemira/utilities/positioning.py
+++ b/bluemira/utilities/positioning.py
@@ -20,7 +20,6 @@ import abc
 import numpy as np
 from scipy.spatial import ConvexHull
 
-from bluemira.base.constants import EPS
 from bluemira.geometry.constants import VERY_BIG
 from bluemira.geometry.plane import BluemiraPlane
 from bluemira.geometry.tools import slice_shape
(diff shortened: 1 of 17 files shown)


bluemira/utilities/positioning.py
```diff
@@ -173,14 +172,17 @@ def _check_geometry_feasibility(self, geometry: BluemiraWire):
         if not self.geometry.is_closed:
             raise PositionerError("RegionInterpolator can only handle closed wires.")
 
-        xz_coordinates = self._get_xz_coordinates()
+        xz_coordinates = self._get_xz_coordinates(10000)
         hull = ConvexHull(xz_coordinates.T)
         # Yes, the "area" of a 2-D scipy ConvexHull is its perimeter...
-        if not np.allclose(hull.area, geometry.length, atol=EPS):
+        if not np.allclose(hull.area, geometry.length, rtol=1e-4, atol=0):
```
je-cook, line 178: Is this level of discretisation needed or could we do:
```python
if not np.allclose(hull.area, geometry.length, atol=D_TOLERANCE)
```
obviously needs a bit of further discussion is we talked about earlier

OceanNuclear: Resolved in the next commit, used a more lenient  as discussed.

je-cook (approved):

PR #3018, opened 2024-02-13
Coilset optimisation state for enabling SymmetricCircuit position optimisation 

oliverfunk:
## Linked Issues

Closes #{ID}

## Checklist


oliverfunk:
Thanks @CoronelBuendia

It's still very much WIP at the moment and contains some changed to spherical harmonics that should be moved to another PR.

I'll ping once it's ready, cheers!

bluemira/equilibria/plotting.py
oliverfunk, line 1: @je-cook quite awhile we had a chat about this. From what I remember, I think you said you would take a look.

I'm posting this here as a reminder.

Ping me if you want to chat.

je-cook: The subcoil (bad name) plotting is how we plot the discretised sections of a coil. That means for each coil you need to plot all of the quadratures used in calculations related to coils

Simple example before discretisation increased:
```python
from bluemira.equilibria.coils import CoilSet, Coil
coilset = CoilSet(Coil(1, 1, 0.5, 0.5), Coil(2, 2, 0.5, 0.5))
coilset.plot()
plt.show()
```
image

and increasing the discretisation on just one coil (by default it is not discretised at all, ie `discretisation = np.nan`, but the only effect for plotting is the number of quadratures)
```python
coilset['NONE_1'].discretisation = 0.2
coilset.plot()
plt.show()
```
image

oliverfunk:
All tests are passing locally for me.

This now needs to be rebased onto the py310 dev branch.

bluemira/equilibria/coils/_grouping.py
```diff
@@ -532,6 +537,7 @@ def get_max_current(self, max_current: float = np.inf) -> np.ndarray:
         return np.where(
             np.isnan(self.j_max) | ~self._flag_sizefix,  # or not
             max_current,
+            # todo: these are np arrays, not floats
             get_max_current(self.dx, self.dz, self.j_max),
```
je-cook, line 541: Whats todo?

bluemira/equilibria/coils/_grouping.py
```diff
@@ -800,10 +858,15 @@ def __init__(
 
         self.modify_symmetry(symmetry_line)
         diff = self._symmetrise()
-        self._coils[1].x = self._coils[1].x - diff[0]
-        self._coils[1].z = self._coils[1].z - diff[1]
+        self.symmetric_group.x -= diff[0]
+        self.symmetric_group.z -= diff[1]
```
je-cook, line 862: I'm sure you've checked this (as it works) but modification in place now works for positions in all cases (eg a symmetric circuit with more than one coil in each group)?

Or should we disable that usecase?

bluemira/equilibria/optimisation/problem/_breakdown.py
```diff
@@ -201,49 +201,55 @@ def __init__(
...
         )
-
         self._constraints = constraints
         if self._constraints is not None:
             self._constraints.append(stray_field_cons)
         else:
             self._constraints = [stray_field_cons]
 
-        max_currents = np.atleast_1d(max_currents)
-        self.bounds = (-max_currents / self.scale, max_currents / self.scale)
```
je-cook, line 223: Could you comment on this removal?

bluemira/equilibria/optimisation/problem/_nested_position.py
```diff
@@ -87,15 +84,13 @@ def __init__(
         self,
         sub_opt: CoilsetOptimisationProblem,
         eq: Equilibrium,
-        targets: MagneticConstraintSet,
```
je-cook, line 90: Why is this removed?

bluemira/equilibria/optimisation/problem/_nested_position.py
```diff
@@ -264,9 +262,17 @@ def sub_opt_objective(
...
-                sub_opt_prob.coilset[coil].position = position
+            sub_opt_prob.coilset.set_optimisation_state(coil_position_map=pos_map)
+
+            # TODO: is this necessary?
+            # I though this had be done after coils positions are changed
+            # sub_opt_prob.eq._remap_greens()
+            # sub_opt_prob.eq._clear_OX_points()
+
+            # todo: Why would you use the self.initial_currents?
+            # And not just the coilset's current currents?
```
je-cook, line 273: These need to be removed/resolved

je-cook (changes requested):
This is just an initial once through, I havent run it yet. 

My only overriding issue is the addition of optimisation methods onto the coil object as before it was just an object that could be used in optimisation whereas now its very much more interlinked.
This possibly is a bit of double duty but I see the benefits. I'm unsure of my full feelings on this and this comment is to stimulate a discussion.

On the whole great work getting this going all my other comments are either minor or me trying to understand whats going on.

bluemira/equilibria/coils/_grouping.py
oliverfunk: Was a note to correct the types for the func. params.
I'll update

bluemira/equilibria/coils/_grouping.py
oliverfunk: I dono, it's the same thing that was happening before.

I think we talked about it and I think if it's a coilgroup, then if say diff[1] is a single float, coils are moved st. diff[1] is at the center otherwise there need to be enough values in diff[1] for it to set the z's for every coil

bluemira/equilibria/optimisation/constraints.py
```diff
@@ -725,7 +744,7 @@ class MagneticConstraintSet(ABC):
         - Populate constraints with super().__init__(List[MagneticConstraint])
     """
 
-    __slots__ = ("A", "background", "coilset", "constraints", "eq", "target", "w")
+    __slots__ = ["A", "background", "coilset", "constraints", "eq", "target", "w"]
```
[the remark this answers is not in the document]
oliverfunk: That's werid, I think ruff changed that but I'm not sure.

bluemira/equilibria/optimisation/harmonics/harmonics_approx_functions.py
```diff
@@ -647,14 +657,16 @@ def spherical_harmonic_approximation(
...
-        if degree == max_degree:
-            raise BluemiraError(
+        if degree == degree_limit_opt_coils:
+            bluemira_print("Uh oh, you cannot use more degrees than optimisable coils.")
+            break
+        if degree == degree_limit_collocation:
+            bluemira_print(
                 "Uh oh, you may need to use more degrees for a fit metric of"
                 f" {acceptable_fit_metric}! Use a greater number of collocation points"
                 " please."
```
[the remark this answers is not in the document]
oliverfunk: Georgi likes her Uh oh!'s

bluemira/equilibria/optimisation/objectives.py
```diff
@@ -168,6 +174,8 @@ def tikhonov(a_mat: np.ndarray, b_vec: np.ndarray, gamma: float) -> np.ndarray:
     x:
         The result vector
     """
+    if currents_rep_mat is not None:
+        a_mat = a_mat @ currents_rep_mat
```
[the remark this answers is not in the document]
oliverfunk: I think I tried that and numpy threw an error. I think it said in-place mat-mul isn't supported yet or something.

bluemira/equilibria/optimisation/objectives.py
```diff
@@ -108,7 +109,7 @@ class CoilCurrentsObjective(ObjectiveFunction):
     @staticmethod
     def f_objective(vector: npt.NDArray[np.float64]) -> float:
         """Objective function for an optimisation."""
-        return np.sum(vector**2)
+        return np.sum(np.square(vector))
```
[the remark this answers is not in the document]
oliverfunk: I think it was the return type was better, also just think it's better to use.

bluemira/equilibria/optimisation/problem/_breakdown.py
oliverfunk: Not removed, just moved. Do for consistency between the COP's

bluemira/equilibria/optimisation/problem/_nested_position.py
oliverfunk: They aren't used.

This COP is doing pos. opt. only. It does not use a magnetic constraint set.
(the sub_opt may have one/use one but that would be handled entirely within the sub_opt)

bluemira/equilibria/run.py
```diff
@@ -422,7 +422,7 @@ def run_premagnetisation(self):
                 opt_conditions=self.bd_settings.opt_conditions,
                 constraints=constraints,
             )
-            result = problem.optimise(x0=max_currents, fixed_coils=False)
+            result = problem.optimise(fixed_coils=False)
```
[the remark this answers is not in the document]
oliverfunk: Yeah, it's because, before this update, the BreakdownCOP ignored x0 even if one was passed in.
So this change maintains the behaviour.

If I leave it, EUDEMO doesn't converge as well as it did before.

tests/equilibria/test_optimiser.py
```diff
@@ -82,33 +82,42 @@ def setup_class(cls):
                 make_polygon({"x": [xl, xu, xu, xl, xl], "z": [zl, zl, zu, zu, zl]})
             )
 
+        # TODO, change the stuff here
```
oliverfunk, line 85: remove

bluemira/equilibria/optimisation/objectives.py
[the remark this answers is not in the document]
je-cook: I think it is with all the new stuff coming in with the numpy updates with python3.10. Good to check

bluemira/equilibria/coils/_coil.py
```diff
@@ -154,8 +154,8 @@ def __init__(
 
         self.x = x
         self.z = z
-        self.dx = dx
-        self.dz = dz
+        self._dx = dx
+        self._dz = dz
```
je-cook, line 158: revert I think? the setters are needed otherwise this is just a repeat of L147/148

bluemira/equilibria/coils/_tools.py
```diff
@@ -106,7 +106,9 @@ def check_coilset_symmetric(coilset: CoilSet) -> bool:
     return True
 
 
-def get_max_current(dx: float, dz: float, j_max: float) -> float:
+def get_max_current(
+    dx: float | np.ndarray, dz: float | np.ndarray, j_max: float | np.ndarray
+) -> np.float64 | np.ndarray:
```
je-cook, line 111:
```suggestion
) -> float | np.ndarray:
```

bluemira/equilibria/coils/_tools.py
```diff
@@ -123,4 +125,4 @@ def get_max_current(dx: float, dz: float, j_max: float) -> float:
     -------
     Maximum current [A]
     """
-    return abs(j_max * (4 * dx * dz))
+    return np.abs(j_max * (4 * dx * dz))
```
je-cook, line 128: this is actually not needed:
image

bluemira/equilibria/optimisation/objectives.py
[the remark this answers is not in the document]
je-cook: have you checked this since the rebase? Just curious

je-cook (commented):
Some more minors

bluemira/equilibria/coils/_coil.py
oliverfunk: Ah right, I was getting confused.

bluemira/equilibria/optimisation/objectives.py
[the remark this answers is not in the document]
oliverfunk: No it won't work because after the op. the shape of a_mat changes. (i.e. in-place mutation is invalid in this case)

oliverfunk:
@CoronelBuendia I need to chat to you about the Jacbobian/derivative functions for constraints and how they work when using sym. ccts.

Take for ex. the CoilForceConstraint, if it's used with a CoilSet that has a mixture of SymmetricCircuit and normal coils, the force on one of the coils (in the SymmetricCircuit) will be different.

At the moment, I matmul the result with the 'representation matrix', which effectively sum's the derivatives, which I know is not correct.

bluemira/equilibria/optimisation/problem/base.py
```diff
@@ -273,25 +199,61 @@ def update_magnetic_constraints(
...
+            f_c = f.f_constraint
+            df_c = getattr(f, "df_constraint", None)
+
+            if coilset._opt_currents_symmetry_status in {
+                CoilSetSymmetryStatus.FULL,
+                CoilSetSymmetryStatus.PARTIAL,
+            }:
+                # wrap the constraint function
+                def wrapped_f_c(x, f=f):
+                    return f.f_constraint(coilset._opt_currents_expand_mat @ x)
```
je-cook, line 221: to preserve any docstrings in f.f_constraint, you'll need to import functools probs
```suggestion
                @functools.wraps(f.f_constraint)
                def wrapped_f_c(x, f=f):
                    return f.f_constraint(coilset._opt_currents_expand_mat @ x)
```

bluemira/equilibria/optimisation/problem/base.py
```diff
@@ -273,25 +199,61 @@ def update_magnetic_constraints(
...
+                    df_c = None
+                else:
+                    # if fully symmetric, the result of the derivative
+                    # function is multiplied by the repetition matrix
+                    # (to reduce the shape to the number of current optimisable coils)
+                    # by multiplying the result with the repetition matrix
+                    # and dividing by 2 (as the values are added together)
+                    def wrapped_df_c(x, f=f):
+                        df_res = f.df_constraint(coilset._opt_currents_expand_mat @ x)
+                        return df_res @ coilset._opt_currents_sym_reduce_mat
```
je-cook, line 242:
```suggestion
                    @functools.wraps(f.df_constraint)
                    def wrapped_df_c(x, f=f):
                        df_res = f.df_constraint(coilset._opt_currents_expand_mat @ x)
                        return df_res @ coilset._opt_currents_sym_reduce_mat
```

je-cook (approved):
Two minors otherwise from my point of view I'm happy

bluemira/equilibria/coils/_grouping.py
```diff
@@ -1015,3 +1157,287 @@ def from_group_vecs(
...
+        n_opt_coils = cc.n_current_optimisable_coils
+        n_distinct_coils_and_groupings = len(cc._coils)
+
+        if cc._opt_currents_symmetry_status == CoilSetSymmetryStatus.NONE:
+            return np.eye(n_all_coils)
+
+        # this should be true as, at the top level, the number
+        # of coil or group objects should be the same as the no
+        # of optimisable coils
+        assert n_opt_coils == n_distinct_coils_and_groupings  # noqa: S101
```
CoronelBuendia, line 1305: Hmm, last time I checked (and it was a while ago) there should be no asserts in production code. Is this necessary now that development on this branch is closing up?

bluemira/equilibria/coils/_grouping.py
```diff
@@ -1015,3 +1157,287 @@ def from_group_vecs(
...
+                "Symmetry reduce matrix can only be used with a CoilSet "
+                "that only has SymmetricCircuits"
+            )
+
+        mat = np.zeros((n_all_coils, n_opt_coils))
+        i_row = 0
+        for i_col in range(n_opt_coils):
+            mat[i_row, i_col] = 1
+            mat[i_row + 1, i_col] = 0
+            i_row += 2
```
CoronelBuendia, line 1347: I haven't looked at this file in detail, but I notice just that in the corresponding function above, you use `n_coils_in_group`, whereas here you use 2. Is there some mechanism that limits the symmetrycircuits to 2 coils? Should these two related functions be "similar" in implementation, or is there a reason for this difference?

CoronelBuendia (changes requested):
OK I have left some comments on the implementation, and I have a couple of general ones:
* Could we split out the spherical harmonics development work from this branch?
* Needs an integration test with actual solving of very simple optimisation problem using these mechanics.
* Needs to be worked into a relevant example
* Needs some minimal documentation, including description of how many constraints exist per sym circuit, and what happens in the partial case.
Other than looks good in theory!

scripts/update_dependencies.sh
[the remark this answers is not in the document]
oliverfunk: The file's permission was updated to make it executable, it hasn't been deleted.

bluemira/equilibria/optimisation/harmonics/harmonics_constraints.py
```diff
@@ -54,13 +54,16 @@ def __init__(
         ref_harmonics: npt.NDArray[np.float64],
         r_t: float,
         sh_coil_names: list,
-        tolerance: float | npt.NDArray = None,
```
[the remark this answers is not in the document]
oliverfunk: That's a good idea, I'l look at doing that.

bluemira/equilibria/coils/_grouping.py
oliverfunk: Yeah that's well observed.

I can make it clearer with a comment but it's becuase I raise if `cc._opt_currents_symmetry_status != CoilSetSymmetryStatus.FULL:` which means every element in the coilset is a SymmCct. Which means I just need a matrix of alternationg 1's and 0's per col.

bluemira/equilibria/coils/_grouping.py
oliverfunk: That is true.

There some aasserts in BM. They are usually optimsed away and only enforced during development but I'm not entirly sure how that works in python (as there is no build step).

I'll probably just convert it into a raise but not really an error, it's more like logically these two values must be equal.

bluemira/equilibria/optimisation/constraint_funcs.py
```diff
@@ -303,8 +306,9 @@ def __init__(
 
     def f_constraint(self, vector: npt.NDArray[np.float64]) -> npt.NDArray[np.float64]:
         """Constraint function"""
-        n_coils = len(vector)
         currents = self.scale * vector
+
+        n_coils = self.n_CS + self.n_PF
```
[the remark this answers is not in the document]
oliverfunk: n_CS and n_PF were already passed into the init of this class and are used in the math that follows (i.e. the assumption was already made).

I could probably change it back to how it was before due the way I do the vector expansion now before the vector is passed into this function.

bluemira/equilibria/coils/_grouping.py
```diff
@@ -1015,3 +1136,118 @@ def from_group_vecs(
...
+
+    @property
+    def position_optimisable_coil_names(self) -> list[str]:
+        """
+        Get the names of the coils that can be position optimised.
+        """
+        optimisable_coil_names = [
+            c.primary_coil.name if isinstance(c, SymmetricCircuit) else c.name
+            for c in self._coils
+        ]
```
je-cook, line 1206: I think if we change this to a generator we will only iterate through it once instead of twice
```suggestion
        optimisable_coil_names = (
            c.primary_coil.name if isinstance(c, SymmetricCircuit) else c.name
            for c in self._coils
        )
```

bluemira/equilibria/optimisation/constraint_funcs.py
je-cook, line 311: I think we want to put an error in the init so that we catch the case where n_CS and n_PF are 0. 
As we've mentioned in passing this would error slighly further down thew function without this change anyway if they were 0

bluemira/equilibria/optimisation/problem/_position.py
```diff
@@ -94,33 +95,46 @@ def __init__(
...
+            )
+
+            len_mapped_pos = len(initial_mapped_positions)
+            x0 = np.concatenate((initial_mapped_positions, cs_opt_state.currents))
+
         eq_constraints, ineq_constraints = self._make_numerical_constraints()
         opt_result = optimise(
-            f_objective=self.objective,
-            x0=np.concatenate((initial_mapped_positions, initial_currents)),
+            f_objective=lambda vector: self.objective(vector, len_mapped_pos),
```
je-cook, line 122: This will fail if `x0` is provided because `len_mapped_pos` will not be defined

bluemira/equilibria/optimisation/problem/_tikhonov.py
```diff
@@ -183,10 +190,14 @@ def optimise(self, **_) -> CoilsetOptimiserResult:
...
-        f_x = np.linalg.norm(a_mat @ current - b_vec) + np.linalg.norm(
-            self.gamma * current
+        opt_currents = self.coilset.get_control_coils().current + current_adjustment
+        self.coilset.set_optimisation_state(opt_currents=opt_currents, current_scale=1.0)
+
+        current = self.coilset.current
+
+        f_x = floatify(
+            np.linalg.norm(a_mat @ current - b_vec)
+            + np.linalg.norm(self.gamma * current)
```
je-cook, line 200: you could just use `.item()` here as you know it will be a numpy array. Up to you.

bluemira/equilibria/optimisation/problem/_position.py
oliverfunk: Good catch!

je-cook (approved):
As we discussed it would be nice to have a position optimisation example/test. We havent had a good one so far so I don't particularly see it as blocking but it is definitely a want to have.

PR #3023, opened 2024-02-15
WIP: SH Plotting

clmould:
## Linked Issues

Closes #{ID}

## Checklist


bluemira/utilities/plot_tools.py
```diff
@@ -123,6 +123,43 @@ def make_gif(
...
+        "f": f,
+        "ax": ax,
+        "pname": pname,
+        "folder": folder,
+        "save": save,
+    }
+
+
 def save_figure(
     fig, name, save=False, folder=None, dpi=600, file_format="png", **kwargs
```
je-cook, line 164: Just had a quick look in passing and noticed this.
If possible could we change the file_format default to 'svg' here? 
It means that the figures will be vector graphics by default and therefore not pixelate on resizing. It does make having a default dpi a bit irrelevant but if it doesnt break anything leave that as is.

bluemira/equilibria/plotting.py
```diff
@@ -100,18 +101,35 @@ class Plotter:
     Utility plotter abstract object
     """
 
-    def __init__(self, ax=None, **kwargs):
+    def __init__(self, ax=None, subplots=False, **kwargs):
```
je-cook, line 104: This will fix your linter issues 
```suggestion
    def __init__(self, ax=None, *, subplots=False, **kwargs):
```

geograham:
Hey, could you add a few lines to the description? I don't think the original issue fully covers the extra plotting features you have added.

geograham:
Currently we have the option to plot the 'full' psi or 'split' psi contributions. The alternative is to plot all three (total, plasma, coilset) along side each other. Pro = would remove a plotting option input. Con = might be a lot to look at at once during an optimisation. What do you (@clmould, @oliverfunk, @je-cook) think? 

(This is something we had talked about before but not decided on).

je-cook:
> What do you (@clmould, @oliverfunk, @je-cook) think?
I have no preference it more what people think is useful. 

My only blocker on here atm is there is no test to cover 80% of this which is needed. Btw to view figures while testing you can use the `--plotting-on` pytest flag.

PR #3043, opened 2024-02-26
OBSOLETE: TF coils winding pack

### LINKED ISSUE #138: Winding pack design optimisations
In GitLab by @CoronelBuendia on Jun 24, 2020, 15:51

CEA have a tool that does this, but it only supports their winding packs. It is a "proprietary" Excel spreadsheet... at least last time I saw it in action; it isn't automated, it needs to be used and frequently tweaked by an expert. All things considered it's fairly crude, if useful.

A group in SPC (specifically, Paul Scherrer Institut, Villigen) and another in ENEA have similar capabilities for their own winding pack designs. Although they don't call them tools as such, they have this "capability" which I imagine isn't entirely manual. The difficulty is, all of these are proprietary or "in-confidence" and the relevant knowledge doesn't really exist inside the organisation. These approaches also are specific to winding pack designs. Many of the conductors used in these designs have been tested in representative EM conditions, so there are quite a few experimental tweaking factors in these tools

The way winding packs are designed is to first have a guess on the size, sometimes using existing 1-D models (e.g. in PROCESS) or just saying current density = 12.5 MA/m^2, then have a stab in 2-D, and then check everything works in higher fidelity models (3-D FEA and other studies). We'd like to improve on the 2-D design step: replicating some existing capability in other labs, but improving and automating the process. We'd also like to prepare inputs for the higher fidelity models. It would need to run fairly fast (~seconds) and could potentially be used to make reduced-order models informing our simple guesstimates on things like "average winding pack current density" for a design.

* Inputs (from other parts of the BLUEPRINT code):
  * The geometrical cross-section of the winding pack, and the shape of the winding (circle for a PF coil, other shapes for a TF coil)
  * The currents required to be generated by the winding pack
  * The loads the winding pack is subject to
    * EM loadings
    * Magnetic fields across the winding pack (important for current density of the superconductor)
    * Neutronic heating deposited in the winding pack
    * Quench scenario
  * Winding pack parameters and design choices
    * Superconductor type (e.g. NbTi, Nb3Sn, etc.)
    * Winding type (pancake vs. layer)
    * Insulation thicknesses (inter-turn, inter-layer/inter-pancake, and ground)
    * Reference operation T (e.g. 4.2 K)
    * Coolant (e.g. He)
    * The desired T margin to quench (e.g. 1.5 K)
    * Maximum single conductor current rating (e.g. 100 kA)
    * Other thicknesses and values, etc.
  * Material properties
    * Superconducting properties and parameterisations
    * Steel properties
    * Coolant properties
* Tool activities:
  * Optimise the design of the winding pack for minimal size, so that it meets its current density requirement within the available space and withstands the applied loads
    * Determine the number of conductors in the winding and the layout of the conductors
    * Make the structures in the winding pack can withstand the stresses
    * Make the copper shunt fraction sufficient to withstand a quench
    * If the space available is too much / insufficient -> shrink / grow the WP geometry accordingly
  * Outputs for use in the BLUEPRINT code
    * Winding pack layout and 2-D geometry
    * Volumetric fractions of copper, superconductor, steel, insulation, ground insulation, etc.
    * Margin to the various constraints
  * Prepare inputs for higher-fidelity analyses
    * CAD models
    * Smeared material properties

ivanmaione:
## Linked Issues
Closes #138

## Checklist


je-cook:
Just something I noticed in matplotlib while starting to have a play with this. When using `ax.fill` by default mpl will snap patches to the nearest pixel this created some weird artefacts in the figures of the winding pack in favour of being faster to render and plot.

To have a more accurate representation of the patch you can supply the `snap` kwarg to `ax.fill` like so: 
`ax.fill(...., snap=False)`

It produces this for the TF example which while still having some artefacting is slightly better if being slightly slower to plot:
image

CoronelBuendia:
Could we possibly get a rebase? I changed the target branch to the feature branch but that may have messed things up.. Will certainly make it easier for me to review!

PR #3093, opened 2024-03-13
WIP: New div opt

geograham:
## Description

Connection length FOM, divertor leg length FOM and grazing angle constraint for optimisation. 

## Checklist


bluemira/equilibria/optimisation/constraint_funcs.py
```diff
@@ -254,6 +261,67 @@ def f_constraint(self, vector: npt.NDArray[np.float64]) -> npt.NDArray[np.float6
         return max(lcfs.x) - self.radius
 
 
+class GrazingAngleConstraintFunction(ConstraintFunction):
```
CoronelBuendia, line 264: Hmm... this is something we used to have in BLUEPRINT, but it was quite different... 

Grazing angle is in 3-D, so you need target inclination as well, btw. But essentially the grazing angle at a point (and target inclination) is just a function of field, so it could be a `MagneticConstraint`. Doing it this way, you are exposing yourself to the non-linear "movement" of the plasma and its legs, so you will not be able to use this in a `MagneticConstraintSet` or with a gradient-based optimiser.

The way it used to be done in BLUEPRINT was with the equivalent of a `MagneticConstraint` which enforced effectively a poloidal field angle at a fixed location. This was usually done in combination with an `IsofluxConstraint` for the divertor strike point.

I worry this approach will not be very stable, at least in terms of an optimisation problem that produces equilibria from scratch. If you are operating on an existing equilibrium that you are seeking to change slightly, this may be OK. How is it behaving so far on your use-case?

geograham:
This is work is now going to be addressed in a new pull request that references this one.

PR #3096, opened 2024-03-13
Addition of discretize_reduced function to improve TF coil magnetostatics computation time

### LINKED ISSUE #3094: Inefficient TF coil discretisations for magnetostatics

When using a PictureFrame TF coil the xz shape gets built. Then for the magnetostatics this shape gets discretized into a series of points so that a circuit of current sources can be created for it. Currently the shape is split up evenly everywhere, so for a straight segment it is split up, increasing the computation time, unnecessarily.

## Proposed solution

To solve this can take the discretized points and iterate over them to remove the middle points in a straight segment. This will help reduce the computation time for magnetostatics calculations on TF coils.

kj5248:
## Linked Issues

Closes #3094 

## Description

Added function that takes discretized set of points and reduces it by removing middle points along a straight line. Below fig shows this for TF coil, where crosses are original points and dots are the reduced set of points. 

discretize_reduced

When run on a simple case using ArbitraryPlanarRectangularXSCircuit using the PrincetonD shape discretized with 100 points and calculating the field for a 20x20 grid the run times are: 0.677s for the original and 0.317s for the reduced.

## Checklist

### DIFF
--- a/bluemira/geometry/wire.py
+++ b/bluemira/geometry/wire.py
@@ -10,7 +10,7 @@ Wrapper for FreeCAD Part.Wire objects
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Iterable, List, Optional, Tuple, Union
+from typing import TYPE_CHECKING
 
 import bluemira.codes._freecadapi as cadapi
 from bluemira.base.look_and_feel import LOGGER, bluemira_warn


bluemira/geometry/wire.py
```diff
@@ -153,6 +152,44 @@ def discretize(
...
+        for i in range(1, len(points) - 1):
+            # go in reverse so element deletion doesn't change positional argument
+            j = len(points) - 1 - i
+            p0 = points[:, j + 1]
+            p1 = points[:, j]
+            p2 = points[:, j - 1]
+            l1 = (p1 - p0) / np.linalg.norm(p1 - p0)
+            l2 = (p2 - p1) / np.linalg.norm(p2 - p1)
+            if np.equal(np.dot(l1, l2), 1.0):
+                points_copy = np.delete(points_copy, j, 1)
```
je-cook, line 190: can we vectorise this. The steps would be:

1) get all l1 and l2 as a 2d array something like (I think my indexes are wrong on the points)
```
p0 = points[:, 2:-2]
p1 = points[:, 1: -3]
p2 = points[:,  : -4]
l1 = p1 -p0/ np.linalg.norm(p1 - p0, axis=1)
l2 = p2 -p1/ np.linalg.norm(p2 - p1, axis=1)
```
2) `mask = np.ones_like(points, dtype=bool)`
3) `mask[:, np.equal(np.dot(..., axis=..), 1.0)] = False`
2) `return Coordinates(points[mask])`

je-cook (changes requested):
I think we can combine the function into the existing discretise with an addition to the extension of `reduce:bool = False`.

CoronelBuendia:
This caught my eye. What current sources are you using?

The picture frame can be modelled with a grand total of 8 sources, with a rectangular cross-section conductor of constant current density, with relatively high fidelity and no sensitivity to discretisation. If the inboard leg is trapezoidal, 8 sources still holds.

You can take each sub-wire that is not a CircleArc, and model that with a single current source, and take each CircleArc and model that with a single current source, no discretisation other than that needed technically.

If you are using Biot-Savart, you still need to preserve many points in the straight legs.

bluemira/geometry/wire.py
kj5248: Made these changes, the dot product came out quite messy not sure if there is a better way to do it than I have...

```
diag_dot = np.around(
                np.append(np.append(0.0, np.diag(np.inner(l1, l2))), 0.0)
            ).astype(int)
```

numpy dot (or other versions of it) don't take dot products for each row and put into a new array of size N but just performs matrix multiplication. Inner was the closest I could get but it does all the combinations, hence the diagonalisation. The rounding and changing type to integer was because otherwise the numpy equals comparison gave wrong answers.

bluemira/geometry/wire.py
```diff
@@ -151,6 +154,18 @@ def discretize(
...
+            p0 = points[:-2, :]
+            p1 = points[1:-1, :]
+            p2 = points[2:, :]
+            l1 = (p1 - p0) / np.linalg.norm((p1 - p0)[0, :])
+            l2 = (p2 - p1) / np.linalg.norm((p2 - p1)[0, :])
+            mask = np.ones_like(points[:, 0], dtype=bool)
+            diag_dot = np.around(
+                np.append(np.append(0.0, np.diag(np.inner(l1, l2))), 0.0)
+            ).astype(int)
+            mask[np.equal(diag_dot, 1)] = False
```
je-cook, line 167:
```suggestion
            p0 = points[:-2]
            p1 = points[1:-1]
            p2 = points[2:]
            l1 = (p1 - p0) / np.linalg.norm((p1 - p0), axis=1)
            l2 = (p2 - p1) / np.linalg.norm((p2 - p1), axis=1)
            mask = np.ones_like(points[:, 0], dtype=bool)
            diag_dot = np.hstack([0, np.einsum('ij, ij -> i', l11, l22).astype(int), 0])
            mask[np.equal(diag_dot, 1)] = False
```

bluemira/geometry/wire.py
```diff
@@ -151,6 +154,18 @@ def discretize(
             points = cadapi.discretize_by_edges(self.shape, ndiscr=ndiscr, dl=dl)
         else:
             points = cadapi.discretize(self.shape, ndiscr=ndiscr, dl=dl)
+        if reduce is True:
```
je-cook, line 157:
```suggestion
        if reduce:
```

CoronelBuendia:
> > This caught my eye. What current sources are you using?
> > The picture frame can be modelled with a grand total of 8 sources, with a rectangular cross-section conductor of constant current density, with relatively high fidelity and no sensitivity to discretisation. If the inboard leg is trapezoidal, 8 sources still holds.
> > You can take each sub-wire that is not a CircleArc, and model that with a single current source, and take each CircleArc and model that with a single current source, no discretisation other than that needed technically.
> > If you are using Biot-Savart, you still need to preserve many points in the straight legs.
> 
> @CoronelBuendia The reason I added this is because in STEP bluemira for the TF coil builder. When creating a current source circuit using ArbitraryPlanarRectangularXSCircuit it uses a discretisation of the TF centreline with a ndiscr of 100. I thought this seemed overkill for the straight edges but assumed this level of discretisation was needed for the corners?

Yes, I understand. To me this is something that should handled specifically for your purposes, and not adding functionality to `BluemiraWire`.

If you always, always, always use a `PictureFrame` parameterisation, then it makes sense to have a magnetostatics representation of it that is as computationally efficient as possible. Rather than using `ArbitraryPlanarRectangularXSCircuit` you could write a simple function that takes the underlying CAD shape (composed of 4 straight lines and 4 circle arcs) and make a `SourceGroup(List[TrapezoidalPrismCurrentSource]+List[CircularArcCurrentSource])` you would have 8 current sources per TF coil (as opposed to 100), and absolutely no loss of accuracy due to discretisation.

PR #3114, opened 2024-03-20
Rectangular XS current source from a `BluemiraWire`

CoronelBuendia:
## Linked Issues

Closes #{ID}

## Description

I think something like this would do the job for highest-fidelity fastest-run-time for the picture frame coils, but the function has some severe limitations at present, and lacks in generality.

image

This is quite clearly not yet working for TripleArc, and I am not sure why... It's either something really stupid with coordinate systems, or something much more sinister.

image

## Checklist


CoronelBuendia:
@kj5248 have a go with this and see if it will do what you want. It needs testing (do not trust pictures! - in this case I have just used them to tell me if the field is symmetric about the Z-plane, and that we recover 1/r behaviour). It still isn't quite there, but I've just been doing this during a workshop and have exceeded the time I wish to spend on this.

PR #3115, opened 2024-03-20
Types updates for python3.10

je-cook:
## Description

Updated types due to deprecation in python3.10

## Checklist


bluemira/equilibria/num_control.py
```diff
@@ -47,9 +47,9 @@ def psi(self) -> npt.NDArray[np.float64]:
 
     @staticmethod
     def Bx(
-        x: Union[float, npt.NDArray[np.float64]],
-        z: Union[float, npt.NDArray[np.float64]],  # noqa: ARG004
-    ) -> Union[float, npt.NDArray[np.float64]]:
+        x: npt.ArrayLike,
```
oliverfunk, line 50: This makes me think, should we not use this type for all np array inputs? Unless we want to be super clear about it being np.float64 for example.

Or this could replace all float | np.ndarray type?

oliverfunk (approved):
I couldn't do them all but from the majority I did scan, they all look good to me!

Nice!

bluemira/equilibria/num_control.py
je-cook: `npt.ArrayLike` is too general for a return type but we could do this for input types in certain places. It would mean most functions would need an `np.asarray` call at the beginning. I think more generally we will need a typing file in base at some point where we can stick typing aliases for things like this.

PR #3117, opened 2024-03-20
Reduce warning count on python310

je-cook:
## Description

numpy 1.25 deprecated float conversion for rank >0 scalars 

```python
float(np.array([5]))  # warning
float(np.array(5)) # ok
```
## Checklist

### DIFF
--- a/bluemira/utilities/opt_tools.py
+++ b/bluemira/utilities/opt_tools.py
@@ -13,6 +13,7 @@ import warnings
 import numpy as np
 
 from bluemira.base.look_and_feel import bluemira_warn
+from bluemira.utilities.tools import floatify
 from bluemira.utilities.error import InternalOptError
 
 warnings.warn(
@@ -92,7 +93,7 @@ def regularised_lsq_fom(x, A, b, gamma):
         Residual vector (Ax - b)
     """
     residual = np.dot(A, x) - b
-    number_of_targets = float(len(residual))
+    number_of_targets = floatify(len(residual))
     fom = residual.T @ residual / number_of_targets + gamma * gamma * x.T @ x
 
     if fom <= 0:
--- a/bluemira/utilities/opt_tools.py
+++ b/bluemira/utilities/opt_tools.py
@@ -13,7 +13,6 @@ import warnings
 import numpy as np
 
 from bluemira.base.look_and_feel import bluemira_warn
-from bluemira.utilities.tools import floatify
 from bluemira.utilities.error import InternalOptError
 
 warnings.warn(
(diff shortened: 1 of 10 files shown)


je-cook:
ok I think this is as many as can be removed without rewriting a lot of array processing code, I will open an issue once this goes into develop

bluemira/utilities/opt_tools.py
```diff
@@ -92,7 +93,7 @@ def regularised_lsq_fom(x, A, b, gamma):
         Residual vector (Ax - b)
     """
     residual = np.dot(A, x) - b
-    number_of_targets = float(len(residual))
+    number_of_targets = floatify(len(residual))
```
oliverfunk, line 96: Does len not return an int? Is floatify needed here?

bluemira/utilities/tools.py
```diff
@@ -299,6 +300,15 @@ def cross(
 # =====================================================
 
 
+def floatify(x: npt.ArrayLike):
+    """
+    Avoid numpy warnings for float(x) for >0 rank scalars
+    """
+    if x is None:
+        raise TypeError("argument must be a string or a real number, not 'NoneType'")
+    return np.asarray(x, dtype=float).item()
```
oliverfunk, line 309:
```suggestion
    return np.asarray(x, dtype=float).item(0)
```

I think this is more explicit

bluemira/utilities/tools.py
```diff
@@ -299,6 +300,15 @@ def cross(
 # =====================================================
 
 
+def floatify(x: npt.ArrayLike):
```
oliverfunk, line 303:
```suggestion
def floatify(x: npt.ArrayLike) -> float:
```

bluemira/utilities/tools.py
```diff
@@ -299,6 +300,15 @@ def cross(
...
+    """
+    Avoid numpy warnings for float(x) for >0 rank scalars
```
oliverfunk, line 305: Maybe a slightly more explicate explanation

"Converts the np array or float into a float by returning the first element or the element itself"

bluemira/utilities/tools.py
```diff
@@ -299,6 +300,15 @@ def cross(
...
+    """
+    if x is None:
+        raise TypeError("argument must be a string or a real number, not 'NoneType'")
```
oliverfunk, line 308: The argument could be an array too no?

Maybe just "The argument cannot be None" or something

oliverfunk (changes requested):

bluemira/utilities/tools.py
je-cook: I'm not going to do this because it will then not error when converting a multi element array to a float (which we want if we're emulating the previous behaviour). I will add this to the docstring.

bluemira/utilities/opt_tools.py
je-cook: No it doesnt AFAIK

PR #3153, opened 2024-04-04
Add functionality for uneven endcaps polyhedral current source

kj5248:
## Description

Adding functionality to allow for ArbitraryPlanarPolyhedralXSCircuit to work when PolyhedralPrismCurrentSource has non equal end cap angles. Ideally a way to do it such that it is obvious when done and has to be intentional. Likely add a warning when used for this effect and to point out that there is reduction in accuracy of result when doing this. For use in STEP bluemira so can have TF coils with non-rectangular cross sections.

## Checklist

### DIFF
--- a/bluemira/magnetostatics/circuits.py
+++ b/bluemira/magnetostatics/circuits.py
@@ -58,6 +58,7 @@ class PlanarCircuit(SourceGroup):
         self.d_l = np.diff(self.shape, axis=0)
         self.midpoints = self.shape[:-1, :] + 0.5 * self.d_l
         sources = []
+        warning_called = False
         for midpoint, d_l, beta, alpha in zip(
             self.midpoints, self.d_l, betas, alphas, strict=False
         ):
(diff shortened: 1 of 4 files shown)


bluemira/magnetostatics/circuits.py
```diff
@@ -58,22 +58,44 @@ def _generate_sources(
...
+                source = source_class(
+                    midpoint,
+                    d_l,
+                    normal,
+                    t_vec,
+                    *xs_args,
+                    alpha=alpha,
+                    beta=beta,
+                    current=current,
+                    bypass_endcap_error=True,
```
kj5248, line 77: Not sure this is the right idea as it means if you use polyhedralprism for circuits it will turn off even if you don't want to. Admittedly this will not be for many case (ring or perfectly discretised rectangle) but should I add an extra qualifier to this class to handle this?

je-cook: I think as long it warns its ok, but we just need to make people aware that the polyhedralprism has some sharp edges.

je-cook (approved):
Happy for now hopefully in future we can get rid of these qualifiers

PR #3174, opened 2024-04-12
✨ Add functionality to add custom mapping when running process

athoynilimanew:
## Linked Issues

None

## Description

Currently, to run process radial build with the wrapper and Solver, some default mapping is used. This PR just gives the user the option to create a new dataclass inherited from the defualt one and use it

This a simple change, but in future would need to be done in a better way
## Checklist


bluemira/codes/wrapper.py
```diff
@@ -24,6 +24,7 @@ def systems_code_solver(
     params: ParameterFrame,
     build_config: BuildConfig,
     module: str = "PROCESS",
+    params_cls=None,
```
je-cook, line 27: you'll also need to put the `MappedParameterFrame` import in the TYPE_CHECKING block
```suggestion
    params_cls: type[MappedParameterFrame] | None =None,
```

je-cook (changes requested):
Code looks fine just one minor. Could you also add the new parameter to the docstring of `systems_code_solver`

athoynilimanew:
> Code looks fine just one minor. Could you also add the new parameter to the docstring of `systems_code_solver`

added

PR #3186, opened 2024-04-16
Papercut typing fixes #3022 jmatthews

### LINKED ISSUE #3022: Improve typing across codebase

We should improve typing across the codebase with the eventual aim of turning mypy on in our code qa tests.

## Proposed solution

**Slowly** make PRs solving ~100 type errors from running 
`mypy --config-file pyproject.toml bluemira`

## Alternative solutions

Have bad typing forever...

### Stylistic points:

- No implicit None return types:

bad
```python
def a(var: str) -> None:
     print(var)
```
good
```python
def a(var: str):
    print(var)
```
- Use `numpy.typing` for arrays:

```python
import numpy.typing as npt

def a(var: npt.NDArray[float]):
    ...
```
Unfortunately this is not valid according to mypy therefore we have to use the overly restrictive `np.float64` or possibly build a `np.floating` type but for now use the below:

```python
import numpy.typing as npt

def a(var: npt.NDArray[np.float64]):
    ...
```
there is also `npt.ArrayLike` if a function takes for instance either floats or arrays

- Use of `Any` and `object` typing should be avoided these are very generic.
- Types of variables to functions can be loose but types returned from functions should be explicit and confined
### List of PRs
- [x] #3053 
- [x] #3092 
- [x] #3186 
- [x] #3267
- [x] #3305 
-  [ ] #3975

kj5248:
## Linked Issues

Issue #3022 on fixing typing in codebase

## Checklist

### DIFF
--- a/bluemira/geometry/parameterisations.py
+++ b/bluemira/geometry/parameterisations.py
@@ -18,11 +18,7 @@ from dataclasses import dataclass
 from enum import Enum
 from functools import partial
 from pathlib import Path
-from typing import (
-    Generic,
-    TextIO,
-    TypeVar,
-)
+from typing import TYPE_CHECKING, Generic, TextIO, TypeVar
 
 import matplotlib.pyplot as plt
 import numpy as np
@@ -42,6 +38,9 @@ from bluemira.geometry.tools import (
 from bluemira.geometry.wire import BluemiraWire
 from bluemira.utilities.opt_variables import OptVariable, OptVariablesFrame, VarDictT, ov
 from bluemira.utilities.plot_tools import str_to_latex
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
 
 __all__ = [
     "GeometryParameterisation",
@@ -1372,7 +1371,7 @@ class PictureFrameTools:
         z_top: float,
         z_mid: float,
         ri: float,
-        axis: tuple[float, float, float] = (0, -1, 0),
+        axis: Iterable[float] = (0, -1, 0),
         *,
         flip: bool = False,
     ) -> BluemiraWire:
@@ -1526,7 +1525,7 @@ class PictureFrameTools:
         z: float,
         r_i: float,
         r_o: float,
-        axis: tuple[float, float, float] = (0, 1, 0),
+        axis: Iterable[float] = (0, 1, 0),
         *,
         flip: bool = False,
     ) -> BluemiraWire:
--- a/bluemira/geometry/parameterisations.py
+++ b/bluemira/geometry/parameterisations.py
@@ -19,7 +19,6 @@ from enum import Enum
 from functools import partial
 from pathlib import Path
 from typing import (
-    TYPE_CHECKING,
     Generic,
     TextIO,
     TypeVar,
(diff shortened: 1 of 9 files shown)


bluemira/magnetostatics/biot_savart.py
```diff
@@ -13,6 +13,7 @@
 from typing import TYPE_CHECKING
 
 import numpy as np
+import numpy.typing as npt
```
je-cook, line 16: Its already in the TYPE_CHECKING block
```suggestion
```

bluemira/magnetostatics/circuits.py
```diff
@@ -33,10 +35,16 @@ class PlanarCircuit(SourceGroup):
...
     current: float
 
-    def _generate_sources(self, shape, current, source_class, xs_args):
+    def _generate_sources(
+        self,
+        shape: npt.NDArray[np.float64] | Coordinates,
+        current: float,
+        source_class: Any,
+        xs_args: npt.NDArray[np.float64],
+    ) -> list[Any]:
```
je-cook, line 47: Any is bad don't use Any where possible, is there either a abstract baseclass or a protocol we can create?

bluemira/utilities/opt_variables.py
```diff
@@ -16,6 +16,7 @@
 from typing import TYPE_CHECKING, TextIO, TypedDict
 
 import numpy as np
+import numpy.typing as npt
```
je-cook, line 19: Move to TYPE_CHECKING block

je-cook (changes requested):
One question about axis and Iterable and one about the use of Any, otherwise looks good

bluemira/geometry/parameterisations.py
```diff
@@ -1350,7 +1372,7 @@ def _make_domed_leg(
         z_top: float,
         z_mid: float,
         ri: float,
-        axis: Iterable[float] = (0, -1, 0),
+        axis: tuple[float, float, float] = (0, -1, 0),
```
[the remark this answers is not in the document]
kj5248: When you use Iterable there is a typing mismatch to other files and functions. Wasn't sure whether to change those to Iterable or this to tuple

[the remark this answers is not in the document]
je-cook: I think they should be Iterable too as I think the underlying matplotlib function is agnostic

bluemira/magnetostatics/circuits.py
kj5248: Yeah I tried a couple of things but got a variety of errors. Do I have to go all the way to the baseclass for that?

je-cook: So it need to include every possible source class (eg the base class or a common ancestor) or a protocol which implements stubs of the methods a source will have (see your next papercut for an example)

kj5248: So when I use a common ancestor for PolyhedralPrismCurrentSource and TrapezoidalPrismCurrentSource which is CrossSectionCurrentSource it throws the error that it is not callable and also that "Type[TrapezoidalPrismCurrentSource]" doesn't match "CrossSectionCurrentSource"

je-cook: You need as the type `type[CrossSectionCurrentSource]` because you are passing in an unitialised class

kj5248: Ah I was using type() before. That makes a difference, although it seems to have a weird issue...  error: "TrapezoidalPrismCurrentSource" gets multiple values for keyword argument "alpha",

je-cook: resolved offline

je-cook (approved):

PR #3242, opened 2024-05-02
✅ Add coil names and Types to be written

athoynilimanew:
## Linked Issues

Closes #{3234}

## Checklist


je-cook (dismissed):
This is not ready yet. There needs to be some use of the new functionality in the tests and the difference between a json eqdsk and a normal eqdsk needs to be proven

athoynilimanew:
> This is not ready yet. There needs to be some use of the new functionality in the tests and the difference between a json eqdsk and a normal eqdsk needs to be proven

First Part:
I basically manually changed a test file to include the coilnames and types, and use that file in test_write_then_read_in_json_format().

I was not sure to manipulate existing DN-DEMO_eqref.json file for this so just duplicated it and added the coilnames and types. If you prefer, I can stick to just one, but would first like to know your opinion.

je-cook:
Great thanks, I think the duplication of eqdsk files here is fine. We are trying to show a few things:

1) that we can read in a eqdsk json file with and without coilset name/type information
2) we can write out that same information and it is the same as the original file

It looks like the first is definitely covered by the test but as far as I can see although the file is written out to a mock file the contents of the `written` mock variable is not checked (and wasnt before, which is bad). 

If we could check the contents of `written` is the same as the original file I think that would finish this off.

tests/equilibria/test_file.py
```diff
@@ -198,6 +224,32 @@ def test_write_then_read_in_json_format(self):
...
         assert eq2.nz == 151
         assert eq2.nbdry == 72
+        assert eq2.coil_names is None
+        assert eq2.coil_types is None
+
+        # Test with DN-DEMO_eqref_withCoilNames.json (with coil information)
+        eq3 = EQDSKInterface.from_file(self.testfiles[3])
+
+        with open(str(self.testfiles[3])) as data_file:
+            eq3_string = json.load(data_file)
```
oliverfunk, line 234:
```suggestion
            eq3_json_dict = json.load(data_file)
```

tests/equilibria/test_file.py
```diff
@@ -198,6 +224,32 @@ def test_write_then_read_in_json_format(self):
...
+
+        with mock.patch(OPEN, new_callable=mock.mock_open) as open_mock:
+            eq3.write("some/path.json", file_format="json")
+        written = combine_text_mock_write_calls(open_mock)
+
+        # check the contents of the written mock variable
+        assert_dict_subset(eq3_string, json.loads(written))
```
oliverfunk, line 241: Replace this with the `compare_dicts` function `utilities/tools`.

After looking at it with @je-cook the `eq3_string` and what's returned from `json.loads(written)` should be identical.

Because what is written by the `eq3.write` call, should be identical to the json file on disk already (as it, in theory, should have written it in the past at some point).

oliverfunk (changes requested):

tests/equilibria/test_file.py
athoynilimanew: Is your point that this check is not required?

oliverfunk (approved):
Thanks @athoynilimanew !

PR #3250, opened 2024-05-06
add functions to manipulate (x,y) sets of vectors to enable representation of time-dependent mathematical functions

### LINKED ISSUE #3210: Add `unique_domain` and other functions to pre-process time-evolution data.
Accounting transient power loads, voltages, currents, etc, requires transforming the data that defines them into objects of classes that represent them as time-evolution functions.

Mathematical 1D functions essentially define a mapping of a series of values in a Domain to a series of values in an Image (x -> y). These relationships must obey certain conditions, especially to ensure compatibility with other methods/modules, such as `scipy.interpolate.interp1d`. This also ensures these time-evolution representations may be added, subtracted and so on.

Creation of any object that represents the time-evolution of a physical quantity must pre-process x-values and y-values to make them qualify as domain and image, ensuring the following for each case:
- new object from raw data: the domain must be composed of unique values (no x value is repeated), _e.g._ using a `unique_domain(x,y)` function.
- new object from other objects: the images of multiple objects must be expanded to match a single domain, composed by the union of all individual domains, _e.g._ using a `match_domains(x_set,y_set)`function.

tplobo:
## Linked Issues

Closes #3210

## Checklist


je-cook (changes requested):
Bit of work still to do here. I've also push a small update to the branch to fix the linter problems on the feature branch so you will need to rebase.

Make sure the CI passes.

tplobo:
> Bit of work still to do here. I've also push a small update to the branch to fix the linter problems on the feature branch so you will need to rebase.
> 
> Make sure the CI passes.

Just completed rebase.

bluemira/power_cycle/tools.py
```diff
@@ -31,3 +34,93 @@ def create_axes(ax=None):
     if ax is None:
         _, ax = plt.subplots()
     return ax
+
+
+def unique_domain(x, y, epslon=1e-10, mode="careful"):
```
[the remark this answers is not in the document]
tplobo: Answer to each of your points:
- y is passed to the function to ensure that x and y are a valid pair of domain/image vectors; often you will encounter vectors of different lengths that need to be adjusted, for example by padding one of them. I have just not implemented that yet due to lack of time.
- the slow version is "more correct" just because it does not risk accumulating a nudge that becomes significant when you're dealing with many points... but running it becomes unfeasible when dealing with domain/image sets in the order of hundreds of millions of points that must be down-sampled, especially if you have multiple sets... that is exactly what happened with the original coil supply system data I got... so I still think the fast mode is necessary

bluemira/power_cycle/tools.py
```diff
@@ -31,3 +34,93 @@ def create_axes(ax=None):
...
+        raise ValueError(
+            f"Invalid argument: '{mode}'. The parameter 'mode' can only"
+            f"assume one of the following values: {slow_identifiers} or "
+            f"{fast_identifiers}."
+        )
+    return new_x, new_y
+
+
+def match_domains(
+    x_set: list[np.ndarray],
```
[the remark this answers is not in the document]
tplobo: got it, but domain is not a good word because the function should start with x vectors that are not necessarily a domain and make them into one... I'll think of a better name

bluemira/power_cycle/tools.py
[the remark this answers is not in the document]
je-cook: 1) isnt that what your match_domains does or is meant to do?
2) I'm still on the side of disagree but if you _really_ need it its a kw-only boolean flag that warns when used:
```python
if flag:
    bluemira_warn('be careful with this')  # be slightly more descriptive but should fit on a line
else:
    # normal version
```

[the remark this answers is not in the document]
tplobo: also, the use of pairwise breaks the case in which len(x) = 1; the function should just return the same x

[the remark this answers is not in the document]
je-cook: then the function doesnt do anything and should return early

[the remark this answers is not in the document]
tplobo: > the mode needs to be selected more cleanly. We also do not need to provide short hand selections. The below enum will cover all needs and also remove the need for the ValueError raise in the function.
> 
> ```python
> from enum import Enum, auto
> 
> class UniqueDomainMode(Enum):
>     CAREFUL = auto()
>     QUICK = auto()
>     
>     @classmethod
>     def _missing_(cls, value: str):
>         try:
>             return cls[value.upper()]
>         except KeyError:
>             raise AttributeError(...)
> ```

can we add the [Advanced Enum package](https://pypi.org/project/aenum/) to the requirements? I'd like to use multi-valued enums.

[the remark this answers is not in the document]
tplobo: > 1. isnt that what your match_domains does or is meant to do?

exactly: I was concerned about some edge cases in which one might not call `match_domains`... so I was precisely debating whether to shift this particular result of `interp` into `unique_domain`, and have it called twice... but I have not decided on this yet; seems inefficient to call the same function twice.




> 2. I'm still on the side of disagree, millions of points is not our domain. If you _really_ need it its a kw-only boolean flag that warns when used:
> ```python
> if flag:
>     bluemira_warn('be careful with this')  # be slightly more descriptive but should fit on a line
> else:
>     # normal version
> ```
> 
> I'm still not sold on adding it

millions of points are not calculated by our models, but it is necessary for dealing with data files for testing, verification and generating input files...
when you talk about a flag, you mean adding the fast mode and only use it with a "hidden" key in a kwargs?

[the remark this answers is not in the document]
je-cook: See https://docs.astral.sh/ruff/rules/boolean-default-value-positional-argument/ for boolean args
```python
def func(a: float, *, b: bool = False):
    print(a, b)

func(1, False) # Will error
func(1, b=False) # prints 1 False
```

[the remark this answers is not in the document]
je-cook: If its only for testing and verification dependent on the time it takes, that seems like overkill for our purposes.

bluemira/power_cycle/tools.py
```diff
@@ -31,3 +36,116 @@ def create_axes(ax=None):
     if ax is None:
         _, ax = plt.subplots()
     return ax
+
+
+'''
+from aenum import MultiValueEnum
```
je-cook, line 42: we're not adding that as a dependency

bluemira/power_cycle/tools.py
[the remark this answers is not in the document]
je-cook: Also if you use numba its not slow:

```python
import numba as nb

@nb.jit
def unique_domain(x, epslon=1e-10):
    """
    Ensure x has only unique values to make (Domain: x -> Image: y) a function.

    Epslon must be small enough so that consecutive values can be considered
    equal within the context/scale in which the function is defined.

    Careful mode:
    -------------
    Nudge forward each non-unique element in x by (N * epslon), given N times,
    after the first appearance, that value has appeared in x before.

    Fast mode:
    ---------
    Nudge forward every element in x by (N * epslon), with N being the index
    of that element in x. More prone to errors if the number of elements in
    x is large.
    """
    new_x = [x[0]]
    nudge = 0
    for x_last, x_this in zip(x[:-1], x[1:]):
        if np.isclose(x_last, x_this, rtol=EPS):
            nudge += epslon
            x_this = x_last + nudge
        else:
            nudge = 0
        new_x.append(x_this)
    return np.asarray(new_x)
```
10M points take 105ms

image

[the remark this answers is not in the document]
tplobo: > See https://docs.astral.sh/ruff/rules/boolean-default-value-positional-argument/ for boolean args

I still don't get your point, you suggested the boolean flag but why can't we have a keyword argument with a default value? it will never be modified, unless necessary.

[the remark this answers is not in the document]
je-cook: as per my comment above we dont need fast because slow is fast. Secondly there is only two options fast_on/fast_off and really we dont want people to use the fast option regularly therefore the req of keyword. Thirdly string inputs are a bad design pattern.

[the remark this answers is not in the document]
tplobo: > 10M points take 105ms, if you're using more points than that you need to preprocess the data properly

"pre-process the data properly" is precisely the problem, it is [extremely non-trivial](https://www.mathworks.com/matlabcentral/answers/1454924-downsample-data-adapively-intelligently#answer_789429) and there are multiple potential algorithms that may or may not work... "proper" pre-processing is a type of optimization and that is an open field

this function is what I developed for when someone sends me data from system X in plant design, sometimes with hundreds of points, sometimes with hundreds of millions (_e.g._ data from control engineering)...

if it takes too long to just prepare a representative set of data points before even starting the down-sampling algorithms for comparison, the fast mode in matching the domain of vectors helps a lot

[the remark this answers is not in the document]
je-cook: Just to confirm my point of not needing fast. Fast is only 10% faster while also possibly being wrong. Thats not a good enough reason to include it.

image

tplobo:
> Also if you use numba its not slow:

@je-cook, I've had to stop the RUF007 check on the `zip` call because `numba` apparently cannot work with `pairwise`

bluemira/power_cycle/tools.py
```diff
@@ -31,3 +34,93 @@ def create_axes(ax=None):
...
+    else:
+        raise ValueError(
+            f"Invalid argument: '{mode}'. The parameter 'mode' can only"
+            f"assume one of the following values: {slow_identifiers} or "
+            f"{fast_identifiers}."
+        )
+    return new_x, new_y
+
+
+def match_domains(
```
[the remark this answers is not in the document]
tplobo: take a look at the new version now

bluemira/power_cycle/tools.py
```diff
@@ -31,3 +37,65 @@ def create_axes(ax=None):
     if ax is None:
         _, ax = plt.subplots()
     return ax
+
+
+@nb.jit
+def unique_domain(x: np.ndarray, epsilon: np.number = DEFAULT_EPSILON):
```
je-cook, line 43: the default can just sit in the function definition in both cases, no need for a global variable. Also, the typing should just be float.
```suggestion
def unique_domain(x: np.ndarray, epsilon: float = 1e-10):
```

bluemira/power_cycle/tools.py
```diff
@@ -12,6 +12,12 @@
 from typing import Any
 
 import matplotlib.pyplot as plt
+import numba as nb
+import numpy as np
+
+from bluemira.base.constants import EPS
+
+DEFAULT_EPSILON = 1e-10
```
je-cook, line 20:
```suggestion
```

bluemira/power_cycle/tools.py
```diff
@@ -31,3 +37,65 @@ def create_axes(ax=None):
...
+
+    Epsilon must be small enough so that consecutive values can be considered
+    equal within the context/scale in which the function is defined.
+
+    Nudge forward each non-unique element in x by (N * epsilon), given N times,
+    after the first appearance, that value has appeared in x before.
+    """
+    n_points = len(x)
+    new_x = [x[0]]
+    if n_points > 1:
```
je-cook, line 55: n points is only used in if statement, no need for a variable.
```suggestion
    new_x = [x[0]]
    if  len(x) > 1:
```

bluemira/power_cycle/tools.py
```diff
@@ -31,3 +37,65 @@ def create_axes(ax=None):
...
+    to ensure that all of them have one element associated to every element
+    of the union of all distinct 'x' vectors. Values defined before and after
+    the original domain 'x' of the image 'y' are set to zero.
+    """
+    n_vectors = len(all_x)
+    for v in range(n_vectors):
+        all_x[v] = unique_domain(all_x[v], epsilon=epsilon)
+
+    x_matched = np.concatenate(all_x)
+    x_matched = np.unique(x_matched)
```
je-cook, line 91:
```suggestion
    x_matched = np.unique(np.concatenate(all_x))
```

bluemira/power_cycle/tools.py
```diff
@@ -31,3 +37,65 @@ def create_axes(ax=None):
...
+    all_y_matched = all_y.copy()
+
+    for v in range(n_vectors):
+        x = all_x[v]
+        y = all_y[v]
+
+        y = np.interp(x_matched, x, y, left=0, right=0)
+        all_y_matched[v] = y
```
je-cook, line 99:
```suggestion
        all_y_matched[v] =  np.interp(x_matched, all_x[v], all_y[v], left=0, right=0)
```

bluemira/power_cycle/tools.py
```diff
@@ -31,3 +37,65 @@ def create_axes(ax=None):
...
+                nudge = 0
+            new_x.append(new_x_this)
+
+    return np.asarray(new_x)
+
+
+def match_domains(
+    all_x: list[np.ndarray],
+    all_y: list[np.ndarray],
+    epsilon: np.number = DEFAULT_EPSILON,
```
je-cook, line 72:
```suggestion
    epsilon: float = 1e-10,
```

tests/power_cycle/test_tools.py
```diff
@@ -0,0 +1,54 @@
...
+
+@pytest.mark.parametrize("epsilon", [1e-10, 1e-7, 1e-5])
+def test_timing_unique_domain(epsilon):
+    n_repeat = 1000
+    times = ti.Timer(partial(test_unique_domain, epsilon)).repeat(
+        repeat=10, number=n_repeat
+    )
+    avg_time = min(times) / n_repeat
+    print(f"For epsilon={epsilon}, avg.'unique_domain' runtime: {avg_time} s")
+    assert avg_time < 1
```
je-cook, line 38: We dont need to explicitly timing test code
```suggestion
```

bluemira/power_cycle/tools.py
```diff
@@ -31,3 +37,65 @@ def create_axes(ax=None):
...
+    all_y_matched = all_y.copy()
+
+    for v in range(n_vectors):
+        x = all_x[v]
+        y = all_y[v]
+
+        y = np.interp(x_matched, x, y, left=0, right=0)
+        all_y_matched[v] = y
+
+    return x_matched, all_y_matched
```
je-cook, line 101: If y is longer than x this will return two uneven lengthed arrays. That feels odd. Is there a good reason to need this? I dont see there being a reasonable one that would not be a possible location of a bug in users code.

 ```suggestion
    return np.unique(np.concatenate(all_x)), np.asarray([
        np.interp(x_matched, all_x[v], all_y[v], left=0, right=0)
        for v in range(n_vectors)
    ])
```

**Review by je-cook — CHANGES_REQUESTED** [review]
A few more things

File: bluemira/power_cycle/tools.py (hunk 7 above)
Comment (tplobo) [reply bluemira/power_cycle/tools.py:55]: it's legacy, I used it more before the last modifications you suggested

File: tests/power_cycle/test_tools.py (hunk 11 above)
Comment (tplobo) [reply tests/power_cycle/test_tools.py:38]: Ah, sure, I was just checking if what you told me last week was for some reason an issue with my computer. You were right, it is fast enough not to warrant the  "fast mode".

File: bluemira/power_cycle/tools.py (hunk 12 above)
Comment (tplobo) [reply bluemira/power_cycle/tools.py:101]: if `y` is longer than `x`, `interp` fails because it needs even-lengthed domain/image; it will automatically issue an exception

**Comment by tplobo** [comment]
@je-cook, I think I managed to fix `unique_domain` for those edge cases I mentioned... I have expanded the tests to include them, and the new function passes them. Ensuring the new function works with `numba` was a pain though...
Let me know what you think now!

**Comment by je-cook** [comment]
Can you rebase this so its easier for me to have a look?

**Comment by je-cook** [comment]
@tplobo your rebase didnt go correctly the following should fix it for you:

1) Checkout and pull `feature/powercycle`
2) Re-checkout `tplobo/powercycle_tools`
3) `git rebase --onto feature/powercycle 335672b9` (the commit before this branch is currently)
4) there will be a conflict on the powercycle base commit, skip it  `git rebase --skip`
5) `git push --force` the result

Thanks!

File: tests/power_cycle/.gitignore [hunk 13]
```diff
@@ -0,0 +1,2 @@
+# Images
+*.svg
```
je-cook, line 2: We shouldnt need to do this because all test output files should be saved to `generated_data/test_data`. This filepath is set up to be given by the `tmp_path` fixture. See https://docs.pytest.org/en/8.2.x/how-to/tmp_path.html for usage examples.

Please remove

bluemira/power_cycle/tools.py
```diff
@@ -31,3 +35,92 @@ def create_axes(ax=None):
...
+    if len(x) > 1:
+        for x_this in x[1:]:
+            nudge = 0
+            new_x_this = x_this
+            needs_nudge = _nudge_check(new_x[-1], x_this)
+            if needs_nudge:
+                while needs_nudge:
+                    nudge += epsilon
+                    new_x_this = x_this + nudge
+                    needs_nudge = _nudge_check(new_x[-1], new_x_this)
```
je-cook, line 86: This could easily be a near infinite loop. Maybe something like the below at least makes the function loops a known max.
```suggestion
           if _nudge_check(new_x[-1], new_x_this):
                for _ in range(10):
                    nudge += epsilon
                    new_x_this = x_this + nudge
                    if _nudge_check(new_x[-1], new_x_this):
                        break
                else:
                    raise ValueError("Failed to create unique domain")
```

bluemira/power_cycle/tools.py
```diff
@@ -31,3 +35,92 @@ def create_axes(ax=None):
...
+            else:
+                new_x_this = x_this
+            new_x.append(float(new_x_this))
+    validate_monotonic_increase(new_x, strict_flag=True)
```
je-cook, line 90: Do we really need this and the nudge check? We've already checked at the beginning. I cant see a way this wouldnt be always true, especially with nudge?

bluemira/power_cycle/tools.py
```diff
@@ -31,3 +35,92 @@ def create_axes(ax=None):
...
+    on 'x_matched' later can neglect step functions).
+
+    Then, for each pair (x,y), values in every 'y' vector are interpolated,
+    to ensure that all of them have one element associated to every element
+    of the union of all distinct 'x' vectors. Values defined before and after
+    the original domain 'x' of the image 'y' are set to zero.
+    """
+    n_vectors = len(all_x)
+    for v in range(n_vectors):
+        numba_safe_x = nb.typed.List([float(x) for x in all_x[v]])
```
je-cook, line 113: This will fail when numpy v2 is released because float(x) when x is a single element array is no longer allowed, use `x.item()`.

je-cook (changes requested):

tests/power_cycle/.gitignore
tplobo: ufff... this took me longer than I care to admit to learn how to use... XD

bluemira/power_cycle/tools.py
tplobo: There was a `not` missing in your suggestion, but it is implemented.

bluemira/power_cycle/tools.py
tplobo: I require the `float(x)` for `numba`; is this what you meant: `float(x.item())`?

tplobo: Ocean sent this suggestion: `np.array([x]).item`, but that does not ensure it to become a float, right? Can I float the output of that?
Current state of the code:
`numba_safe_x = nb.typed.List([float(x) for x in all_x[v]])`

bluemira/power_cycle/tools.py
je-cook: It looks mathematically impossible for this to ever fail because you've already done a nudge check with a monotonically increasing nudge. Even if points reverse it wouldnt matter because we are dealing with x not y. Even if this is run with y it would have so little effect to be irrelevant.

tests/power_cycle/.gitignore
je-cook: still need to remove this file!

tplobo: I had removed it from the coils branch instead, sorry!

bluemira/power_cycle/tools.py
tplobo: I agree, it's been removed

je-cook (approved):

PR #3292, opened 2024-05-20
SymmetricCircuits & Current opt state vector reduction

oliverfunk:
## Linked Issues

Closes #{ID}

## Description

Briefly, here is a proof for why the derivative values can be added when you transform a function into a new function by setting some parameters to equal to each other:

Say we have some scalar function f of two variables x and y:

image

The gradient of f is:

image

Say we now have a new function u that sets y = x:

image

Then the gradient of u can then be expressed as:

image

An example

Say we have f:

image

Then we have u:

image

One can see, if you substitute y=x into the grad f function you get grad u.

## Checklist

### DIFF
--- a/bluemira/equilibria/coils/_grouping.py
+++ b/bluemira/equilibria/coils/_grouping.py
@@ -1193,7 +1193,7 @@ class CoilSet(CoilSetFieldsMixin, CoilGroup):
         """
         cc = self.get_control_coils()
         xs, zs = cc._get_opt_positions(position_coil_names)
-        currents = cc.current / current_scale
+        currents = cc._opt_currents / current_scale
         return CoilSetOptimisationState(
             currents=currents,
             xs=xs,
@@ -1220,7 +1220,7 @@ class CoilSet(CoilSetFieldsMixin, CoilGroup):
         """
         cc = self.get_control_coils()
         if opt_currents is not None:
-            cc.current = opt_currents * current_scale
+            cc._opt_currents = opt_currents * current_scale
         if coil_position_map is not None:
             cc._set_opt_positions(coil_position_map)
 
--- a/bluemira/equilibria/coils/_grouping.py
+++ b/bluemira/equilibria/coils/_grouping.py
@@ -1338,37 +1338,6 @@ class CoilSet(CoilSetFieldsMixin, CoilGroup):
         return mat
 
     @property
-    def _opt_currents_sym_reduce_mat(self) -> np.ndarray:
-        """
-        Get the optimisation currents symmetry reduce matrix.
-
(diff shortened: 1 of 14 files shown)


oliverfunk:
This is nearly ready.

I need to redo the `CoilSetSymmetryStatus` feature, as will not work properly if one uses a `Circuit` instead of Sym. Circuit.

bluemira/equilibria/coils/_grouping.py
```diff
@@ -1184,10 +1184,143 @@ def set_optimisation_state(
...
+            for c in self._coils
+        ]
+        return [*flatten_iterable(optimisable_coil_names)]
+
+    @property
+    def all_current_optimisable_coils(self) -> list[Coil]:
+        """
+        Get the names of all coils that can be current optimised.
+        """
+        return [self[cn] for cn in self.current_optimisable_coil_names]
```
geograham, line 1214: """ Get list of all coils that can be current optimised."""

bluemira/equilibria/coils/_grouping.py
```diff
@@ -1184,10 +1184,143 @@ def set_optimisation_state(
...
+
+    def get_current_optimisable_coils(
+        self, coil_names: list[str] | None = None
+    ) -> list[Coil]:
+        """
+        Get the coils that can be current optimised.
```
geograham, line 1220: Should this be changed to something like 'get list of selected coils from coils that can be current optimised'?

bluemira/equilibria/optimisation/problem/base.py
```diff
@@ -189,26 +199,58 @@ def update_magnetic_constraints(
...
+
             d: ConstraintT = {
                 "name": f_constraint.name,
-                "f_constraint": f_constraint.f_constraint,
-                "df_constraint": getattr(f_constraint, "df_constraint", None),
+                "f_constraint": f_c,
+                "df_constraint": df_c,
                 "tolerance": constraint.tolerance,
             }
             # TODO: tidy this up, so the interface guarantees this works!
```
geograham, line 248: Just wondering what this #TODO is from? Do we still need it?

geograham:
Hello, looks good, I think that you should also update the 'Circuits' section of the docs. Maybe with a note on what a a fully, partial and non-symmetric circuit is, a diagram, and the f and df modifications?

bluemira/equilibria/optimisation/problem/base.py
oliverfunk: Yeah.. It's because the ABC doesn't have a good interface. It should all be improved.

geograham (dismissed):

je-cook (approved):
Congrats!

Issue #870, opened 2022-03-28
Magnetic field side name in advective_transport.py


DarioV86:
It is just a minor issue, how we use `lfs` and `hfs` in `advective_transport.py`.

I understand it is the easiest way to make the solver flexible and usable for both single null and double null, but it could be confusing.
For someone who never read the code, seeing `lfs` and `hfs` used as a sort of bottom and top could be misleading.

Could we use, for instance, ob_down and ob_up (as well as ib_down and ib_up) meaning that we are moving downward and upward from the upstream?

PR #3335, opened 2024-06-04
Tests for advective transport in radiation transport

### LINKED ISSUE #3329: Add more tests for radiation source model: advective_transport
From issue #3212. Add more tests for advective_transport.py. Current breakdown of functions within file that are tested (in test_advective_transport):

ChargedParticleSolver
- [x]  flux_surfaces
- [x] _check_params
- [x] _process_first_wall
- [x] _get_arrays
- [x] _make_flux_surfaces_ob
- [x] _make_flux_surfaces_ib
- [x] _clip_flux_surfaces
- [x] analyse
- [x] _analyse_SN
- [x] _analyse_DN
- [x] _q_par
- [x] plot
- [x] _make_params

kj5248:
## Linked Issues

Closes #3329 

## Description

Adds tests to complete coverage of advective transport.

### DIFF
--- a/tests/radiation_transport/test_advective_transport.py
+++ b/tests/radiation_transport/test_advective_transport.py
@@ -14,7 +14,10 @@ import bluemira.radiation_transport.flux_surfaces_maker as fsm
 from bluemira.base.file import get_bluemira_path
 from bluemira.equilibria.equilibrium import Equilibrium
 from bluemira.geometry.coordinates import Coordinates
-from bluemira.radiation_transport.advective_transport import ChargedParticleSolver
+from bluemira.radiation_transport.advective_transport import (
+    ChargedParticleSolver,
+    ChargedParticleSolverParams,
+)
 from bluemira.radiation_transport.error import AdvectionTransportError
 
 TEST_PATH = get_bluemira_path("radiation_transport/test_data", subfolder="tests")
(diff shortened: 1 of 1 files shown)


tests/radiation_transport/test_advective_transport.py
```diff
@@ -134,6 +164,37 @@ def test_plotting(self):
         ax = self.solver.plot(show=True)
         assert len(ax.lines) > 2
 
+    def test_analyse_SN(self, caplog):
```
kj5248, line 167: Not sure this is needed as the function is used within analyse for the single null class which this is within. However that should also be the case for the double null function using analyse within the double null class, which seemed to need the extra test.

tests/radiation_transport/test_advective_transport.py
```diff
@@ -134,6 +164,37 @@ def test_plotting(self):
...
+        fw = deepcopy(self.solver.first_wall)
+        self.solver.flux_surfaces_ob_hfs = []
```
je-cook, line 169: If you look at the naming in the double null test, it has changed slightly as per #870 so I think it will be `ob_up` and `ob_down`

tests/radiation_transport/test_advective_transport.py
je-cook: It may be overkill but we could refactor that at a later date but I'm not too worried right now

Issue #3331, opened 2024-06-03
Add more tests for radiation source model: flux_surfaces_maker


kj5248:
From issue #3212. Add tests for flux_surfaces_maker.py. Currently no tests.

PR #3365, opened 2024-06-14
add some tests for flux surface maker

geograham:
## Linked Issues

Papercut  #3331

## Checklist

### DIFF
--- a/tests/radiation_transport/test_flux_surface_maker.py
+++ b/tests/radiation_transport/test_flux_surface_maker.py
@@ -29,8 +29,8 @@ TEST_PATH = get_bluemira_path("equilibria/test_data", subfolder="tests")
 
 def test_process_first_wall():
     converted_from_open_cw = _process_first_wall(
-        Coordinates({"x" : [1, 1, 2, 2], "z" : [1, 2, 2, 1]})
-        )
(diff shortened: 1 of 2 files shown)


tests/radiation_transport/test_flux_surface_maker.py
```diff
@@ -0,0 +1,114 @@
...
+        lower_fs_ib, upper_fs_ib = _make_flux_surfaces_ibob(
+            0.001,
+            self.eq,
+            self.o_point,
+            self.yz_plane,
+            x_sep_imp,
+            x_out_imp,
+            outboard=False,
+        )
+        assert len(lower_fs_ob) == 1070
```
je-cook, line 105: Could I have a one line comment or two telling me what this is testing because its not obvious to me what the benefit of having both the below blocks is?
Ie why would I be interested in the length of the flux surface?
If its just a way to run the bit of code, thats fine but is there any information about the flux surface that is interesting to test?

je-cook (approved):
Just the one comment but otherwise good

tests/radiation_transport/test_flux_surface_maker.py
geograham: I have added a comment - it is a check to make sure you get the same number of flux surfaces returned as expected for your list.

PR #3375, opened 2024-06-18
Replace pyclipper -> pyclipr

### LINKED ISSUE #3197: Look into clipper 2 and pyclipr as a pyclipper upgrade

The author of clipper has released a new version of the library clipper 2 which he suggests in much better than the original version. We use a python wrapper pyclipper to wrap clipper 1 for our fallback offsetting. It is probably worth having a look at clipper 2.

https://github.com/AngusJohnson/Clipper2

and the 3rd party python wrapper

https://github.com/drlukeparry/pyclipr 

## Proposed solution

Consider whether we replace pyclipper with pyclipr to get access to the improvements in clipper 2

oliverfunk:
## Linked Issues

Closes #3197 

## Description

Moves to the new pyclipr package based on Clipper2 for the fallback offsetting library.

## Interface Changes

None, only internal api's.

## Checklist


oliverfunk:
2 problems:

- Seemingly cannot set Miter limit
- The accuracy of the returned offset no longer passes the "isclose" checks unless the precision is reduced

je-cook:
Cool! how do you feel about the package, good change?

oliverfunk:
I've realised after testing, this seemingly cannot do negative offsets.

CoronelBuendia:
@je-cook I think I've cracked this nut. There's this one major oddity regarding `arcTolerance` but I've added tests over a big range of offset values and my solution seems to work well.

I can't say whether or not it is more robust or faster than pyclipper, tbh.

I do think this should be experimented with more seriously before being merged, it has the potential to mess up some fallback offset operations by making polygons with many sides which are then difficult for subsequent operations. This could disrupt people's workflows.

je-cook:
> @je-cook I think I've cracked this nut. There's this one major oddity regarding `arcTolerance` but I've added tests over a big range of offset values and my solution seems to work well.
> 
> I can't say whether or not it is more robust or faster than pyclipper, tbh.
> 
> I do think this should be experimented with more seriously before being merged, it has the potential to mess up some fallback offset operations by making polygons with many sides which are then difficult for subsequent operations. This could disrupt people's workflows.

Interesting. Yes we should battle test it first definitely

tests/geometry/test_pyclipper_offset.py
```diff
@@ -42,12 +43,19 @@ def test_complex_polygon(self, x, y, delta, method):
...
+            abs(delta),
+            rtol=1e-5,  # ~5 significant digits
+            atol=1e-5,  # ~ 0.01 mm tolerance
+        )
 
     @pytest.mark.parametrize("method", options)
     def test_complex_polygon_overoffset_raises_error(self, method):
         coordinates = Coordinates({"x": self.x, "y": self.y, "z": 0})
         with pytest.raises(GeometryError):
+            # this does not raise because it does not do negative offsets
```
je-cook:
```suggestion
```

I assume this now works as the tests pass

PR #3391, opened 2024-06-24
🐛 Catch debug folder path doesnt exist

je-cook:
## Description

If bluemira is installed to site-packages this folder wont exist therefore we make one in the cwd if required.

## Checklist

### DIFF
--- a/bluemira/base/logs.py
+++ b/bluemira/base/logs.py
@@ -36,7 +36,7 @@ class LogLevel(Enum):
             return cls(value * 10)
         try:
             return cls[value.upper()]
-        except KeyError:
+        except (KeyError, AttributeError):
             raise LogsError(
                 f"Unknown severity level: {value}. Choose from: {(*cls._member_names_,)}"
             ) from None
--- a/bluemira/base/logs.py
+++ b/bluemira/base/logs.py
@@ -7,12 +7,17 @@
 
 """Logging system setup and control."""
 
+from __future__ import annotations
+
 import logging
 import sys
-from collections.abc import Iterable
 from enum import Enum
+from typing import TYPE_CHECKING
 
 from bluemira.base.error import LogsError
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
 
 
 class LogLevel(Enum):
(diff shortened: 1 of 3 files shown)


oliverfunk (commented):
While the changes made are good, I don't believe it's necessary to save the underlying geometry as a default for failed ops.

As an end user, one does not really care. If the op fails, the pyclipper fallback is used and a warning is logged.

This feels more like something that was needed when Bluemira was first being written or in fairly extreme debugging situations.

I think the params for `interpolate_bspline` and `offset_wire` functions should have a save_failed_offset: bool option (that defaults to False, that would enable this when desired.

je-cook:
As discussed changed slightly the serialisation on error behaviour

bluemira/geometry/tools.py
```diff
@@ -151,22 +157,26 @@ def wrapper(*args, **kwargs):
...
-                bluemira_debug(
-                    f"Function call {func_name} failed. Debugging information was saved"
-                    f" to: {filename}"
-                )
-            except Exception:  # noqa: BLE001
-                bluemira_warn(
-                    f"Failed to save the failed geometry operation {func_name} to JSON."
-                )
+            else:
+                bluemira_warn("Set logging level to debug to serialise function call")
```
oliverfunk, line 179: I think this could be clearer

"Set logging level to debug to save geometry used in offset"

Or something slightly more descriptive.

bluemira/base/logs.py
```diff
@@ -93,8 +105,12 @@ def set_log_level(
         logger = logging.getLogger(logger_name)
 
         current_level = logger.getEffectiveLevel() if increase else 0
-        new_level = _convert_log_level(verbose, current_level)
-        _modify_handler(new_level, logger)
+        _modify_handler(
+            LogLevel(verbose)
+            if isinstance(verbose, str)
+            else LogLevel(int(current_level + (verbose * 10))),
```
oliverfunk, line 111: I understand logging is very jank at the moment but I'm struggling to read the whole   log level *10 I've seen in the PR.

Esp. here with int(current_level + (verbose * 10))

If this kinda of math is needed for some logging stuff, it should be abstracted away into a function. Otherwise it's difficult to read and understand why/what's going on.

oliverfunk (commented):
Some fairly minor ones.

bluemira/base/logs.py
je-cook: Issue opened in #3399

PR #3394, opened 2024-06-25
Add radiation tools tests

### LINKED ISSUE #3330: Add more tests for radiation source model: radiation_tools
From issue #3212. Add more tests for radiation_tools.py. Current breakdown of functions within file that are tested:

- [x] upstream_temperature (tested in test_radiation_profile)
- [x] target_temperature (tested in test_radiation_profile)
- [ ] specific_point_temperature
- [x] electron_density_and_temperature_sol_decay (tested in test_radiation_profile)
- [x] gaussian_decay
- [x] exponential_decay
- [x] ion_front_distance (tested in test_radiation_profile)
- [x] calculate_z_species
- [x] radiative_loss_function_values
- [ ] radiative_loss_function_plot
- [x] calculate_line_radiation_loss
- [x] linear_interpolator
- [x] interpolated_field_values
- [ ] grid_interpolator
- [ ] pfr_filter
- [ ] filtering_in_or_out
- [ ] get_impurity_data
- [ ] detect_radiation
- [ ] make_wall_detectors
- [ ] plot_radiation_loads
- [ ] FirstWallRadiationSolver

athoynilimanew:
## Description

Added tests for the following

- [x] interpolated_field_values
- [x] filtering_in_or_out
- [x] pfr_filter
- [x] make_wall_detectors
- [x] FirstWallRadiationSolver

## Checklist

### DIFF
--- a/tests/radiation_transport/test_radiation_profile.py
+++ b/tests/radiation_transport/test_radiation_profile.py
@@ -351,3 +351,7 @@ class TestCoreRadiation:
 
         # check if the arrays are of same length
         assert len(wall_loads.detector_numbers) == len(wall_loads.detected_power)
+
+        # the solver gives slightly different powers in each run
+        # so just asserting the order of total power
+        assert 10**8 <= wall_loads.total_power <= 10**9
--- a/tests/radiation_transport/test_radiation_profile.py
+++ b/tests/radiation_transport/test_radiation_profile.py
@@ -333,8 +333,8 @@ class TestCoreRadiation:
         )
 
         # SOL radiation grid
-        x_sol = np.linspace(min(self.fw_shape.x), max(self.fw_shape.x), 1000)
-        z_sol = np.linspace(min(self.fw_shape.z), max(self.fw_shape.z), 1500)
+        x_sol = np.linspace(min(self.fw_shape.x), max(self.fw_shape.x), 4)
+        z_sol = np.linspace(min(self.fw_shape.z), max(self.fw_shape.z), 4)
 
         rad_sol_grid = interpolated_field_values(x_sol, z_sol, f_sol)
         func = grid_interpolator(x_sol, z_sol, rad_sol_grid)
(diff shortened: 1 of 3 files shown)


OceanNuclear (approved):
A little comment/documentation would help but code-wise, this looks good :)

je-cook (changes requested):
I agree with @OceanNuclear's comment.

Also this doesnt close #3330, there are some more tests to be written specifically something that calls FirstWallRadiationSolver. It should be a very simple case with only a couple of detectors.

athoynilimanew:
> I agree with @OceanNuclear's comment.
> 
> Also this doesnt close #3330, there are some more tests to be written specifically something that calls FirstWallRadiationSolver. It should be a very simple case with only a couple of detectors.

Added test for FirstWallRadiationSolver (but the test requires a bit of time to run)

tests/radiation_transport/test_radiation_profile.py
```diff
@@ -296,3 +305,49 @@ def test_radiative_loss_function_values(self):
...
+        assert solver.fw_shape == self.fw_shape
+
+        wall_loads = solver.solve(50, 1, 10, plot=False)
+
+        # check return types
+        assert isinstance(wall_loads, DetectedRadiation)
+        assert isinstance(wall_loads.total_power, float)
+
+        # check if the arrays are of same length
+        assert len(wall_loads.detector_numbers) == len(wall_loads.detected_power)
```
je-cook, line 353: the asserts here are no thorough enough for a test. At the moment it is just essentially checking it runs which is not enough for a good test.

Can you add some more information that we can assert on

je-cook (changes requested):
Overall nearly there just the one comment, I've added a few customisations to the Solver to allow us to customise it to make the test quicker.

This will also need a rebase for the tests to pass because of unrelated bugs that are now fixes

tests/radiation_transport/test_radiation_profile.py
athoynilimanew: The problem is, the solver do not always give exact same results (e.g. wall_loads.total_power gives different values in different runs). I have now added an assertion to check if the wall_loads.total_power gives a value of a certain order.

PR #3396, opened 2024-06-26
✨ Add Neutronics Manager to eudemo

je-cook:
## Description

Initial neutronics manager, for now just stores stuff and allows plotting of openmc figure

## Checklist

### DIFF
--- a/eudemo/eudemo/model_managers.py
+++ b/eudemo/eudemo/model_managers.py
@@ -19,7 +19,9 @@ from PIL import Image
 from bluemira.base.look_and_feel import bluemira_warn
 
 if TYPE_CHECKING:
+    from bluemira.codes.openmc.output import OpenMCResult
     from bluemira.equilibria.run import Snapshot
+    from eudemo.eudemo.neutronics.run import EUDEMONeutronicsCSGReactor
 
 
 class EquilibriumManager:
@@ -58,7 +60,11 @@ class EquilibriumManager:
 class NeutronicsManager:
     """Manager for neutronics"""
 
-    def __init__(self, csg_reactor, results):
+    def __init__(
+        self,
+        csg_reactor: EUDEMONeutronicsCSGReactor,
+        results: OpenMCResult | dict[int, float],
+    ):
         self.csg_reactor = csg_reactor
         self.results = results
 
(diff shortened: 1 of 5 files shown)


eudemo/eudemo/model_managers.py
```diff
@@ -48,3 +53,31 @@ def get_state(self, name: str) -> None | Snapshot:
         Get an equilibrium state from the Equilibrium manager.
         """
         return self.states.get(name, None)
+
+
+class NeutronicsManager:
+    """Manager for neutronics"""
+
+    def __init__(self, csg_reactor, results):
```
oliverfunk, line 61: Some type would be really useful.

eudemo/eudemo/model_managers.py
```diff
@@ -48,3 +53,31 @@ def get_state(self, name: str) -> None | Snapshot:
...
+        self.csg_reactor = csg_reactor
+        self.results = results
+
+    def plot(self):
```
oliverfunk, line 65: Also maybe a tablulate() method

eudemo/eudemo/reactor.py
```diff
@@ -480,14 +481,18 @@ def build_radiation_plugs(params, build_config, cr_ports, radiation_xz_boundary)
...
+        *run_neutronics(
+            reactor_config.params_for("Neutronics"),
+            reactor_config.config_for("Neutronics"),
+            blanket=reactor.blanket,
+            vacuum_vessel=reactor.vacuum_vessel,
+            ivc_shapes=ivc_shapes,
+        )
     )
 
+    reactor.neutronics.plot()
```
oliverfunk, line 494: perhaps behind some neutronics "plot" config

eudemo/eudemo/model_managers.py
je-cook: I've just added a __str__ method so that you can just print the manager. This is because I for a reason I cant remember left the results underlying tabulate as a private method.

Secondly the volume calculation stuff has not been full fed through (this is where the dict type comes from. So I've ignored that for now but it should be integrated at some point

oliverfunk: Umm okay, I'm not fully following.

The __str__ is fine/good but I think it is much more intuitive to call .tabulate() then print(str(...))

Why not have both?

oliverfunk: Or as you have it

`bluemira_print_clean(f"{reactor.neutronics}")`

could be something like bluemira_print_clean(str(self)

oliverfunk (approved):
Small comment

PR #3453, opened 2024-07-19
Corrections and example updated for SH 

geograham:
## Description

This fixes errors in the constraint and adds how to use it to the already existing notebook. 

## Checklist

### DIFF
--- a/bluemira/equilibria/optimisation/harmonics/harmonics_approx_functions.py
+++ b/bluemira/equilibria/optimisation/harmonics/harmonics_approx_functions.py
@@ -37,6 +37,35 @@ from bluemira.geometry.face import BluemiraFace
 from bluemira.geometry.tools import boolean_cut, make_polygon
 
 
+def ten_power(x):
+    """Get the power for the base ten notation, set 0 to 0."""
+    tp = np.array(np.floor(np.log10(np.abs(x))))
+    # zero correct
+    tp[(x == 0)] = 0.
+    return tp
+
+
+def sig_fig_round(x, s, low_lim=-16):
+    """
+    Fuction to round to a given number of significant figures,
+    with any number below a lower limit set to zero.
+
+    Parameters
+    ----------
+    x:
+        value or values to round.
+    s:
+        number of significant figures
+    low_lim:
+        power below which values are set to 0,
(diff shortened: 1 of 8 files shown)


bluemira/equilibria/optimisation/harmonics/harmonics_approx_functions.py
```diff
@@ -110,7 +110,7 @@ def coil_harmonic_amplitude_matrix(
         / np.sqrt(degrees * (degrees + 1))
     )
 
-    return currents2harmonics
+    return np.array(currents2harmonics, dtype=np.float32)
```
je-cook, line 113: I think downcasting is wrong here. It will lead to hard to track down problems elsewhere because the arrays will then be upcast to float64 as some point. 

image

je-cook:
also #3454 has finally managed to raise the error in the CI that I've been seeing for months. It might actually be connected to the psi changes though...
https://github.com/Fusion-Power-Plant-Framework/bluemira/actions/runs/10035990810/job/27732910420

bluemira/equilibria/optimisation/harmonics/harmonics_approx_functions.py
geograham: Thanks @je-cook - I have removed this and round instead.

bluemira/equilibria/optimisation/harmonics/harmonics_constraint_functions.py
```diff
@@ -49,9 +49,9 @@ def f_constraint(self, vector: npt.NDArray) -> npt.NDArray:
...
         currents = self.scale * vector
 
-        result = self.a_mat[1:,] @ currents
+        result = self.a_mat[:,] @ currents
         return result - self.b_vec - self.value
 
     def df_constraint(self, _vector: npt.NDArray[np.float64]) -> npt.NDArray[np.float64]:
         """Constraint derivative"""
-        return self.scale * self.a_mat[1:,]
+        return self.scale * self.a_mat[:,]
```
je-cook, line 57: No need now, this will actually add an extra copy in here I think
```suggestion
        return self.scale * self.a_mat
```

je-cook (approved):
Thanks looks good to me!

PR #3454, opened 2024-07-22
Thread in semianalytic psi into coil calculations

### LINKED ISSUE #191: Thread in semianalytic psi for use in Coil
In GitLab by @CoronelBuendia on May 18, 2021, 09:43

Now that we have accurate psi resolution inside rectangular cross-section PF coils, need to thread this into Coil via `control_psi` etc.

This will change the amount of O and X points that are found inside the coils when solving equilibria, which has the potential to change the results from equilibria.

je-cook:
## Linked Issues

Closes #191 

## Description

Threads in the semi analytic psi function into the coil field calculations, and some cleanup of unneeded functions.

This should be thoroughly benchmarked before merge.

## Checklist

### DIFF
--- a/bluemira/equilibria/coils/_field.py
+++ b/bluemira/equilibria/coils/_field.py
@@ -268,7 +268,7 @@ class CoilGroupFieldsMixin:
         semianalytic_func: Callable,
     ) -> float | np.ndarray:
         """
-        Combine semianalytic and greens function calculation of magnetic field
+        Combine semianalytic and greens function calculation of [psi, Bx, Bz]
 
         Used for situation where there are calculation points both inside and
         outside the coil boundaries.
@@ -385,7 +385,7 @@ class CoilGroupFieldsMixin:
         _quad_weight: np.ndarray | None = None,
     ) -> float | np.ndarray:
         """
-        Calculate magnetic field B response at (x, z) due to a unit
+        Calculate [psi, Bx, Bz] response at (x, z) due to a unit
         current using Green's functions.
 
         Parameters
@@ -407,7 +407,7 @@ class CoilGroupFieldsMixin:
 
         Returns
         -------
-        Magnetic field response
+        Greens response
         """
         if not split:
             _quad_x = self._quad_x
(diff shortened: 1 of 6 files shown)


je-cook:
blocked on #3453

bluemira/equilibria/coils/_field.py
```diff
@@ -1026,14 +822,18 @@ def _combined_control(
...
         if np.any(points):
-            response[points] = semianalytic_func(x[points], z[points])
+            response[points] = self._response_analytical(
+                semianalytic_func, x[points], z[points]
+            )
 
         return response
 
-    def _B_response_analytical(
+    def _response_analytical(
```
geograham, line 836: I think I may have missed something, but can we not use _response_analytical from CoilGroupFieldsMixin in CoilFieldsMixin's _combined_control?

geograham (commented):
Tests pass and tidy 👍

bluemira/equilibria/coils/_field.py
je-cook: Those three functions on CoilFieldsMixin are needed to get the right shaped array into other functions if you are using just the one coil (not wrapped in a coilgroup). 

It might be possible to adjust this but eg the x z positions are not arrays on Coils whereas they are on all of the grouping objects. This was so that you could do coil.x and get a number and not an array which at the time was a desired thing. 

I did try previously to reduce this but this was the best I could do.

bluemira/equilibria/coils/_field.py
```diff
@@ -145,16 +106,112 @@ def Bz_response(
...
-        )
+        return self._mix_control_method(x, z, greens_Bz, semianalytic_Bz)
 
     def Bp(self, x: float | np.ndarray, z: float | np.ndarray):
         """
         Calculate poloidal magnetic field Bp at (x, z)
         """
         return np.hypot(self.Bx(x, z), self.Bz(x, z))
 
+    def F(self, eqcoil: CoilGroup) -> np.ndarray:
```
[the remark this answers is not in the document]
je-cook: this was just moved to a more sane order in the class. I don't know where the name comes from

geograham (approved):

oliverfunk (approved):

PR #3518, opened 2024-08-12
🎨 Control coil enforcement, constraint stabilisation and general cleanup

je-cook:
## Description

- Fixes some bugs where control coils were not explicitly selected in parts of the equilibria optimisation problems.
- Added rounding to field and force constraints (to 16 dp) which seems to stabilise them in local testing. I have it on good authority that the condition matrix of force constraint is infinite therefore very unstable. This seems to help a bit.
- Some general cleanup around properties on component managers

This was done while doing some work towards passive coils

## Checklist

### DIFF
--- a/bluemira/equilibria/optimisation/harmonics/harmonics_approx_functions.py
+++ b/bluemira/equilibria/optimisation/harmonics/harmonics_approx_functions.py
@@ -109,8 +109,7 @@ def coil_harmonic_amplitude_matrix(
         * lpmv(ones, degrees, np.cos(theta_f)[None, :])
         / np.sqrt(degrees * (degrees + 1))
     )
-
-    return sig_fig_round(currents2harmonics, 8)
+    return sig_fig_round(currents2harmonics, 15)
 
 
 def harmonic_amplitude_marix(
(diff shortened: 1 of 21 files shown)


bluemira/equilibria/optimisation/harmonics/harmonics_approx_functions.py
```diff
@@ -137,8 +109,7 @@ def coil_harmonic_amplitude_matrix(
         * lpmv(ones, degrees, np.cos(theta_f)[None, :])
         / np.sqrt(degrees * (degrees + 1))
     )
-
-    return sig_fig_round(currents2harmonics, 8)
+    return sig_fig_round(currents2harmonics, 15)
```
oliverfunk, line 112: I think these 15's should be passed in as params and be defaulted to 15

oliverfunk (commented):
I've gone through it quite carefully.

Two things:
- All the spherical harmioncs stuff is using sig_fig_round, where the constraints use np.round. They obviously do different things, but I find the inconsistency slightly worrying.

- The 15 sig figs used in the SH stuff should be a param.

bluemira/equilibria/optimisation/harmonics/harmonics_approx_functions.py
geograham: Hey, I have no problem with making it a default rather than hard coding it in. I have done that and it is in the latest commit.

geograham (approved):
Been through the changes, all looks sensible and ready to merge.

oliverfunk (approved):
Okay lgtm! 

Perpahs in some doc. it could be stated that rounding constraint return values is generally recommended for improving numeric stability.

As a recommendation for constraint implementers.

je-cook:
As we talked about I think its very constraint specific for those constraints that are numerically unstable. As you said we could do something as part of a wider refactor of the eq opt stuff

Issue #3557, opened 2024-08-28
advective_transport: incorrect heat flux corrective factor


DarioV86:
In `advective_transport.py` in `_analyse_SN` and `_analyse_DN` there is a corrective factor (`f_correct_power`) that should re-enter the tail of the exponential decay used to describe the power exhaustion within the scrape-off layer and ensure the energy conservation.
As such, this corrective factor would be expected to be an amount slightly greater than one. Instead this seems to be approximately 0.5 by effectively halving the first wall heat flux.

As the missing power to be re-entered comes from the flux that at the mid-plane falls outside the first wall, therefore unable to "carry" power by parallel transport, a way to correct the above mentioned factor and upgrade the broader heat flux calculation subject, is to perpendicularly distribute the missing power from the outermost flux line.

PR #3558, opened 2024-08-28
Fixing heat flux corrective factor

DarioV86:
#3557


je-cook (commented):
Thanks, good looks fine, just need to fix the test and add the explanatory note we talked about to the analyse function

je-cook (approved):

PR #3606, opened 2024-09-13
RZIp vertical stability criterion

je-cook:
## Description

Adds the RZIp to the code along side some greens functions and inductance calculations

## Checklist

### DIFF
--- a/bluemira/equilibria/coils/_grouping.py
+++ b/bluemira/equilibria/coils/_grouping.py
@@ -699,6 +699,11 @@ class CoilGroup(CoilGroupFieldsMixin):
         return self.__getter("b_max")
 
     @property
+    def resistance(self) -> np.ndarray:
+        """Get coil resistance"""
+        return self.__getter("resistance")
+
+    @property
     def discretisation(self) -> np.ndarray:
         """Get coil discretisations"""
         return self.__getter("discretisation")
@@ -827,6 +832,11 @@ class CoilGroup(CoilGroupFieldsMixin):
     @b_max.setter
     def b_max(self, values: float | Iterable[float]):
         """Set coil max fields"""
+        self.__setter("b_max", values)
+
+    @resistance.setter
+    def resistance(self, values: float | Iterable[float]):
+        """Set coil resistance"""
         self.__setter("b_max", values)
 
     @discretisation.setter
--- a/bluemira/equilibria/coils/_grouping.py
+++ b/bluemira/equilibria/coils/_grouping.py
@@ -114,7 +114,7 @@ class CoilGroup(CoilGroupFieldsMixin):
         Bx_analytic: bool = True,
         Bz_analytic: bool = True,
     ):
-        if any(not isinstance(c, Coil | CoilGroup) for c in coils):
+        if any(not isinstance(c, Coil | CoilGroup) for c in coils) or not coils:
             raise TypeError("Not all arguments are a Coil or CoilGroup.")
         self._coils = coils
         self._pad_discretisation(self.__list_getter("_quad_x"))
@@ -334,7 +334,7 @@ class CoilGroup(CoilGroupFieldsMixin):
                 pfcoils.append(coil)
             coils = pfcoils
             bluemira_warn(
-                "EQDSK coilset empty - dummy coilset in use."
+                "EQDSK coilset empty - dummy coilset in use. "
                 "Please replace with an appropriate coilset."
             )
             return cls(*coils)
@@ -548,14 +548,14 @@ class CoilGroup(CoilGroupFieldsMixin):
 
         raise KeyError(f"Coil '{name}' not found in Group")
 
-    def _get_coiltype(self, ctype: CoilType | str) -> list[Coil]:
+    def _get_coiltype(self, *ctype: CoilType | str) -> list[Coil]:
         """Find coil by type"""
         coils = []
-        ctype = CoilType(ctype)
+        ctype = tuple(CoilType(ct) for ct in ctype)
         for c in self._coils:
             if isinstance(c, CoilGroup):
-                coils.extend(c._get_coiltype(ctype))
-            elif c.ctype == ctype:
+                coils.extend(c._get_coiltype(*ctype))
+            elif c.ctype in ctype:
                 coils.append(c)
         return coils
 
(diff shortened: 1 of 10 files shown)


bluemira/equilibria/plotting.py
```diff
@@ -110,6 +110,7 @@
             "CS": "#003688",
             "Plasma": "r",
             "NONE": "grey",
+            "DUM": "green",
```
CoronelBuendia, line 90: "Passive"? Or what is this for?

bluemira/equilibria/vertical_stability.py
```diff
@@ -0,0 +1,221 @@
...
+
+    stabilising = np.einsum(
+        "d, db, bc, ac, a",
+        i_plasma,
+        msp_prime,
+        np.linalg.inv(mss),
+        msp_prime,
+        i_plasma,
+        optimize=["einsum_path", (0, 1), (1, 2), (0, 1), (0, 1)],
+    )
```
CoronelBuendia, line 201: I can confirm that this implementation using einsum and the explicit matrix inversion is correct, and is practically an order of magnitude faster than using `np.linalg.solve` for reasons that befuddle me.

bluemira/magnetostatics/greens.py
```diff
@@ -280,6 +433,71 @@ def greens_dpsi_dx(
     return MU_0_2PI * x * ((xc**2 - (z - zc) ** 2 - x**2) * i2 + i1)
 
 
+@nb.jit(nopython=True)
+def greens_dbz_dx(
```
CoronelBuendia, line 437: I have performed a like-for-like comparison with data from FIESTA for this green's function and the results are identical.

image

I have also compared this to the results using finite differences of the dBz/dx and dBx/dz and can confirm that they are identical and pretty damn close to this analytical form.

Note that using FD on the semianalytic forms of Bx for these derivatives in a rectangular XS coil of uniform current density, dBx/dz does look quite different around a coil:

image

My only comment (not to be treated in this PR) is that we now use different mathematical notation for our Green's functions in terms of the code, the equations, and the docstrings. This is upsetting on an aesthetic level, and perhaps I will clean it up in future.

bluemira/equilibria/vertical_stability.py
```diff
@@ -0,0 +1,218 @@
...
+        return stab_destab(
+            cc_current=cc.current,
+            ind_mat=self.ind_mat,
+            control_ind=self.coilset._control_ind,
+            uncontrolled_ind=list(
+                set(self.coilset._get_type_index()) - set(self.coilset._control_ind)
+            ),
+            r_struct=self.coilset.x,
+            x_ac=cc.x,
+            i_plasma=eq._jtor * eq.grid.step,
```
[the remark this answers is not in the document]
CoronelBuendia: TBH I think this is fine, it takes about 0.5 seconds on my machine with a 129 x 257 grid and no coil discretisation. I'd imagine it would slow down a fair bit with sub-coil filaments, but if that's the case we can treat this down the line. In terms of accuracy this is best anyway.

bluemira/equilibria/plotting.py
je-cook: this was added because there was no colour ever added when dummy coils were added when https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/2617 was merged so it crashed when trying to plot dummy coils IIRC

je-cook: passive coils are NONE but maybe they get renamed?

CoronelBuendia: OK, perhaps we take a look at this another time

bluemira/equilibria/coils/_tools.py
```diff
@@ -55,29 +61,61 @@ def make_mutual_inductance_matrix(coilset: CoilSet) -> np.ndarray:
...
+                )
+        else:
+            n_turns = coilset.n_turns
+    else:
+        n_coils = coilset.n_coils()
+        xcoord = coilset.x
+        zcoord = coilset.z
+        dx = coilset.dx
+        dz = coilset.dz
+        n_turns = coilset.n_turns
```
kj5248: @CoronelBuendia I have made changes to the with_quadratures section in a way that I hope is more robust than before. This fixes (alongside a try and except statement in vertical_stability.py) your commented issue where coilset discretisation causes it to fail.

I feel like this should not be needed in this section of the code and is only needed because coilset.n_turns doesn't function correctly. However I tried and failed to figure out what was causing it to not return the proper sized arrays and values in CoilSet so I fixed it here.

bluemira/magnetostatics/greens.py
CoronelBuendia: Spun off here #4259

bluemira/equilibria/coils/_tools.py
CoronelBuendia: Hi @kj5248, I think that this (as is) will be problematic. This partly stems from the fact that we generally rarely use `n_turns`.

To calculate the mutual inductance between two coils with, let's say 3 x 3 filaments each, we need to calculate the mutual inductance between each of the filaments in each coil w.r.t each of the filaments in the other coil. The problem is that this technically means the sum of 81 contributions, or an equivalent mutual inductance of two coils with 9 turns each (just positioned slightly differently).

If we want to use `n_turns` correctly, and separately from the filamentary discretisation of the coils, we should normalise mutual inductance between the two coils with n_filaments[i,j] and n_turns_[i,j] as M_ij = n_turns_i * n_turns_j * M_ij/(n_filaments_i * n_filaments_j).

This is a bit of the cuff and not very thought out  do you follow my thinking?

CoronelBuendia: Note that in the case that each turn in a coil is discretised as a filament, this approach cancels out and it falls back to just the mutual inductance with no normalisation.

bluemira/equilibria/coils/_grouping.py
```diff
@@ -555,16 +555,21 @@ def _find_coil(self, name):
...
+    def _get_coiltype(self, *ctype: CoilType | str) -> list[Coil]:
         """Find coil by type"""  # noqa: DOC201
         coils = []
-        ctype = CoilType(ctype)
+        if not isinstance(ctype, list):
+            ctype = list(ctype)
+        if isinstance(ctype[0], list):
+            ctype = [CoilType(c) for ct in ctype for c in ct]
+        else:
+            ctype = [CoilType(ct) for ct in ctype]
```
je-cook: I'd like to try and get to the bottom of this change, `*ctype` will be a tuple of `CoilType|str` so unless I'm mistaken the fix is to do just the if/else not the first if although I'm not 100% on that

kj5248: I made this change as I was struggling to get it to work with the *ctype change. It was initially this: `ctype = tuple(CoilType(ct) for ct in ctype)` but I found that resulted in ctype getting put into more and more layers of tuples as the function looped back into itself so I switched to lists as it worked better in terms of accessing and extracting it to avoid unnecessary nesting.

If you have some suggestions for how to achieve this with tuples I can give it a go and see if it causes the tests to fail or not.

je-cook: I think its just this a few well placed `*`:
```
        coils = []

        ctype = tuple(CoilType(ct) for ct in ctype)

        for c in self._coils:
            if isinstance(c, Circuit) and len(c._get_coiltype(*ctype)) > 0:
                coils.append(c)
            elif isinstance(c, CoilGroup):
                coils.extend(c._get_coiltype(*ctype))
            elif c.ctype in ctype:
                coils.append(c)
        return coils
```

je-cook: when was it failing before and I can try and test

Issue #3399, opened 2024-06-28
Clean up logging


je-cook:
## Description of issue / requirement to address

Our logging system is a little bit janky unlder the hood. It could really do with some TLC see #3391 for an example

## Proposed solution

- Possibly log per module.
- ~Deal with log formatting so new lines are put in when necessary~ #3435 
- Possibly beautify with rich or similar

## Alternative solutions

Leave as is it does work but isnt the nicest under the hood.

PR #3614, opened 2024-09-18
🎨 Logging cleanup

je-cook:
## Description

Moves some logging mechanics to a lower level so it is collated, some work towards #3399 

## Checklist

### DIFF
--- a/bluemira/base/logs.py
+++ b/bluemira/base/logs.py
@@ -12,9 +12,11 @@ from __future__ import annotations
 import logging
 import sys
 from enum import Enum
+from textwrap import dedent, wrap
 from types import DynamicClassAttribute
 from typing import TYPE_CHECKING
 
+from bluemira.base.constants import ANSI_COLOR, EXIT_COLOR
 from bluemira.base.error import LogsError
 
 if TYPE_CHECKING:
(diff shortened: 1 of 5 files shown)


bluemira/base/look_and_feel.py
```diff
@@ -401,7 +314,7 @@ def bluemira_error_clean(string: str):
     string:
         The string to colour print
     """
-    _terminator_handler(LOGGER.error, _print_color(string, "red"))
+    _terminator_handler(LOGGER.error, string, fmt=False)
```
oliverfunk, line 317:
```suggestion
    _terminator_handler(LOGGER.error, string, clean=True, fmt=False)
```

oliverfunk: Not sure? But it seems it's needed

bluemira/base/logs.py
```diff
@@ -69,6 +71,167 @@ def format(self, record) -> str:
...
+    information.
+
+    Parameters
+    ----------
+    string:
+        The string of text to colour and box
+    width:
+        The width of the box, default = 73 (leave this alone for best results)
+    color:
+        The color to print the text in from `bluemira.base.constants.ANSI_COLOR`
```
oliverfunk, line 206: no fmt, flush doc

oliverfunk: This is quite a nit, but fmt is so general in this case and when reading this code, I got confused trying to understand what it did and why sometimes fmt=True, and sometimes fmt=False.

I think if this option was changed to bluemira_fmt or bm_fmt, it would be a lot clearer. It's printing in the Bluemira format, the boxed decorated format.

oliverfunk (commented):
Nice cleanup

je-cook:
I've rearranged some things here so the comment dont really fit could you have another look?

bluemira/base/logs.py
je-cook: I'm trying to move most of this to lower level, I agree generally and I dont think its there yet but a user should never touch this

PR #3616, opened 2024-09-18
Changes to get flux_surfaces and find_legs

geograham:
## Description

find_legs.py 
- added function calculate_connection_length that takes target coordinate and can use either the flux surface geometry or field line tracer methods
- default rtol value changed in LegFlux
- converted get_legs_length to get_legs_length_and_angle

flux_surfaces.py 
- corrections (e.g. <0 set to <= 0) 
- fs input option for calculate_connection_length_fs

test_find.py 
- update for changed default tolerance. 

## Checklist

### DIFF
--- a/bluemira/equilibria/find_legs.py
+++ b/bluemira/equilibria/find_legs.py
@@ -13,7 +13,6 @@ from enum import Enum, auto
 import numpy as np
 
 from bluemira.base.error import BluemiraError
-
 from bluemira.equilibria.equilibrium import Equilibrium, Grid
 from bluemira.equilibria.error import EquilibriaError
 from bluemira.equilibria.find import (
@@ -216,8 +215,8 @@ class LegFlux:
             leg_dict = get_single_null_legs(
                 self.separatrix,
                 self.delta,
-                self.x_points,
                 self.o_point,
+                x_points=self.x_points,
             )
             if self.dx_offsets is not None:
                 return self.get_leg_offsets(leg_dict)
@@ -285,13 +284,14 @@ def get_legs_length_and_angle(
     return length_dict, angle_dict
 
 
-def get_single_null_legs(separatrix, delta, x_points, o_point):
+def get_single_null_legs(separatrix, delta, o_point, x_points=None, imin=None):
     """Get the legs from a single null separatrix and return as a dictionary."""
     sorted_legs = get_leg_list(
         separatrix,
         delta,
         o_p=o_point,
         x_p=x_points,
+        imin=imin,
     )
     return add_pair_to_dict(sorted_legs, x_p=x_points, o_p=o_point)
 
@@ -510,10 +510,10 @@ class CalcMethod(Enum):
     FIELD_LINE_TRACER = auto()
     FLUX_SURFACE_GEOMETRY = auto()
 
+
 def calculate_connection_length(
     eq: Equilibrium,
-    x: float | None = None,
-    z: float | None = None,
+    div_target_start_point: Coordinates | None = None,
     first_wall: Coordinates | Grid | None = None,
     forward: bool = True,
     psi_n_tol: float = 1e-6,
@@ -526,11 +526,16 @@ def calculate_connection_length(
     Calculate the parallel connection length from a starting point to a flux-intercepting
     surface using either flux surface geometry or a field line tracer.
     If no starting point is selected then use the separatrix at the Outboard Midplane.
+
+    Raises
+    ------
+    BluemiraError:
+        If an invalid option calculation_method is selected.
     """
     calculation_method = CalcMethod[calculation_method.upper()]
 
-    # Use Separatrix and OM if x,z not chosen
-    if (x is None) or (z is None):
+    # Use Separatrix (in BM is first 'open' fs) flux if div target point not chosen
+    if div_target_start_point is None:
         legflux = LegFlux(
             eq=eq,
             psi_n_tol=psi_n_tol,
@@ -545,9 +550,21 @@ def calculate_connection_length(
         else:
             f_s = legflux.separatrix
 
-        z_abs = np.abs(f_s.z)
-        z = np.min(z_abs)
-        x = np.max(f_s.x[z_abs == np.min(z_abs)])
+    else:
+        xfs, zfs = find_flux_surface_through_point(
+            eq.x,
+            eq.z,
+            eq.psi(),
+            div_target_start_point.x,
+            div_target_start_point.z,
+            eq.psi(div_target_start_point.x, div_target_start_point.z),
+        )
+        f_s = Coordinates({"x": xfs, "z": zfs})
+
+    # OMP is taken to be start point regardless of input
+    z_abs = np.abs(f_s.z)
+    z = np.min(z_abs)
+    x = np.max(f_s.x[z_abs == np.min(z_abs)])
 
     if calculation_method == CalcMethod.FIELD_LINE_TRACER:
         return calculate_connection_length_flt(
--- a/bluemira/equilibria/find_legs.py
+++ b/bluemira/equilibria/find_legs.py
@@ -11,6 +11,7 @@ import operator
 from enum import Enum, auto
 
 import numpy as np
+import numpy.typing as npt
 
 from bluemira.base.error import BluemiraError
 from bluemira.base.look_and_feel import bluemira_print
@@ -245,7 +246,7 @@ class LegFlux:
 
 def get_legs_length_and_angle(
     eq: Equilibrium,
-    leg_dict: dict,
+    leg_dict: dict[str, npt.NDArray[np.float64] | None],
     plasma_facing_boundary: Grid | Coordinates | None = None,
 ):
     """Calculates the length of all the divertor legs in a dictionary."""
--- a/bluemira/equilibria/find_legs.py
+++ b/bluemira/equilibria/find_legs.py
@@ -253,30 +253,26 @@ def get_legs_length_and_angle(
     length_dict = {}
     angle_dict = {}
     for name, leg_list in leg_dict.items():
-        if not leg_list:
-            lengths = [0.0]
-            angles = [np.pi]
-        else:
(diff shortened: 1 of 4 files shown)


bluemira/equilibria/flux_surfaces.py
```diff
@@ -486,7 +486,9 @@ def clip(self, first_wall: Coordinates):
 
         # Relying on the fact that first wall is ccw, get the intersection angle
         self.alpha = get_angle_between_points(
-            self.coords.points[-2], self.coords.points[-1], first_wall.points[fw_arg]
+            self.coords.points[-2],
+            self.coords.points[-1],
+            first_wall.points[fw_arg],
```
je-cook, line 491: this change is not needed

bluemira/equilibria/flux_surfaces.py
```diff
@@ -850,12 +855,18 @@ def calculate_connection_length_fs(
     passing through Coils, but really they should be intercepted beforehand!
     """
     if first_wall is None:
-        x1, x2 = eq.grid.x_min, eq.grid.x_max
-        z1, z2 = eq.grid.z_min, eq.grid.z_max
+        x1, x2 = eq.grid.x_min + 0.01, eq.grid.x_max - 0.01
+        z1, z2 = eq.grid.z_min + 0.01, eq.grid.z_max - 0.01
```
je-cook, line 859: Why add 0.01?

bluemira/equilibria/flux_surfaces.py
```diff
@@ -408,7 +408,7 @@ def reset_direction(coords):
 
         # Pick the first intersection, travelling from the o_point outwards
         deltas = x_inter - o_point.x
-        arg_inter = np.argmax(deltas > 0)
+        arg_inter = np.argmax(deltas >= 0)
```
je-cook, line 411: Whats the reason for including =0?

bluemira/equilibria/find_legs.py
```diff
@@ -468,3 +499,91 @@ def _extract_leg_using_index_value(
...
+    """
+    Class for use with calculate_connection_length function.
+    User can choose how the connection length is calculated
+    """
+
+    FIELD_LINE_TRACER = auto()
+    FLUX_SURFACE_GEOMETRY = auto()
+
+
+def calculate_connection_length(
```
je-cook, line 514: Can we have a test of this?

bluemira/equilibria/flux_surfaces.py
```diff
@@ -850,12 +855,18 @@ def calculate_connection_length_fs(
...
-    xfs, zfs = find_flux_surface_through_point(eq.x, eq.z, eq.psi(), x, z, eq.psi(x, z))
-    f_s = OpenFluxSurface(Coordinates({"x": xfs, "z": zfs}))
+    if f_s is None:
+        xfs, zfs = find_flux_surface_through_point(
+            eq.x, eq.z, eq.psi(), x, z, eq.psi(x, z)
+        )
+        f_s = Coordinates({"x": xfs, "z": zfs})
+    if f_s.closed:
+        return 0.0
+    f_s = OpenFluxSurface(f_s)
```
je-cook, line 869: Can we have a test for this open flux surface section

bluemira/equilibria/find_legs.py
```diff
@@ -234,34 +246,52 @@ def get_legs(
         return leg_dict
 
 
-def get_legs_length(
+def get_legs_length_and_angle(
```
je-cook, line 249: there are a lot of branches in this function that are untested

bluemira/equilibria/find_legs.py
```diff
@@ -66,7 +77,7 @@ def __init__(
         eq: Equilibrium,
         psi_n_tol: float = 1e-6,
         delta_start: float = 0.01,
-        rtol: float = 1e-1,
+        rtol: float = 1e-3,
```
je-cook, line 80: Can you explain this change?

bluemira/equilibria/find_legs.py
```diff
@@ -31,6 +34,8 @@
 class NumNull(Enum):
     """
     Class for use with LegFlux.
+    Double Null (DN)
```
je-cook, line 37: Put these two lines below each entry then they will be directly documented

je-cook (commented):
I think this needs some more tests before we move forward. Also please fix the linter issues

bluemira/equilibria/find_legs.py
geograham: Setting it to a lower tolerance was part of some testing that was left in, I am returning it to a more strict default.

bluemira/equilibria/flux_surfaces.py
geograham: This needs some name updating, as the point being used is not always the magnetic centre of the plasma, in which case you might need the '='.  Updated the name and comments.

bluemira/equilibria/flux_surfaces.py
geograham: If we don't come slightly in from the grid edge then the flux line does not intersect.

bluemira/equilibria/find_legs.py
```diff
@@ -34,16 +37,22 @@ class NumNull(Enum):
     """
 
     DN = auto()
+    # Double Null
```
je-cook, line 40: Same for the other 3, I meant this:
```suggestion
    """Double Null"""
```

bluemira/equilibria/find_legs.py
```diff
@@ -34,16 +37,22 @@ class NumNull(Enum):
...
     SN = auto()
+    # Single Null
 
 
 class SortSplit(Enum):
     """
     Class for use with LegFlux.
+     (X)
+
```
je-cook, line 49:
```suggestion
```

bluemira/equilibria/flux_surfaces.py
je-cook: Did this work before? Also I feel this should be configurable if we are to do this (probably a kwarg as its not something that will regularly be configured I guess)

bluemira/equilibria/flux_surfaces.py
```diff
@@ -850,12 +856,18 @@ def calculate_connection_length_fs(
...
     if first_wall is None:
-        x1, x2 = eq.grid.x_min, eq.grid.x_max
-        z1, z2 = eq.grid.z_min, eq.grid.z_max
+        x1, x2 = eq.grid.x_min + 0.01, eq.grid.x_max - 0.01
+        z1, z2 = eq.grid.z_min + 0.01, eq.grid.z_max - 0.01
         first_wall = Coordinates({"x": [x1, x2, x2, x1, x1], "z": [z1, z1, z2, z2, z1]})
 
-    xfs, zfs = find_flux_surface_through_point(eq.x, eq.z, eq.psi(), x, z, eq.psi(x, z))
-    f_s = OpenFluxSurface(Coordinates({"x": xfs, "z": zfs}))
+    if f_s is None:
```
je-cook, line 863: Could we test for fs == None, fs.closed would be nice too but thats much simpler

bluemira/equilibria/find_legs.py
```diff
@@ -468,3 +501,91 @@ def _extract_leg_using_index_value(
...
+    Raises
+    ------
+    BluemiraError
+        If an invalid option calculation_method is selected.
+    """
+    calculation_method = CalcMethod[calculation_method.upper()]
+
+    # Use Separatrix (in BM is first 'open' fs) flux if div target point not chosen
+    if div_target_start_point is None:
```
je-cook, line 540: Could we test this branch too.

bluemira/equilibria/find_legs.py
```diff
@@ -234,34 +248,52 @@ def get_legs(
         return leg_dict
 
 
-def get_legs_length(
+def get_legs_length_and_angle(
     eq: Equilibrium,
     leg_dict: dict,
```
je-cook, line 253: I'd like slightly better typing here if possible. I think its this but please check
```suggestion
    leg_dict: dict[str, float | None],
```

bluemira/equilibria/find_legs.py
```diff
@@ -234,34 +248,52 @@ def get_legs(
...
 ):
     """Calculates the length of all the divertor legs in a dictionary."""
     length_dict = {}
-    for n, leg_list in leg_dict.items():
-        if not isinstance(leg_list[0], Coordinates):
+    angle_dict = {}
+    for name, leg_list in leg_dict.items():
+        if not leg_list:
             lengths = [0.0]
+            angles = [np.pi]
```
je-cook, line 262: This is not hit in the tests because you've set the leg length to [None] and not None. Is that intentional?

bluemira/equilibria/find_legs.py
geograham: There is no 'of course!' emoji. Otherwise it would go here.

bluemira/equilibria/flux_surfaces.py
geograham: We can just remove it. It would not be a sensible way to use the functions, as we will not want the angle of intersection with the grid edge.

bluemira/equilibria/flux_surfaces.py
geograham: Have added tests for both.

bluemira/equilibria/find_legs.py
geograham: Good spot, this is out of date, have modified

bluemira/equilibria/flux_surfaces.py
```diff
@@ -581,6 +584,8 @@ def plot(self, ax: plt.Axes | None = None, **kwargs):
         """
         Plot the FieldLine.
 
+        #TODO this needs updating
+
```
je-cook, line 588:
```suggestion
```

je-cook (approved):
Once the plotting issue is opened I will merge

PR #3624, opened 2024-10-01
Core radiation fix

DarioV86:
The core radiation was not calculated correctly because when accessing the tables from ADAS stored in PROCESS, Lz and Zeff where accessed with a reference temperature in keV instead of eV.

Besides the radiation calculation fix, a few checks related to the effective charge profiles and the average Z_eff value across the plasma were added.

Also added a scaling factor to use non-uniform electron temperature and electron density along the closed flux lines in the plasma core.

### DIFF
--- a/bluemira/radiation_transport/radiation_profile.py
+++ b/bluemira/radiation_transport/radiation_profile.py
@@ -29,6 +29,7 @@ from bluemira.radiation_transport.flux_surfaces_maker import (
 )
 from bluemira.radiation_transport.radiation_tools import (
     calculate_line_radiation_loss,
+    calculate_total_radiated_power,
     electron_density_and_temperature_sol_decay,
     exponential_decay,
     gaussian_decay,
@@ -59,7 +60,9 @@ if TYPE_CHECKING:
 class RadiationSourceParams(ParameterFrame):
     """Radiaition source parameter frame"""
 
-    sep_corrector: Parameter[float]
+    sep_corrector_omp: Parameter[float]
+    """Separation correction for double and single null plasma"""
+    sep_corrector_imp: Parameter[float]
     """Separation correction for double and single null plasma"""
     alpha_n: Parameter[float]
     """Density profile factor"""
@@ -192,7 +195,7 @@ class Radiation:
 
         Returns
         -------
-        te:
+        te [keV]:
             poloidal distribution of electron temperature
 
         Raises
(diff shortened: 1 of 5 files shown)


je-cook (commented):
This is going to take me a bit of time to go through. Here are a few things to start

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -410,64 +422,141 @@ def plot_mp_radiation_profile(self):
...
-            for ft, n in zip(self.flux_tubes, self.profiles.ne, strict=False)
-        ]
+        if scaling:
+            # Calculate scaling factors for temperature and density along each flux tube
+            self.scaling_pol_t = [
+                self.scaling_factor_core(ft, alpha=0.5) for ft in self.flux_tubes
+            ]
+            self.scaling_pol_n = [
+                self.scaling_factor_core(ft, alpha=0.5) for ft in self.flux_tubes
+            ]
```
je-cook, line 446: these lines are the same unless I'm missing something? Could we just copy?

bluemira/radiation_transport/radiation_profile.py
```diff
@@ -1404,7 +1494,7 @@ def calculate_sol_distribution(self, firstwall_geom: Grid):
                 flux_tubes=getattr(self, f"flux_tubes_{side}_{low_up}"),
                 x_strike=getattr(self, f"x_strike_{side}"),
                 z_strike=getattr(self, f"z_strike_{side}"),
-                main_ext=None,
+                main_ext=3,
```
je-cook, line 1497: whats this mean?

bluemira/radiation_transport/radiation_tools.py
```diff
@@ -365,31 +373,26 @@ def electron_density_and_temperature_sol_decay(
...
     """
     # temperature and density decay factors
-    t_factor = 7 / 2
-    n_factor = 1
+    if f_exp == 1:
+        t_factor = 7 / 2
+        n_factor = 1 / 3
+    else:
+        t_factor = 7
+        n_factor = 1 / 7
```
je-cook, line 381: where do these come from? Could we get a reference or some comment?

je-cook (changes requested):
Ok, I have no knowledge of where most of this maths come from so could we have a few references or documents for that?

The code itself looks fine although the tests need fixing

bluemira/radiation_transport/radiation_profile.py
DarioV86: @je-cook I wrote two lines because alpha may be different between temperature and density. Should I set it in a different way?

bluemira/radiation_transport/radiation_profile.py
DarioV86: @je-cook I parametrised `main_ext` as it should be a user choice. Does it work better?

bluemira/radiation_transport/radiation_tools.py
DarioV86: @je-cook also parametrised these factors. Going to add references.

bluemira/radiation_transport/radiation_profile.py
je-cook: but here alpha is hardcoded to it will never be different? 
The lines themselves are identical apart from the variable name

DarioV86: I am thinking of removing these scaling factors. They make very minor difference in the overall calculation, and hardcoded are even more useless.

je-cook:
Ok there are tests for calculate_z_species that need changing/updating because you've removed the function here

bluemira/radiation_transport/radiation_tools.py
```diff
@@ -542,30 +557,155 @@ def ion_front_distance(
     return abs(z_strike - x_pt_z) - abs(z_ext)
 
 
-def calculate_z_species(
-    t_ref: np.ndarray, z_ref: np.ndarray, species_frac: float, te: np.ndarray
-) -> np.ndarray:
+def calculate_zeff(
```
je-cook, line 560: Could we have a test for this function?

je-cook (approved):

PR #3680, opened 2024-10-28
Adding materials to DEMO

oliverfunk:
This PR setups the config and mechanics for specifying materials for each DEMO PhysicalCompoents in XYZ.

Part of the challenge in doing this is:
- Getting the necessary and correct material definitions data (in `materials.json`, `mixtures.json`)
- Specifying which components get what material

For some components, the geometry may need to be altered to define different parts of the component.
The breeding blanket for example does not have regions defined for the breeding zone, manifold etc.

## Linked Issues

Closes #{ID}

## Checklist

### DIFF
--- a/bluemira/builders/coil_supports.py
+++ b/bluemira/builders/coil_supports.py
@@ -307,7 +307,9 @@ class ITERGravitySupportBuilder(Builder):
         component = PhysicalComponent(
             "ITER-like gravity support",
             shape,
-            material=get_cached_material(self.build_config.get("material").get("GS")),
+            material=get_cached_material(
+                self.build_config.get("material", {}).get("GS")
+            ),
         )
         apply_component_display_options(component, color=BLUE_PALETTE["TF"][2])
         return component
@@ -376,7 +378,7 @@ class PFCoilSupportBuilder(Builder):
             self.name,
             face,
             material=get_cached_material(
-                self.build_config.get("material").get("PF ICS")
+                self.build_config.get("material", {}).get("PF ICS")
             ),
         )
         apply_component_display_options(component, color=BLUE_PALETTE["TF"][2])
@@ -923,7 +925,7 @@ class OISBuilder(Builder):
                 f"{self.OIS_XZ} {i}",
                 face,
                 material=get_cached_material(
-                    self.build_config.get("material").get(self.OIS_XZ)
+                    self.build_config.get("material", {}).get(self.OIS_XZ)
                 ),
             )
             apply_component_display_options(component, color=BLUE_PALETTE["TF"][2])
--- a/bluemira/builders/coil_supports.py
+++ b/bluemira/builders/coil_supports.py
@@ -42,6 +42,7 @@ from bluemira.geometry.tools import (
     sweep_shape,
 )
 from bluemira.geometry.wire import BluemiraWire
+from bluemira.materials.cache import get_cached_material
 from bluemira.optimisation import OptimisationProblem
 from bluemira.utilities.tools import floatify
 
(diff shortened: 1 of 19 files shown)


bluemira/builders/coil_supports.py
```diff
@@ -283,7 +284,13 @@ def build_xyz(
         # Finally, make the floor block
         shape_list.append(self._make_floor_block(floatify(v1.x), floatify(v4.x)))
         shape = boolean_fuse(shape_list)
-        component = PhysicalComponent("ITER-like gravity support", shape)
+        component = PhysicalComponent(
+            "ITER-like gravity support",
+            shape,
+            material=get_cached_material(
+                self.build_config.get("material", {}).get("GS")
```
je-cook, line 291: Thoughts welcome:
Possibly build_config['material'] should be autocreated in the reactor config and then the materials entry could be a default dict https://docs.python.org/3/library/collections.html#collections.defaultdict. It might mean we can chain gets. Also could well have unintended side effects...probably a bad idea.

je-cook, line 291: The component material maybe should go as a class var so it can be accessed outside? Same as we do in some places for the PhysicalComponent name (and you've done below).

bluemira/builders/pf_coil.py
```diff
@@ -120,16 +121,34 @@ def build_xz(self, shape: BluemiraWire) -> list[PhysicalComponent]:
...
+        )
         apply_component_display_options(ins, color=BLUE_PALETTE["PF"][3])
 
         cas_shape = offset_wire(ins_shape, self.params.tk_casing.value)
-        casing = PhysicalComponent(self.CASING, BluemiraFace([cas_shape, ins_shape]))
+        casing = PhysicalComponent(
+            self.CASING,
+            BluemiraFace([cas_shape, ins_shape]),
+            material=get_cached_material(
+                self.build_config.get("material", {}).get(self.CASING)
```
je-cook, line 149: maybe also we could have a wrapper function that does some of this so the user only passes in a dictionary and a key, possibly harder to work out whats going on:
```python
get_cached_material_for_component(self.build_config, self.CASING)
```

bluemira/materials/cache.py
je-cook, line 1: I really like these changes, nice!

eudemo/config/build_config.json
```diff
@@ -1,7 +1,7 @@
 {
   "params": "$path:params.json",
   "Radial build": {
-    "run_mode": "run",
+    "run_mode": "read",
```
je-cook, line 4: Note to change these back before merge

eudemo/config/build_config.json
```diff
@@ -60,7 +60,7 @@
     }
   },
   "Free boundary equilibrium": {
-    "run_mode": "run",
+    "run_mode": "read",
```
je-cook, line 63: And here, change back before merge, I'm not going to comment them all

eudemo/eudemo/reactor.py
```diff
@@ -410,6 +415,11 @@ def build_radiation_plugs(params, build_config, cr_ports, radiation_xz_boundary)
 
 if __name__ == "__main__":
     set_log_level("INFO")
+
+    establish_material_cache([
+        Path(CONFIG_DIR, "materials.json"),
```
je-cook, line 420: Could these paths come directly from the reactor config?

je-cook (commented):
On the whole its a really nice start, I like how you load and create the material cache.
There are probably some rough edges we can smooth and everything I mentioned is minor. I'll have a go with it and come back round again

bluemira/builders/pf_coil.py
oliverfunk: I think this is best idea

je-cook (approved):
Really nice improvement, thanks

PR #3682, opened 2024-10-30
Replaced plotting using Coordinate class in flux_surfaces.py as this method no longer exists

### LINKED ISSUE #3641: Update FieldLine plot()
## Describe the bug

Plotting for 'FieldLine' (bluemira.equilibria.flux_surfaces.py) needs updating, 'Coordinates' class does not have plot method anymore.

kj5248:
Used plot_coordinates method from plotter.py instead.

Did require fixes to plot_coordinates as there were issues with _get_plan_dims as plot_coordinates was expecting 2 dims out but was getting 3. This was because for 3D equilibria get planar dimensions will return all 3 so it should not have been called. This only happened because after plotting as 3D coordinates successfully it then tried to plot as 2D. This was fixed by added in an else statement tagging on to the 3D section so this no longer occurs. Afterwards the plot worked for both 3D and 2D scenarios

## Linked Issues

Closes #3641 

## Checklist

### DIFF
--- a/bluemira/display/plotter.py
+++ b/bluemira/display/plotter.py
@@ -869,8 +869,6 @@ def plot_coordinates(
     alpha: float
         The transparency to plot the Coordinates fill with
     """
-    from bluemira.utilities.plot_tools import coordinates_to_path  # noqa: PLC0415
-
     ndim = _get_ndim(coords)
 
     fc = kwargs.get("facecolor", "royalblue")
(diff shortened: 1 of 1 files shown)


geograham (changes requested):
SonarCloud is complaining about the complexity of the nested if statements. Could you move your 2D plotting stuff from 'plot coordinates' into a separate function equivalent to the '_plot_3d' that is used?

je-cook (approved):
Not a massive fan of plt.gca() as it could end up adding things to a possibly odd choice of figure. But not going to hold this up as its on other functions here.

PR #3695, opened 2024-11-14
Coil Field and Force Constraints Grad Modification

### LINKED ISSUE #3035: Possible error in grad calculation for field constraint function
## Describe the bug

Matrix multiplication with single value causes error. Gradient calculation to be checked and tested. 

## Additional context

Found when addressing issue #2961.

geograham:
## Linked Issues

Closes #3035 - > field and force constraints modified and now tested using 'approx_derivative'.

## Checklist

### DIFF
--- a/bluemira/equilibria/optimisation/constraint_funcs.py
+++ b/bluemira/equilibria/optimisation/constraint_funcs.py
@@ -228,8 +228,8 @@ class FieldConstraintFunction(ConstraintFunction):
 
         Bx_a = self.ax_mat @ currents
         Bz_a = self.az_mat @ currents
-
         B = np.hypot(Bx_a + self.bxp_vec, Bz_a + self.bzp_vec)
+
         return np.round(B - self.B_max, self._round_dp)
 
     def df_constraint(self, vector: npt.NDArray[np.float64]) -> npt.NDArray[np.float64]:
@@ -240,9 +240,23 @@ class FieldConstraintFunction(ConstraintFunction):
         Bz_a = self.az_mat @ currents
         B = np.hypot(Bx_a + self.bxp_vec, Bz_a + self.bzp_vec)
 
-        Bx = Bx_a * (Bx_a * currents + self.bxp_vec)
-        Bz = Bz_a * (Bz_a * currents + self.bzp_vec)
-        return np.round((Bx + Bz) / (B * self.scale**2), self._round_dp)
+        if len(B) == 1:
+            return np.round(
+                (self.scale / B)
+                * (
+                    self.ax_mat * (Bx_a + self.bxp_vec)
+                    + self.az_mat * (Bz_a + self.bzp_vec)
+                ),
+                self._round_dp,
+            )
+        return np.round(
+            (self.scale / B)[:, np.newaxis]
+            * (
+                self.ax_mat * (Bx_a + self.bxp_vec)[:, np.newaxis]
+                + self.az_mat * (Bz_a + self.bzp_vec)[:, np.newaxis]
+            ),
+            self._round_dp,
+        )
 
 
 class CurrentMidplanceConstraint(ConstraintFunction):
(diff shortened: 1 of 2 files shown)


bluemira/equilibria/optimisation/constraint_funcs.py
```diff
@@ -389,11 +413,16 @@ def cs_fz(self, f_matx):
     def pf_z_constraint(self, f_matx, max_value):
         """Constraint Function Absolute vertical force constraint on PF coils."""
         scaled_max_value = max_value / self.scale
-        self.constraint[: self.n_PF] = f_matx[: self.n_PF, 1] ** 2 - scaled_max_value**2
+        self.constraint[: self.n_PF] = (
+            np.sqrt(f_matx[: self.n_PF, 1] ** 2) - scaled_max_value
```
je-cook, line 417: How does this change make a difference. Would the previous case have the same profile as far as the optimiser is concerned? Also instead of squaring and rooting wouldnt the absolute value have the same effect with less calculation.

```suggestion
            np.abs(f_matx[: self.n_PF, 1]) - scaled_max_value
```

I dont know what the right answer is here but as you can tell I'm not quite comfortable with this

je-cook (changes requested):
I'm probably not best placed to review this, maybe we can have a chat at some point. Either way if this change is correct the squaring and rooting should be replaced with `np.abs`

bluemira/equilibria/optimisation/constraint_funcs.py
```diff
@@ -240,9 +240,23 @@ def df_constraint(self, vector: npt.NDArray[np.float64]) -> npt.NDArray[np.float
...
+                self._round_dp,
+            )
+        return np.round(
+            (self.scale / B)[:, np.newaxis]
+            * (
+                self.ax_mat * (Bx_a + self.bxp_vec)[:, np.newaxis]
+                + self.az_mat * (Bz_a + self.bzp_vec)[:, np.newaxis]
+            ),
+            self._round_dp,
+        )
```
je-cook, line 259: Does this need to happen? I thought numpy would deal with the len(B)==1 implicitly?

bluemira/equilibria/optimisation/constraint_funcs.py
```diff
@@ -474,7 +491,8 @@ def __init__(
         *,
         round_dp: int = 16,
     ):
-        super().__init__(a_mat, b_vec, n_PF, n_CS, scale)
+        super().__init__(a_mat, b_vec, n_PF, n_CS)
+        self.scale = scale
```
je-cook, line 495: Can we revert this otherwise the scale needs to be maintained in multiple places

je-cook (approved):
Thanks very much for this, two minors but otherwise I'm happy

bluemira/equilibria/optimisation/constraint_funcs.py
geograham: We get a mismatch of array sizes without it, probably has not come up before as we would normally have more than one field constraint.

bluemira/equilibria/optimisation/constraint_funcs.py
geograham: reverted

bluemira/equilibria/optimisation/constraint_funcs.py
je-cook: Then at least we're not repeating code
```suggestion
        res = np.round(
            (self.scale / B)[:, np.newaxis]
            * (
                self.ax_mat * (Bx_a + self.bxp_vec)[:, np.newaxis]
                + self.az_mat * (Bz_a + self.bzp_vec)[:, np.newaxis]
            ),
            self._round_dp,
        )

        if len(B) == 1:
            return np.squeeze(res)
        return res
```

je-cook (approved):

PR #3636, opened 2024-10-11
👽️ PROCESS develop api changes

je-cook:
## Description

Updates our PROCESS api for the current develop version of PROCESS.

## Checklist

I confirm that I have completed the following checks:

- [ ] Tests run locally and pass `pytest tests --reactor`
- [ ] Code quality checks run locally and pass `pre-commit run --from-ref develop --to-ref HEAD`
- [ ] Documentation built locally and checked `sphinx-build -W documentation/source documentation/build`

PR #3728, opened 2024-12-03
Update _plotting.py to plot radial build from MFILE for process versi…

### LINKED ISSUE #722: Plotting the radial build from systems code run

It should be easy to plot the radial build of a reactor, resulting from a systems code run.

At present the PROCESS radial build plotting relies on some hacking of the OUT.DAT, because the radial build vector is not stored in variables.

## Proposed solution

We could reconstruct the radial build vector and makes some parameters for these.

Plotting the radial build is then a "view" of the reactors parameters, rather than plotting the solver per se.

## Alternative solutions

PROCESS could also output the radial build vector with some associated variable names we could map to.. maintaining a reconstruction of the radial build vector to mimic PROCESS is surely going to be difficult in the long run.

## Additional context

Probably needs to be actioned after #721

athoynilimanew:
…on >=3.1.0

## Linked Issues

Closes #722 

## Description

Plots the process radial build directly from MFILE.DAT, if created with process version 3.1.0 or higher

For older MFILE versions, simply use older plot routine to plot radial build.

did not delete the older routine as it would break other examples/use cases that just "read" existing older PROCESS OUT.DAT files.

## Checklist


athoynilimanew:
I will resume the outstanding issues in this papercut after I come back from vacation in January

je-cook:
Hi, thanks for this, a few things that I think will make this more robust to future changes:
- I think that _plotting.py should be removed and the plot function should be on the solver. You then have access to the mfile throught the teardown object and do not need a custom mfile reader
- We should target the process develop branch in #3636 as this has all the upcoming changes (including the radial vector). Once process releases their next version we can merge that branch in followed by this one

athoynilimanew:
I am closing it because I messed up with rebases :D I am remaking a different branch from scratch

PR #3735, opened 2024-12-13
put the Monte Carlo back in the fuel cycle model

CoronelBuendia:
## Linked Issues

Took me a while to debug this... I'm not sure of the motivations behind lumping fixed random seeds in an enum, but I can say for sure that it means the fuel cycle model can no longer be run Monte Carlo.

Closes #{ID}

## Checklist

### DIFF
--- a/bluemira/base/constants.py
+++ b/bluemira/base/constants.py
@@ -687,7 +687,3 @@ class RNGSeeds(Enum):
     """
 
     equilibria_harmonics = 2944412338698111642
-    timeline_tools_lognorm = 6613659347120864846
-    timeline_tools_truncnorm = 9523110846560405221
-    timeline_tools_expo = 15335509124046896388
-    timeline_outages = 5876826953682921855
--- a/bluemira/fuel_cycle/timeline.py
+++ b/bluemira/fuel_cycle/timeline.py
@@ -11,7 +11,7 @@ Partially randomised fusion reactor load signal object and tools
 import matplotlib.pyplot as plt
 import numpy as np
 
-from bluemira.base.constants import S_TO_YR, YR_TO_S, RNGSeeds
+from bluemira.base.constants import S_TO_YR, YR_TO_S
 from bluemira.fuel_cycle.timeline_tools import (
     LogNormalAvailabilityStrategy,
     OperationalAvailabilityStrategy,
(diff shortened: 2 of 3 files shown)


je-cook:
The reasoning was reproducability of results. You need to be able to set the random seed to have results that are reproduceable. I feel removing the seeds is possibly the wrong approach. 

The seed should just be a starting point for generation and if we need multiple starting points we should use a specific seed (that can be changed by the user) and we should jump down the generation sequence. We can also change the generation alogorithm from the default [PGC64](https://numpy.org/doc/stable/reference/random/generator.html) and use something like the [mersenne twister](https://numpy.org/doc/stable/reference/random/bit_generators/mt19937.html#numpy.random.MT19937). The jumped function should be enough to move us down the generation sequence far enough to accomplish monte carlo runs which is what I have done previously.

CoronelBuendia:
> The reasoning was reproducability of results. You need to be able to set the random seed to have results that are reproduceable. I feel removing the seeds is possibly the wrong approach.
> 
> The seed should just be a starting point for generation and if we need multiple starting points we should use a specific seed (that can be changed by the user) and we should jump down the generation sequence. We can also change the generation alogorithm from the default [PGC64](https://numpy.org/doc/stable/reference/random/generator.html) and use something like the [mersenne twister](https://numpy.org/doc/stable/reference/random/bit_generators/mt19937.html#numpy.random.MT19937). The `jumped` function on the MT19937 should be enough to move us down the generation sequence far enough to accomplish monte carlo runs which is what I have done previously.

The intended use of the fuel cycle model is as in the example, where the random seed is set at the outermost-level and results are reproducible for the same set of inputs. Right now the example does not work as intended: it just repeats the same results n times. The fact that these modifications just hard-code the random seed inside some utility functions, without taking an optional argument that could potentially override it, means that the fuel cycle model cannot be run Monte Carlo.

It's a bug in functionality that previously worked. This PR fixes it.

What you're talking about is a potential improvement and I have nothing against it, per se.

je-cook:
That may be true but while fixing it, it should be fixed properly

CoronelBuendia:
> That may be true but while fixing it, it should be fixed properly. In fact I would consider the inability to reproduce results in itself a bug

Was there such an inability in the user-facing API? I wasn't aware... The example used to produce reproduce-able results, at least per machine (I never checked across machines). If you want, when I get back to this branch I can run the example as is, and you run it on your machine are see what we get?

je-cook:
Please see #3742 for my implementation of this that hopefully fixes both bugs

PR #3741, opened 2024-12-17
Add titles to all test plots

je-cook:
## Description

Automatically adds a title to figures in tests so they can be tracked down if there is something wrong.
Only works if 'show' is not called within the test

## Checklist


OceanNuclear (approved):
Looks good, helpful for tracking down which plot we're looking at when using `--plotting-on`.

PR #3757, opened 2025-01-07
Ocean/periodic wire types fix

OceanNuclear:
## Linked Issues

Closes #2010 

## Description

This should fix the specific problem of "splitting an arc-of-circle at angle=0 erroneously yields a full-circle + the original arc".

## Interface Changes

N/A

## Checklist

I confirm that I have completed the following checks:

- [x] Tests run locally and pass `pytest tests --reactor`
- [ ] Code quality checks run locally and pass `pre-commit run --from-ref develop --to-ref HEAD`
- [ ] Documentation built locally and checked `sphinx-build -W documentation/source documentation/build`

PR #3758, opened 2025-01-08
Possibly a more proper way of fixing the floating point issue.

OceanNuclear:
(but it's quite verbose) - worth discussing whether this is necessary/ good.

## Linked PR

A more meticulous treatment of floating point numbers  than #3757 - but perhaps too pedantic?

## Description

Because the gap between floating point numbers scales up/down with the size of the float itself, using +/- EPS to create an interval might not yield the desired result (too big of an interval when the number is << 1.0, and zero-sized interval when the number is >>1.0).

## Interface Changes

N/A

## Checklist

### DIFF
--- a/bluemira/codes/_freecadapi.py
+++ b/bluemira/codes/_freecadapi.py
@@ -51,6 +51,7 @@ from bluemira.base.file import force_file_extension
 from bluemira.base.look_and_feel import bluemira_warn
 from bluemira.codes._freecadconfig import _freecad_save_config
 from bluemira.codes.error import FreeCADError, InvalidCADInputsError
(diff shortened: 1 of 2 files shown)


OceanNuclear:
As stated here, it's only going to matter when the the "parameter" argument >>1.0 or <<1.0. I don't know how often that comes up inside a single edge that forms the freecad wire (underlying bluemira wire), might need someone more experienced than me to say for certain.

OceanNuclear:
As discussed with @je-cook earlier in person, this implementation might be too strict, and may end up failing when multiple operations accumulate floating point errors (which I agree).
It's best to do it the hamfisted (but functional) way of +/-`EPS_FREECAD` for now.

PR #3776, opened 2025-01-20
First neutronics improvement

### LINKED ISSUE #3413: make_csg tests

OceanNuclear:
## Linked Issues

Linked PR: #3413 ; does not close the issue.

## Description

Small PR to make review more manageable. This is likely PR 1 of 3.

## Interface Changes

Refactored make_csg into 3 files: `csg_reactor`, `tools`, and `make_cell`. Many functions are relocated, but no change to their input/output signature for now.
`BluemiraNeutronicsCSG` is renamed to `OpenMCEnvironment` to more properly reflect its use (keep track of id of all surfaces and cells)

## Checklist


.gitignore
```diff
@@ -179,6 +179,7 @@ autoapi
 generated_data
 !generated_data/naughty_geometry/README.md
 tests/bluemira/test_generated_data
+tests/radiation_transport/neutronics/step_test/
```
je-cook, line 182: TODO remove when we're ready for this to go in

bluemira/codes/openmc/csg_reactor.py
```diff
@@ -0,0 +1,466 @@
...
+    from bluemira.radiation_transport.neutronics.neutronics_axisymmetric import (
+        NeutronicsReactor,
+    )
+
+
+@dataclass
+class CSGComponents:
+    """A container for each component represented in CSG geometry."""
+
+    blanket: BlanketCellArray
```
je-cook, line 47: Thoughts about naming here for generic situations. Main chamber might be a better name because here we are implicitly saying that if its not divertor its blanket (which is fine here in codes). If its user facing we should probably be a bit more explicit

bluemira/codes/openmc/csg_reactor.py
```diff
@@ -0,0 +1,466 @@
...
+        z_max, z_min, r_max, r_min = pre_cell_reactor.half_bounding_box
+
+        z_min_adj = z_min - D_TOLERANCE
+        z_max_adj = z_max + D_TOLERANCE
+        r_max_adj = r_max + D_TOLERANCE
+
+        rad_shield_wall_tk = pre_cell_reactor.tokamak_dimensions.rad_shield.wall
+
+        # make the universe box, incorporates the radiation shield wall
+        self = cls(CSGComponents(*[None for _ in CSGComponents.__dataclass_fields__]))
```
je-cook, line 411: This is not a nice way to do this, it also is not correct with your typing above because your setting everything to None.

A more obvious and clean way is to just make local variables for universe, blanket etc and put them in the dataclass as the last step. There's possibly some mechanics here that make that harder but that probably should be changed.
For instance you can have post init variables that are created after the initial components are provided

bluemira/codes/openmc/solver.py
```diff
@@ -49,7 +44,7 @@ class OpenMCRunModes(BaseRunMode):
     """OpenMC run modes"""
 
     RUN = openmc.settings.RunMode.FIXED_SOURCE.value
-    RUN_AND_PLOT = auto()
+    RUN_AND_PLOT = "run and plot"
```
je-cook, line 47: We could change this to possibly be a Flag... see https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/5d55ea28b1203960a2548b7f7866012f6bc5125c/bluemira/equilibria/diagnostics.py#L59. Would still need to inherit from BaseRunMode.

Dont change yet but something to look at.

je-cook (changes requested):
Good so far. Definitely some things still to work on

PR #3835, opened 2025-02-26
Used better optimal bounding box

### LINKED ISSUE #3818: The optimised bounding box method should use the FreeCAD call
Our BluemiraGeo.get_optimial_bounding_box is a bit of a hack here:

https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/53484e0fca724152b35c730b170336b27a549747/bluemira/geometry/base.py#L197

We should use the method available on the underlying freecad shape eg `BluemiraWire._shape.optimalBoundingBox()`

OceanNuclear:
## Linked Issues

Closes #3818 

## Description

Updated the "optimal_bounding_box" attribute to one that uses FreeCAD's built-in "optimalBoundingBox" (added in the updated version of freecad.), as a property of BluemiraGeo, rather than a method.

## Interface Changes

`get_optimal_bounding_box()` (a method) -> `optimal_bounding_box` (a property)
Removed the need for the tolerance parameter as well.

## Checklist

### DIFF
--- a/bluemira/codes/_freecadapi.py
+++ b/bluemira/codes/_freecadapi.py
@@ -803,6 +803,16 @@ def bounding_box(obj: apiShape) -> tuple[float, float, float, float, float, floa
     return box.XMin, box.YMin, box.ZMin, box.XMax, box.YMax, box.ZMax
 
 
+def optimal_bounding_box(
+    obj: apiShape,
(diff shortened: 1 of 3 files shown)


oliverfunk (approved):
nice!

PR #3863, opened 2025-04-04
Add labels for height top upper dz flat bottom lower for PolySpline

### LINKED ISSUE #3587: Add labels for l0s - l3s l03 -l3e for PolySpline
https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/8277729541d5421cade5fe0c663441121a801742/bluemira/geometry/parameterisations.py#L1586-L1587

athoynilimanew:
## Linked Issues

Closes #3587 

## Description

Adds labels for height, top, upper, dz, flat, bottom, lower for PolySpline.

Since flat, upper, lower, top, bottom are fractions, labels for the lengths are used where these fractions are multiplied with the height.

Example labels after plotting

image

To reproduce the above plot:

```
from bluemira.geometry.parameterisations import PolySpline
test_spline = PolySpline(
    {
        "flat": {
            "value": 0.3,
            "lower_bound" : 0,
            "upper_bound" : 1,
            "description": "Fraction of straight outboard leg",
        },
        "z2": {
            "value": 0.3,
            "lower_bound" : -2,
            "upper_bound" : 2,
            "description": "Outer note vertical shift",
        },
        "tilt": {
            "value": 25,
            "lower_bound" : -45,
            "upper_bound" : 45,
            "description": "Outboard angle [degrees]",
        }

    }
)

f, ax = plt.subplots()
test_spline.plot(ax=ax, labels=True)

points = test_spline.create_shape().vertexes.T
from bluemira.display.plotter import PointsPlotter
pp = PointsPlotter()
for point in points:
    pp.plot_2d(point, ax=ax, show=False)

ax.grid(visible=True)
plt.show()
```
## Checklist

### DIFF
--- a/bluemira/geometry/parameterisations.py
+++ b/bluemira/geometry/parameterisations.py
@@ -1559,82 +1559,79 @@ class PolySpline(GeometryParameterisation[PolySplineOptVariables]):
 
         x1, x2, z2, height = self.variables.values[:4]
         dz, flat, tilt = self.variables.values[6:9]
-        # Label for xs
-        offset_ar_x = 0
-        for v, name in zip([x1, x2], ["x1", "x2"], strict=False):
-            offset_ar_x = max(z for x, _, z in shape.vertexes.T if np.isclose(x, v)) / 2
-            self._annotator(
-                ax,
-                name,
-                (0, offset_ar_x),
-                (v, offset_ar_x),
-                (v * 0.4, offset_ar_x),
-            )
-            ax.plot([0, 0], [offset_ar_x - 0.1, offset_ar_x + 0.1], color="k")
-            ax.plot([v, v], [offset_ar_x - 0.1, offset_ar_x + 0.1], color="k")
-
-        # Label for z2, dz
-        offset_ar_z = 0
-        for v, name in zip([z2, dz], ["z2", "dz"], strict=False):
-            xcor = shape.center_of_mass[0] + offset_ar_z
-            self._annotator(
-                ax,
-                name,
-                (xcor, 0),
-                (xcor, v),
-                (xcor, v * 0.4),
-            )
-            ax.plot([xcor - 0.1, xcor + 0.1], [0, 0], color="k")
-            ax.plot([xcor - 0.1, xcor + 0.1], [v, v], color="k")
-            offset_ar_z += 2
-
-        # Label for height
-        self._annotator(
-            ax,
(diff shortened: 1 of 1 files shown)


bluemira/geometry/parameterisations.py
```diff
@@ -1552,10 +1554,173 @@ def _label_function(self, ax: plt.Axes, shape: BluemiraWire):
...
+                        if np.isclose(v, x)
+                    ]
+                )
+                else np.mean([
+                    z
+                    for _, z in sorted(
+                        zip(abs(shape.vertexes[0] - v), shape.vertexes[2], strict=False)
+                    )[:2]
+                ])
+            )  # So that the endpoint of the arrow lies on the curve
```
je-cook, line 1598: If I'm not mistaken this is nearly repeated below? Could we move this into a private method eg `def _get_annotated_offset_z(...)`. I think the only different is v vs x1+v which could be an input for the function

bluemira/geometry/parameterisations.py
```diff
@@ -1552,10 +1554,173 @@ def _label_function(self, ax: plt.Axes, shape: BluemiraWire):
...
+                (xcor, zcors[0]),
+                (xcor, zcors[1]),
+                (xcor, zcors[1] * 0.8),
+            )
+            ax.plot([xcor - 0.05, xcor + 0.05], [zcors[0], zcors[0]], color="k")
+            ax.plot([xcor - 0.05, xcor + 0.05], [zcors[1], zcors[1]], color="k")
+            annotate_offset_x += 0.5
+
+        # Label annotation for flat
+        if flat != 0:
```
je-cook, line 1706: Could we have this label in all cases. I think its useful for a prospective user to see where the label would be even if it is 0.

je-cook (changes requested):
Thanks very much for this, obviously harder than it looked! A few minor changes but nothing major

bluemira/geometry/parameterisations.py
athoynilimanew: Done in latest commit, (please resolve if it's ok!)

bluemira/geometry/parameterisations.py
athoynilimanew: Done in latest commit, (please resolve if it's ok!)

je-cook (approved):
Thanks!

PR #3877, opened 2025-04-14
Feed through diagnostics in equilibria run

je-cook:
## Linked Issues

Closes #{ID}

## Description

This feeds through some of the new diagnostic options into the COP problems. Does a little reorganising and cleanup as well

## Checklist

### DIFF
--- a/bluemira/equilibria/diagnostics.py
+++ b/bluemira/equilibria/diagnostics.py
@@ -14,10 +14,17 @@ from enum import Enum, Flag, auto
 from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
+import matplotlib.pyplot as plt
+
 from bluemira.base.file import try_get_bluemira_path
+from bluemira.equilibria.constants import DPI_GIF, PLT_PAUSE
+from bluemira.utilities.plot_tools import make_gif, save_figure
 
 if TYPE_CHECKING:
     from os import PathLike
+
+    from bluemira.equilibria.equilibrium import Equilibrium
+    from bluemira.equilibria.solve import ConvergenceCriterion
 
 
 class GridPlotType(Flag):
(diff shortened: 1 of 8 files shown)


bluemira/equilibria/diagnostics.py
```diff
@@ -286,3 +293,40 @@ def __post_init__(self):
                 "", subfolder="generated_data", allow_missing=not self.gif
             )
         self.figure_folder = figure_folder
+
+        if self.plot is PicardDiagnostic.NO_PLOT:
+            self.update_figure = self.make_gif = self.finalise_plots = self._noop
```
oliverfunk, line 298: Perhaps locally defining _noop would be better, ie in the post_init method

bluemira/equilibria/solve.py
```diff
@@ -563,11 +559,9 @@ def __next__(self):
         self._psi = self.eq.psi()
         self._j_tor = self.eq._jtor
         check = self.check_converged()
-        if self.diagnostic_plotting.plot is not PicardDiagnostic.NO_PLOT:
-            self.update_fig()
+        self.diagnostic_plotting.update_figure(self.eq, self.convergence, self.i)
         if check:
-            if self.diagnostic_plotting.gif:
-                make_gif(self.diagnostic_plotting.figure_folder, self.pname)
+            self.diagnostic_plotting.make_gif()
```
oliverfunk, line 564: Are you missing a check for self.diag_poltting.gif here? as it was prior

oliverfunk (commented):
As discussed in person, I understand why making the opt problems needs to happen in the setting classes but the names are not very clean/descriptive. Although this is a minor point.

Otherwise two small comments

bluemira/equilibria/solve.py
je-cook: no because it now checking withing the call if its enabled

PR #3900, opened 2025-04-30
WIP: Fixed boundary equilibrium fixes, tests, and plotting

CoronelBuendia:
## Linked Issues

Closes #{ID}

## Checklist


bluemira/equilibria/fem_fixed_boundary/file.py
```diff
@@ -145,9 +146,9 @@ def save_fixed_boundary_to_file(
         nz=nz,
         pressure=pressure,
         pprime=p_prime,
-        psi=psi,
+        psi=2.0 * np.pi * psi,
         psibdry=0,
-        psimag=psi_mag,
+        psimag=2.0 * np.pi * psi_mag,
```
je-cook, line 151: Currently for better or worse the COCOS of bluemira is 3 which means that the output is in V.s/rad. Feeds into #318 and our various internal problems with what standard we are using

je-cook: also I should add that IMAS IDS has moved from /is moving to COCOS 1 which is also V.s/rad

CoronelBuendia: OK, I was just preserving existing behaviour, but I have removed now.

bluemira/equilibria/fem_fixed_boundary/fem_magnetostatic_2D.py
```diff
@@ -506,32 +545,32 @@ def _plot_current_iteration(
...
 
-    @staticmethod
-    def _add_colorbar(cm, cax, title):
-        last_axes = plt.gca()
-        ax = cm.axes
-        fig = ax.figure
-        fig.colorbar(cm, cax=cax)
-        cax.set_title(title)
-        plt.sca(last_axes)
+def _plot_array(
```
ivanmaione, line 553: I agree with the change of removing _plot_array and _add_colorbar as @staticmethod (since it wasn’t meaningful in this context). However, it might be better to move these functions outside of the class (into utilities.py?). Could an option be to consider reusing plot_scalar_field?

CoronelBuendia: Hmm.. given the use of tricontour, and the fairly trivial nature, I think it is OK to have them as private functions in this module...

PR #3905, opened 2025-05-01
CAD Imprinting

### LINKED ISSUE #3902: Imprinting CAD

As a modeller interested in running a neutronics simulation on CAD, co-incident faces need to be imrpinted and then merged.

### Proposed solution

Implement an efficient algorithm that can detect (ideally taking to account the geometry in a confromal manner) overlapping or touching solids.

Then use a routine to imprint and merge said solids.

This should work such that the geometry saved into some format (say a .stp file) includes the merged faces.

It could still be fine if this is not that case as a pre-processing step can be run on the meshed geometry (i.e. after some MOAB meshing) that can merge close vertices/faces.

### Alternative solutions

_No response_

### Additional Context

_No response_

oliverfunk:
## Linked Issues

Closes #3902

## Checklist


oliverfunk:
There's an interesting tool I tried called Haystack to help review this PR, check it out here:

https://alb15.haystackserver.com/Fusion-Power-Plant-Framework/bluemira/3905/?folder=/home/coder/repo

bluemira/geometry/tools.py
```diff
@@ -1640,10 +1640,23 @@ def save_cad(
         shapes = list(shapes) if isinstance(shapes, Iterable) else [shapes]
     if names is not None and not isinstance(names, list):
         names = [names]
+    filename = Path(filename).as_posix()
 
+    if cad_format == "dagmc":
+        # for now
+        raise NotImplementedError(
+            "DAGMC export is not yet implemented in Bluemira. "
+            "Please use the FreeCAD GUI to export to DAGMC."
```
je-cook, line 1649: Can this be done or should this really say
```suggestion
            "This will be enabled soon."
```

bluemira/geometry/imprint_solids.py
```diff
@@ -0,0 +1,311 @@
...
+from typing import Any
+
+from rich.progress import track
+
+from bluemira.base.look_and_feel import bluemira_print
+from bluemira.geometry.overlap_checking import find_approx_overlapping_pairs
+from bluemira.geometry.solid import BluemiraSolid
+
+try:
+    from OCC.Core.BOPAlgo import BOPAlgo_MakeConnected
```
je-cook, line 20: Maybe a note about the use of pythonocc here
```suggestion
    # TODO @oliverfunk: pythonocc exposes functionality that cant be accessed 
    # through the freecad API. Should it become available this 
    # should be moved to freecadapi
    from OCC.Core.BOPAlgo import BOPAlgo_MakeConnected
```

je-cook (changes requested):
As we've discussed I'm a bit uncomfortable putting low level cad api stuff in the user facing geometry module. I am wonder about whether some of this should live in codes for that reason.
That would possibly then reduce the higher level developer facing code here (ie in geometry) quite substantially.
Is there anything a user/developer will interact with other than the `imprint_solids` function for imprint_solids.py?
Same conversation with the CGAL stuff I feel like that should just be exposed as an interface from codes with `find_approx_overlapping_pairs` and possibly a few others being exposed at the geometry level.

Could we also have a few more individual tests, it will help with working out what is going on in future (especially with overlap checking). 

The code itself looks fine (as much as I can join together in my head atm), I think its just a question of where we put it.

je-cook (commented):
Yeah just the one comment. I'll approve after we've chatted about environment dependency stuff

bluemira/codes/cgal_ext/_guard.py
```diff
@@ -0,0 +1,20 @@
...
+# SPDX-License-Identifier: LGPL-2.1-or-later
+"""
+API guard for the CGAL package.
+"""
+
+from bluemira.codes.utilities import code_guard, is_code_available
+
+
+def cgal_available():
+    is_code_available("CGAL")
```
je-cook, line 14: this does nothing atm is there meant to be a return? Unless it actually puts the import into sys.modules (I dont think it does?)

tests/codes/test_python_occ.py
```diff
@@ -0,0 +1,58 @@
...
+
+import pytest
+
+from bluemira.codes import python_occ
+from bluemira.geometry.face import BluemiraFace
+from bluemira.geometry.tools import extrude_shape, make_polygon
+
+
+class TestImprintSolids:
+    @pytest.mark.parametrize(
```
je-cook, line 16: will probably need to add a skip if import not available, we do it for a few other tests with openmc polyscope and process

https://docs.pytest.org/en/4.6.x/skipping.html#skipping-on-a-missing-import-dependency

je-cook (changes requested):
Few more but much less major

bluemira/codes/cgal_ext/_guard.py
oliverfunk: Oops yeah!

je-cook (approved):

Issue #3808, opened 2025-02-10
Create a few coilsets from known devices for testing purposes


je-cook:
To make our coilset and surrounding optimisation problems more robust it would be good to have a set of coilsets.
Some coilsets to have:

- [x] ITER
- [x] MAST-U
- [ ] ASDEX
- [ ] STEP CML5

PR #3910, opened 2025-05-06
Add some coilsets and a plotting notebook

geograham:
## Linked Issues

Papercut #3808 

## Checklist


je-cook (changes requested):
Thanks for this. The docs build error is because the example is not in the documentation toctree (I think).
Could we move the data to tests/equilibria/test_data and then in the notebook get it via the relative path to the example?

examples/equilibria/quick_loot_at_example_coilsets.ex.py
```diff
@@ -0,0 +1,93 @@
...
+# SPDX-FileCopyrightText: 2021-present J. Morris, D. Short
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+"""
+Notebook displaying example coilset positions for different devices.
+To be used in examples and tests.
+"""
+
+# In[1]:
```
je-cook, line 13: This needs to be in pypercent format otherwise the documentation won't be able to render it. jupytext can do this automatically in a jupyter notebook. https://jupytext.readthedocs.io/en/latest/formats-scripts.html#the-percent-format

je-cook (approved):
Looks good thanks very much, just need to fix the documentation build

examples/equilibria/quick_loot_at_example_coilsets.ex.py
geograham: Sorry I missed this, it is now percent

PR #3915, opened 2025-05-08
🏗️ Qtapp robustness improvement

je-cook:
## Description

There has been some reports of the qt cad viewer crashing with

```
QWidget: Must construct a QApplication before a QWidget
Aborted (core dumped)
```
this is not great and it is hard to see what we are doing wrong. So I've extracted all QApplication calls to one place and made it a try except instead.

Still needs a bit of checking

## Interface Changes

None

## Checklist

### DIFF
--- a/bluemira/codes/_freecadapi.py
+++ b/bluemira/codes/_freecadapi.py
@@ -2923,7 +2923,7 @@ def show_cad(
         light.intensity = 1.2
 
         viewer.show()
-        app.exec_()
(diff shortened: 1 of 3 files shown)


je-cook:
this doesnt solve the problem apparently, I will open an issue to that effect. 
However it is a cleanup so I think should go in as is as it isolates qt some more

oliverfunk (approved):

PR #3929, opened 2025-05-16
Fix loadspec index error

je-cook:
## Description

Loadspec used in get_module fialed with an index error if you passed in the string eg 'OCC'. It should have failed with a descriptive error.

## Checklist

### DIFF
--- a/bluemira/utilities/tools.py
+++ b/bluemira/utilities/tools.py
@@ -992,7 +992,17 @@ def get_module(name: str) -> ModuleType:
 
     Returns
     -------
-    Loaded module
+    :
+        Loaded module
+
+    Raises
+    ------
+    ImportError
+        Unable to import module
+    FileNotFoundError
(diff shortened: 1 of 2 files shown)


bluemira/utilities/tools.py
```diff
@@ -992,7 +992,17 @@ def get_module(name: str) -> ModuleType:
 
     Returns
     -------
-    Loaded module
+    :
```
oliverfunk, line 995: I think it makes more sense that this func only raises an ImportError

The code to look something like this

```
    try:
        return imp(name)
    except ImportError:
        pass
    try:
        return _loadfromspec(name)
    execept AllTheErrors
        raise ImportError
   ```

oliverfunk (approved):
Nice I just think it would better if get_module raised only an ImportError

oliverfunk (approved):
nice !

PR #3930, opened 2025-05-16
Functionalised join_intersect.

### LINKED ISSUE #3926: join_intersect should not modify the input argument

# Overview
`bluemira.geometry.coordinates.join_intersect` checks for intersection between a target polygon and a reference polygon; and then add these intersection points into the appropriate order into the target polygon, such that the target polygon increases in number of vertices=number of points.

## Why is it bad
1. It modifies an input parameter silently without notifying the user.
2. Also, it modifies the property of the target polygon: if it was a purely-convex polygon before, then you have a non-convex (but also non-concave) after this modification.
3. More importantly, it is not immediately obvious that the target polygon is the one getting modified while the reference polygon remains unchanged.

### Proposed solution

Let `join_intersect` Return one/two parameters rather zero/one parameter, where the extra parameter is a "dest_poly" (polygon in the destination location) containing the Coordinates of the target_polygon as well as the intersection points.

### Alternative solutions

Ignore this bug and leave it unchanged. This means we may have a harder time debugging if we ever write any code that refactors join_intersect.

### Additional Context

The following places uses join_intersect. We need to examine each file and apply the fix downstream from there.
Tick marks indicate if we have checked if it is safe to apply the proposed solution in the following files:
- [x] bluemira/equilibria/find_legs.py
    - [x] _extract_leg uses it, and have confirmed with @geograham that it is safe to NOT modify flux_line (the argument of `_extract_legs` that is currently being mutated by this join_intersect bug), and instead write something like `longer_flux_line, args_inters = join_intersect(flux_line, radial_line, get_arg=True)` to replace the [current line](https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/6d918f386146496efe7120f7aac095496bfe5099/bluemira/equilibria/find_legs.py#L504).
- [ ] bluemira/equilibria/flux_surfaces.py
- [x] bluemira/geometry/coordinates.py
    - That's the module itself, so it's already covered.
- [x] tests/geometry/test_coordinates.py
    - easy to apply changes as tests are standalone

## extra-additional context
`join_intersect` currently calls `get_intersect`, and the former re-do the work done the latter, by deducing which edges of the target_polygon did get_intersect find intersections from, and then add them there.

Maybe it is possible to make this less cumbersome by writing a better `join_intersect` that directly copies from `get_intersect`, or add a conditional branch in `get_intersect` that also records the indices of the edges being intersected with.

OceanNuclear:
## Linked Issues

Closes #3926 

## Description
Changed the signature of join_intersect.

## Interface Changes

join_intersect has been changed. But looks like no end-user will see/use this function directly.

## Checklist

### DIFF
--- a/bluemira/equilibria/flux_surfaces.py
+++ b/bluemira/equilibria/flux_surfaces.py
@@ -142,7 +142,7 @@ class FluxSurface:
             An instance of FluxSurface, with exactly the same class and coordinates,
             but sharing a different underlying copy of the data (Coordinates).
         """
-        return self.__class__(self.coords.copy())
+        return type(self)(self.coords.copy())
 
 
 class ClosedFluxSurface(FluxSurface):
--- a/bluemira/equilibria/flux_surfaces.py
+++ b/bluemira/equilibria/flux_surfaces.py
@@ -481,8 +481,7 @@ class PartialOpenFluxSurface(OpenFluxSurface):
 
         # Because we oriented the Coordinates the "right" way, the first intersection
         # is at the smallest argument
-        shortened_coords = coords_union_inters[:, : min(args) + 1]
-        self.coords = Coordinates(shortened_coords)
+        self.coords = Coordinates(coords_union_inters[:, : min(args) + 1])
 
         fw_arg = int(first_wall.argmin([self.x_end, 0, self.z_end]))
 
--- a/bluemira/equilibria/flux_surfaces.py
+++ b/bluemira/equilibria/flux_surfaces.py
@@ -10,7 +10,6 @@ Flux surface utility classes and calculations
 
 from __future__ import annotations
 
-from copy import deepcopy
 from dataclasses import asdict, dataclass
 from functools import lru_cache
 from typing import TYPE_CHECKING
@@ -136,8 +135,14 @@ class FluxSurface:
     def copy(self):
         """
         Make a deep copy of the FluxSurface.
-        """  # noqa: DOC201
-        return deepcopy(self)
+
+        Returns
+        -------
+        :
+            An instance of FluxSurface, with exactly the same class and coordinates,
+            but sharing a different underlying copy of the data (Coordinates).
+        """
+        return self.__class__(self.coords.copy())
 
 
 class ClosedFluxSurface(FluxSurface):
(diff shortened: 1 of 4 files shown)


bluemira/equilibria/flux_surfaces.py
```diff
@@ -136,8 +135,14 @@ def plot(self, ax: plt.Axes | None = None, **kwargs):
...
-        """  # noqa: DOC201
-        return deepcopy(self)
+
+        Returns
+        -------
+        :
+            An instance of FluxSurface, with exactly the same class and coordinates,
+            but sharing a different underlying copy of the data (Coordinates).
+        """
+        return self.__class__(self.coords.copy())
```
je-cook, line 145:
```suggestion
        return type(self)(self.coords.copy())
```

bluemira/equilibria/flux_surfaces.py
```diff
@@ -477,7 +481,8 @@ def clip(self, first_wall: Coordinates):
 
         # Because we oriented the Coordinates the "right" way, the first intersection
         # is at the smallest argument
-        self.coords = Coordinates(self.coords[:, : min(args) + 1])
+        shortened_coords = coords_union_inters[:, : min(args) + 1]
+        self.coords = Coordinates(shortened_coords)
```
je-cook, line 485:
```suggestion
        self.coords = Coordinates(coords_union_inters[:, : min(args) + 1])
```

bluemira/geometry/coordinates.py
```diff
@@ -1779,6 +1779,17 @@ def __iter__(self):
...
+    def copy(self):
+        """
+        Copy of itself
+
+        Returns
+        -------
+        An instance of Coordinates, with exactly the same class and coordinates,
+        but sharing a different underlying copy of the data array.
+        """
+        return self.__class__(self.as_dict())
```
je-cook, line 1791: This is a bit dangerous because the underlying arrays are not copied therefore the whole stack of copied coords is mutable. Also 
```suggestion
        return type(self)(self.as_dict())
```

je-cook (changes requested):
One pretty dangerous point that needs to be covered eg copying and changing the copy might change the original. We need to disconnect the numpy array pointers

OceanNuclear:
Not sure the reason for type(self) over self.__class__ (I suppose it's something about unifying the style?), but I've implemented that regardless. I thought as_dict would've handled the copying, but I guess not, fixed that as well now :)

je-cook:
> Not sure the reason for type(self) over self.**class** (I suppose it's something about unifying the style?), but I've implemented that regardless.

Its semi stylistic, on the whole we should avoid using dunder methods in code or at least minimise their usage. Using type directly makes sure that any odd edge cases are not our responsibility.

OceanNuclear:
Have checked that all copying functions make copies of the numpy matrix, rather than a pointer to the matrix itself/ a slice of the matrix etc.

PR #3937, opened 2025-05-20
Fix overplotting

geograham:
The physics analysis plots where overpotting onto the previous plots when run as a python file rather than a notebook. This should remove this problem in the examples found in BM, BM training and STEP.

## Checklist

### DIFF
--- a/bluemira/equilibria/analysis.py
+++ b/bluemira/equilibria/analysis.py
@@ -547,7 +547,7 @@ class EqAnalysis:
             raise BluemiraError(
                 "This function can only be used for Free Boundary Equilbria."
             )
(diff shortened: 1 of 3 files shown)


je-cook:
I get this error on the analysis example 
image

je-cook:
Also can you fix the new warnings in test_plotting (https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/9415c78fc3f6328af8f8a94eebfb00ac4e2a676f/tests/equilibria/test_analysis.py#L70) because of these fixes. If they are in a few separate tests the closing is covered by some fixtures that love in contest.py. I think the error is because your creating so many figures in a single test

PR #3979, opened 2025-06-27
Change neutron source for OpenMC

### LINKED ISSUE #4045: Change neutron source to a public repository

Presently we use our own internal, closed-source, pps-isotropic package for the neutron source in a tokamak. At present, it is incorrectly used, I suspect, in terms of the "peaking_factor", which should relate to the plasma density profile, but appears to have a different form in the underlying package.

As we in any case want to transition to a more flexible neutron source, that is more faithful to arbitrary plasma profiles, I suggest moving to the OS https://github.com/fusion-energy/openmc-plasma-source package for the time being. This source should be used in both the 2-D and 3-D OpenMC pipelines.

### Proposed solution

_No response_

### Alternative solutions

_No response_

### Additional Context

_No response_

oliverfunk:
## Linked Issues

Closes #4045 

## Description

Adds a Tokamak Plasma source to the run_cad_neutronics example

## Checklist

### DIFF
--- a/bluemira/codes/openmc/solver.py
+++ b/bluemira/codes/openmc/solver.py
@@ -490,7 +490,7 @@ class OpenMCNeutronicsSolver(CodesSolver):
         if run_mode is OpenMCRunModes.RUN_AND_PLOT:
             for run_mode in (OpenMCRunModes.PLOT, OpenMCRunModes.RUN):
                 result = self._single_run(
-                    run_mode, self.eq, source_params, runtime_params, debug=debug
+                    run_mode, source_params, runtime_params, debug=debug
                 )
             return result
         return self._single_run(run_mode, source_params, runtime_params, debug=debug)


pyproject.toml
```diff
@@ -92,6 +92,7 @@ radiation = ["cherab"]
 dagmc = [
     "cgal>=6.0.1",
     "fast_ctd @ git+https://github.com/Fusion-Power-Plant-Framework/fast_ctd@0.1.1",
+    "openmc_plasma_source>=0.4.0",
```
je-cook, line 95: Can we move this into the openmc extras as its not a req for dagmc

OceanNuclear:
Writing what I learned today down so it's at least documented somewhere:

openmc_plasma_source's tokamak_source makes a randomly generated cloud of 10000 points, each one revolved into a 1d ring source. The sampling is a uniform distribution in minor radius × a uniform distribution in poloidal angle. 
See [openmc-plasma-source](https://github.com/fusion-energy/openmc-plasma-source/blob/main/src/openmc_plasma_source/tokamak_source.py):
```python
    a = rng.random(sample_size) * minor_radius
    alpha = rng.random(sample_size) * 2 * np.pi
    ...
    RZ = tokamak_convert_a_alpha_to_R_Z(
        a=a,
        alpha=alpha,
        ...
    )
    ...
    for reaction in reactions: # "DD", "DT", "TT"
        neutron_source_density[reaction] = tokamak_neutron_source_density(
            fuel_density, temperatures, reaction
        )
        ...
        strengths = neutron_source_density[reaction] / total_source_density
        sources = tokamak_make_openmc_sources(
            strengths=strengths,
            ...
            temperatures=temperatures,
            fuel=fuel,
            RZ=RZ,
        )
        all_sources = all_sources + sources 
```

IMO this would artificially inflate the number of neutron sources near the center of the plasma. I would have preferred a 100×100 (or n×m) grid of rectangles, each revolved into 3D annular sources.
DT-dist

bluemira/codes/openmc/sources.py
```diff
@@ -7,43 +7,59 @@
...
+        ion_density_pedestal=source_parameters.ion_density_ped,
+        ion_density_peaking_factor=source_parameters.ion_density_alpha,
+        ion_density_separatrix=source_parameters.ion_density_sep,
+        ion_temperature_centre=source_parameters.ion_temperature_core,
+        ion_temperature_pedestal=source_parameters.ion_temperature_ped,
+        ion_temperature_separatrix=source_parameters.ion_temperature_sep,
+        ion_temperature_peaking_factor=source_parameters.ion_density_alpha,
+        ion_temperature_beta=source_parameters.ion_temperature_beta,
+        # shaping
+        shafranov_factor=source_parameters.shaf_shift,
```
CoronelBuendia, line 59: As discussed, this should be in [m], and converted to [cm]

bluemira/codes/openmc/sources.py
```diff
@@ -7,43 +7,59 @@
...
-        triangularity=source_parameters.plasma_physics_units.triangularity,
-        # plasma geometry
-        peaking_factor=source_parameters.plasma_physics_units.peaking_factor,
-        temperature=source_parameters.plasma_physics_units.temperature,
-        radial_shift=source_parameters.plasma_physics_units.shaf_shift,
-        vertical_shift=source_parameters.plasma_physics_units.vertical_shift,
-        # plasma type
-        mode="DT",
+        major_radius=raw_uc(source_parameters.major_radius, "m", "cm"),
+        minor_radius=raw_uc(source_parameters.minor_radius, "m", "cm"),
```
CoronelBuendia, line 44: Minor comment, but I think this would help reviewers in future: all the other unit conversions are done in the params class, except these. I think we should keep all the conversions in one place.

CoronelBuendia (changes requested):
Minor, but want to be sure. LGTM otherwise!

eudemo/eudemo/neutronics/run.py
```diff
@@ -100,17 +100,12 @@ def run_neutronics(
...
-            from bluemira.codes.openmc.sources import make_pps_source  # noqa: PLC0415
-        except ImportError:
-            raise NeutronicsError("Cannot import neutronics source") from None
 
     solver = neutronics_code_solver(
         params,
         build_config,
         neutronics_csg,
-        source=source or make_pps_source,
+        source=source or make_tokamak_source,
```
je-cook, line 108: This currently doesnt work because the solver expects this type:

https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/43a75afbc18f5b0d0cfa09a07f332bc3b0d37a0e/bluemira/codes/openmc/solver.py#L420

whereas it needs at least an eqdsk too in make_tokamak_source. Also do we want to make it from an eqdsk or from an equilibria object?

Options I can think of:

Pass in an half initialised source.
Rework it so that the flux map is part of the plasma_parameters
initialise the source beforehand (this will mean that the plasma_parameters may become out of sync maybe)

tests/codes/openmc/test_source_params.py
```diff
@@ -0,0 +1,75 @@
...
+with open(PARAMS_PATH) as j:
+    params = json.load(j)
+openmc_solver_pm_frame = make_parameter_frame(
+    params, OpenMCNeutronicsSolverParams, allow_unknown=True
+)
+pms_params = PlasmaSourceParameters.from_parameterframe(openmc_solver_pm_frame)
+
+
+def test_pedestal_location_in_normalized_radius():
+    assert pms_params.rho_pedestal == 0.94  # [dimensionless]
```
je-cook: Need this on the tests in this module.
```suggestion
    assert pms_params.rho_pedestal == pytest.approx(0.94)  # [dimensionless]
```

pyproject.toml
```diff
@@ -85,8 +85,12 @@ pinned = [
 ]
 
 conda_pinned = ["fenics-dolfinx==0.9.0", "freecad==1.0.0"]
-openmc = ["openmc>=0.14.0", "openmc_data==2.3.1"]
 process = ["process @ git+https://github.com/ukaea/process@v3.2.0"]
+openmc = [
+    "openmc>=0.14.0",
+    "openmc_data==2.3.1",
+    "tokamak-neutron-source @ git+ssh:github.com/Fusion-Power-Plant-Framework/tokamak-neutron-source",
```
je-cook: Once the package is opened up (before we have a version)
```suggestion
    "tokamak-neutron-source @ git+https://github.com/Fusion-Power-Plant-Framework/tokamak-neutron-source@main",
```

eudemo/eudemo/neutronics/run.py
CoronelBuendia: OK @je-cook, following our discussion, I changed the signature to take an Equilibrium, as agreed.

I can run EU-DEMO up to the neutronics now (will clean up the config file when I'm done), but I'm not convinced I haven't made this more complicated that it needs to be... It felt like I added the Equilibrium everywhere as an argument - not sure if this was really necessary..

bluemira/codes/openmc/sources.py
```diff
@@ -5,46 +5,95 @@
...
+        The dimension of the squares with which to discretise the neutron source
+
+    Returns
+    -------
+    source:
+        Fusion source for OpenMC
+    """
+    data = eq.to_dict()
+    data["name"] = ""
+    eq = EQDSKInterface(**data)
```
CoronelBuendia: This is also moderately dodgy, but it feels nuts to have to save an Equilibrium file just in order to do this.

bluemira/codes/openmc/sources.py
```diff
@@ -5,45 +5,101 @@
...
+    density_profile.set_scale(
+        source_parameters.va_fuel_ion_density / source_parameters.va_electron_density
+    )
+    transport = TransportInformation.from_parameterisations(
+        ion_temperature_profile=temperature_profile,
+        fuel_density_profile=density_profile,
+        rho_profile=rho_profile,
+        fuel_composition=FractionalFuelComposition(D=0.5, T=0.5),
+    )
+    flux_map = FluxMap.from_eqdsk(eq)
```
je-cook: Instead of this how about creating a subclass of FluxMap for bluemira

```
class BMFluxMap(FluxMap):
    @classmethod
    def from equilibrium(cls, eq: Equilibria):
            ....
            return cls(...)
```
then you avoid the making of an eqdsk unecesarrily as you can just exract directly what you need

tests/codes/openmc/test_source_params.py
```diff
@@ -0,0 +1,75 @@
...
+with open(PARAMS_PATH) as j:
+    params = json.load(j)
+openmc_solver_pm_frame = make_parameter_frame(
+    params, OpenMCNeutronicsSolverParams, allow_unknown=True
+)
+pms_params = PlasmaSourceParameters.from_parameterframe(openmc_solver_pm_frame)
+
+
+def test_pedestal_location_in_normalized_radius():
+    assert pms_params.rho_pedestal == pytest.approx(0.94)  # [dimensionless]
```
je-cook, line 29: this is needed for all in this file

eudemo/eudemo/neutronics/run.py
je-cook: I know what you mean. Trying to think of better ways:

1) class with `__call__` method for the source. Then we create the class with an equilibria as an init arg and eq is not needed to be passed in (we can revert to the previous signature)
```python
class NS:
    def __init__(self, eq):
    
    def __call__(self, params) -> openmc_source:
```
2) ....Will come back to this if I have more ideas

bluemira/codes/openmc/sources.py
je-cook: see flux map idea below?

eudemo/eudemo/neutronics/run.py
CoronelBuendia: I haven't touched it. I think this needs to be reviewed alongside our user-facing API, as I feel I need to move on...

pyproject.toml
```diff
@@ -85,8 +85,13 @@ pinned = [
 ]
 
 conda_pinned = ["fenics-dolfinx==0.9.0", "freecad==1.0.0"]
-openmc = ["openmc>=0.14.0", "openmc_data==2.3.1"]
+
 process = ["process @ git+https://github.com/ukaea/process@v3.2.1"]
+openmc = [
+    "openmc>=0.14.0",
+    "openmc_data==2.3.1",
+    "tokamak-neutron-source @ git+https://github.com/Fusion-Power-Plant-Framework/tokamak-neutron-source@main",
```
je-cook:
```suggestion
    "tokamak-neutron-source >=0.1.1",
```

bluemira/codes/openmc/solver.py
```diff
@@ -471,14 +490,15 @@ def execute(self, *, debug=False) -> OpenMCResult | dict[int, float]:
         if run_mode is OpenMCRunModes.RUN_AND_PLOT:
             for run_mode in (OpenMCRunModes.PLOT, OpenMCRunModes.RUN):
                 result = self._single_run(
-                    run_mode, source_params, runtime_params, debug=debug
+                    run_mode, self.eq, source_params, runtime_params, debug=debug
                 )
             return result
         return self._single_run(run_mode, source_params, runtime_params, debug=debug)
```
je-cook, line 496:
```suggestion
        return self._single_run(run_mode, self.eq, source_params, runtime_params, debug=debug)
```

CoronelBuendia:
OK so EU-DEMO doesn't actually run as is in this PR, due to some hard-coded thicknesses being used, which I had never noticed before. I opted not to fix that in this PR though.

EDIT: Obviously, I modified the thicknesses locally and the source does actually run as intended.

je-cook (approved):

PR #3988, opened 2025-06-30
Ocean/pc 3040 docstring

### LINKED ISSUE #3040: Minor docstring/behaviour mismatch in 2D and 3D coordinates/geometries.
## Describe the papercut

The input to bluemira.geometry.coordinates.normal_vector should be of shape (2, N), not (N, 2).

This is so minor that it'll take about 10 seconds to fix. But I just needed to raise a papercut because I'm currently busy working on a branch that I've not committed.

To turn this into a proper papercut, maybe I'll propose one more thing to do/check:

1. Run `grep -r "2, N"`, `grep -r "N, 2"`, `grep -r "3, N"`, `grep -r "N, 3"` on bluemira/geometry/ and bluemira/codes/. They should return about 15 instances. My question is: why do some use (2, N)/(3, N) format, while others take the (N, 2)/(N, 3) format? Is there a rhyme or reason, and if not, can we have the same format everywhere? (I guess the papercut can be to turn this into a discussion if we don't come to an agreement? Or simply make the change and commit it if the conclusion is simple and straightforward.)

OceanNuclear:
## Linked Issues

Closes #3040 

## Description

Minor docstring changes, and modified one function (used only in neutronics) to use the (2, N) signature instead of (N, 2).

## Interface Changes

`bluemira.geometry.tools.polygon_revolve_signed_volume`'s input requires transposing.

## Checklist

### DIFF
--- a/bluemira/codes/openmc/make_csg.py
+++ b/bluemira/codes/openmc/make_csg.py
@@ -118,7 +118,7 @@ class CellStage:
         # self.universe.volume = total_universe_volume
 
         outer_boundary_volume = to_cm3(
-            polygon_revolve_signed_volume(ext_vertices[:, ::2])
+            polygon_revolve_signed_volume(ext_vertices[:, ::2].T)
         )
         ext_void_volume = total_universe_volume - outer_boundary_volume
         if self.tf_coils:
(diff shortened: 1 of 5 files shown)


je-cook (approved):
thanks!

PR #4002, opened 2025-07-09
WIP: Magnet module optimisation problem generalisation attempt

CoronelBuendia:
## Description

This was my attempt at generalising the TF + WP optimisation problem, such that in future new objectives and constraints could be added more flexibly.

I also tried to make the interface a little less Russian-doll like, such that one could specify e.g. which conductor one wanted without first having to initialise a strand and cable. Because of the existing registry/metaclass stuff, I had to do this in a hacky way. I didn't want to change too much of the underlying structure so I found a way to work with what was there (I am NOT suggesting we do what I have done here, obviously!)

With the current structure, this does not really work for the following reasons:
* The defensive ValueErrors for unphysical geometry make it impossible for the optimiser to recover from infeasible situations
* I tried to replace this with some geometry constraints, but I am not even sure I implemented these correctly, and in any case I still run into ValueErrors
* The `rearrange_..` method obviously means that the gradients are going to go haywire, which something like COBYLA should be able to handle, but yeah, impossible to really get it going at the moment.
* The data structure is nested and complex, and depends heavily on properties. I am therefore not sure I have correctly implemented the `update` method in the new example file.
* It's clear that in the present structure, one could solve for the stabiliser strands separately, but in the spirit of flexibility I have opted to do this simultaneously, which clearly is harder.


CoronelBuendia:
@ivanmaione

CoronelBuendia:
There are no intentions of merging this

PR #4035, opened 2025-08-04
WIP Toroidal Harmonics Brute Force

clmould:
## Linked Issues

Closes #{ID}

## Description

Implemented a brute force method of finding the significant harmonics to use.
TODO:
- [ ] Fix broken test - test_ToroidalHarmonicConstraintFunction
- [ ] Add tests for no cos/sin degrees chosen
- [ ] Redo the notebooks, combine into one instead of split across 3
- [ ] Add unit tests
- [ ] Add in optimisation into a notebook - currently cut this out of toroidal_harmonics_approximation_basic_use.ex.py because it was not working properly

## Checklist


clmould:
Deleted a notebook, will update and add it back in along with documentation in #4095 and get this done before the next release.

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
```diff
@@ -422,224 +527,662 @@ def toroidal_harmonic_approximate_psi(
...
-    )
+        A_cos = np.sqrt(2 / np.pi) * (  # noqa: N806
+            np.einsum(
+                "ijkl, i, kl, j -> kl",
+                Am_cos_matrix,
+                epsilon,
+                np.sqrt(Delta),
+                currents,
+            )
+        )
```
je-cook: this could all be a separate function maybe? Same for the sin. I'm probably mistaken here but they nearly look structurally the same apart from the choice of a few arguments. Meaning it could be one function and then here it will be two lines:

```python
A_cos = func(...)
A_sin = func(...)
```

je-cook (commented):
Overall looks good to me structurally. One change possibly to reduce duplication slightly

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
geograham: You are right, have changed to one function

bluemira/equilibria/optimisation/harmonics/harmonics_constraints.py
```diff
@@ -200,40 +196,45 @@ def __init__(
...
             self.target_harmonics_cos = np.append(
-                ref_harmonics_cos, ref_harmonics_cos, axis=0
+                ref_harmonics_cos_amplitudes, ref_harmonics_sin_amplitudes, axis=0
             )
-            self.target_harmonics_sin = np.append(
-                ref_harmonics_sin, ref_harmonics_sin, axis=0
+            self.target_harmonics_sin = -np.append(
+                ref_harmonics_cos_amplitudes, ref_harmonics_sin_amplitudes, axis=0
             )
+            self.tolerance = np.tile(tolerance, 2)
```
CoronelBuendia, line 228: Can't put in a "suggestion" on Github, but here is one anyway:

```python
            self.target_harmonics_cos = np.append(
                ref_harmonics_cos_amplitudes, ref_harmonics_sin_amplitudes, axis=0
            ) + tolerance
            self.target_harmonics_sin = -np.append(
                ref_harmonics_cos_amplitudes, ref_harmonics_sin_amplitudes, axis=0
            ) - tolerance
            self.tolerance = np.zeros(len(self))
```

Here you are relying on the tolerance of the two competing inequality constraints being handled in the optimisation algorithm, but they are exactly conflicting here. Often the tolerance in optimisation algorithms is just enforced as a numerical thing separately to the true nature of the constraint.

If you run your notebook with `"inequality"`, it won't work well at all at present. If you directly account for the tolerances in the inequalities, it does a much better job.

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
```diff
@@ -422,224 +527,663 @@ def toroidal_harmonic_approximate_psi(
...
+    psi_norm: float = 0.95,
+    n_degrees_of_freedom: int | None = None,
+    max_harmonic_mode: int = 5,
     *,
-    plot: bool = False,
-) -> tuple[
-    ToroidalHarmonicsParams, np.ndarray, np.ndarray, int, float, np.ndarray, np.ndarray
-]:
+    plasma_mask: bool = False,
+    cl: bool = False,
```
CoronelBuendia: This is undocumented. Can we call it `collocation`?

It complicates the implementation quite significantly - can this not be split into two functions that share sub-functions? We can still have this as the parent function, it's just we are reading and doing maths on two things simultaneously, only to only use one of the things...

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
```diff
@@ -422,224 +527,663 @@ def toroidal_harmonic_approximate_psi(
...
+    )
+
+
+def plot_toroidal_harmonic_approximation(
+    eq: Equilibrium,
+    th_params: ToroidalHarmonicsParams,
+    result: ToroidalHarmonicsSelectionResult,
+    psi_norm: float = 0.95,
+    *,
+    cl: bool = False,
```
CoronelBuendia: Is this even necessary? It's just deciding which psi to plot... The psi in a result should just be the psi. Not carry around both and a flag.

CoronelBuendia (changes requested):
A few issues still with running the notebook with different combinations of inputs (for me at least). Until this is cleared up a couple from my side. One technical issue for sure, and one in terms of complexity which I think needs addressing.

bluemira/equilibria/optimisation/harmonics/harmonics_constraints.py
CoronelBuendia: You may be able to forget this... Now that I have understood how to run the notebook.

PR #4037, opened 2025-08-07
Created the correct tally for total power.

### LINKED ISSUE #4009: CSG neutronics total heating filter not producing correct results`
### Describe the bug

In `OpenMCResult`, if I add this:

```python
        total_power, total_power_err = cls._load_filter_power_err(statepoint, src_rate, "Total power")
        print(f"Total power: {total_power}")
 ```

I would expect to see the total power deposited in all materials in the model.

Instead, for reasons I don't understand, I see values much lower. When summing just blanket divertor and vessel power, and adding them together, in the EUDEMO example, I get:

Total power: 1508204413.664835
Total power sum: 2104545955.9506643

So something is going wrong.

### Steps to reproduce

_No response_

### Tracebacks

```bash

```

### Version

latest matti/je-cook/process_develop_EUDEMO_LAR

### OS

### OS details

_No response_

OceanNuclear:
## Linked Issues

Closes #4009 

## Description

Solves the issue of not having a tally for the total power.

## Interface Changes

`reactor.neutronics.results` (which is an `OpenMCResult` object) has two new attribute `total_power` and `total_power_err`.

The `divertor damage` tally should've also reduced as a result (previous definition of the div_cell_filter accidentally included the slices of the vacuum vessel surrounding the divertor as well.)

## Checklist

### DIFF
--- a/bluemira/codes/openmc/output.py
+++ b/bluemira/codes/openmc/output.py
@@ -22,7 +22,6 @@ from bluemira.base.parameter_frame._parameter import Parameter
 from bluemira.codes.openmc.make_csg import CellStage
 from bluemira.plasma_physics.reactions import E_DT_fusion
 from bluemira.radiation_transport.neutronics.constants import (
-    DPACoefficients,
     FE_DPA_THRESHOLD_EV,
     get_dpa_coefficients,
 )
--- a/bluemira/radiation_transport/neutronics/constants.py
+++ b/bluemira/radiation_transport/neutronics/constants.py
@@ -6,6 +6,7 @@
 """constants used for the neutronics module"""
 
 from periodictable import elements
+import numpy.typing as npt
 
 from bluemira.base.constants import (
     ELECTRON_MOLAR_MASS,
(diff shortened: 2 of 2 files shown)


je-cook:
We dont need a statepoint file. We just need to mock `cls._load_dataframe_from_statepoint` to output a pandas dataclass with some dummy contents. The test needs to be skipped if openmc is not installed because pandas is not a direct dependency to bluemira, only to openmc

OceanNuclear:
>includes deposition terms in the TF and radiation shield

Thanks for that, spotted the mistake in tallying.py, fixed in the latest commits.

bluemira/codes/openmc/output.py
```diff
@@ -448,13 +451,19 @@ def from_openmc_csg_result(cls, result: OpenMCResult):
         """
         source = "OpenMC CSG"
 
+        auxiliary_power = (
+            result.total_power
+            - result.blanket_power
+            - result.divertor_power
+            - result.vessel_power
+        )
```
CoronelBuendia, line 459: P_aux is usually defined for power deposited in "auxiliary" systems, which is typically used for in-vessel components that are not the blankets or divertor (e.g. HCD systems, port plugs, whatever).

Here we don't have any auxiliary systems modelled as it is axisymmetric.

CoronelBuendia:
Seems more reasonable now:

image

CoronelBuendia (approved):

PR #4046, opened 2025-08-15
Simple Panelling Optimisation Problem

athoynilimanew:
## Linked Issues

Adds a simple panelling (actually, discretising) optimisation problem , upon discussion with @OceanNuclear 

Closes #{ID}

## Description

Optimisation problem to minimise the change in panel areas (or volume) while panelling (read, discretising) a wire, assuming the wire is simple.

It will basically be used in Neutronics. The word "Panelling" does not necessarily need to mean the "designed as per engineering requirements" of first wall panels.

## Checklist

### DIFF
--- a/bluemira/geometry/tools.py
+++ b/bluemira/geometry/tools.py
@@ -16,6 +16,7 @@ import functools
 import inspect
 import json
 from collections.abc import Callable, Iterable, Sequence
+from copy import deepcopy
 from enum import Enum, auto
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Protocol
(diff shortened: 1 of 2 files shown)


CoronelBuendia:
@athoynilimanew is this still in works? I've not read this in detail, so I'm not entirely sure what the use case is when compared with already existing panelling stuff..?

athoynilimanew:
> @athoynilimanew is this still in works? I've not read this in detail, so I'm not entirely sure what the use case is when compared with already existing panelling stuff..?

I think it is a simple panelling routine that is only used for Neutronics Purposes (it is not a "Physical Panelling"). It was used for, e.g. the work I have done in neutronics. 

**However, it can be triaged as the neutronics work in Bluemira is now going to different directions and this should no longer be required.**

PR #4074, opened 2025-09-09
Fix SOL radiation solver source

### LINKED ISSUE #3944: SOL radiation solver producing strange source term and heat flux results
### Describe the bug

**Description**

The plotted heatmap of the SOL radiation source term exhibits visible discontinuities along flux surfaces, despite (my personal) expectations of a relatively continuous structure. These discontinuities appear to originate from the underlying source term values themselves.

Image

**Observed Behavior:**

The source term shows sharp changes or "jumps" in value between neighboring cells along a flux surface, creating diagonal banding.

**Expected Behavior:**

The source term should vary smoothly, without abrupt discontinuities, along a flux surface.

**Possible Causes:**

Incorrect offset, indexing, or interpolation during source term evaluation.

Note that it does not appear to be a source term plotting issue, as these discontinuities also appear in the first wall heat flux.

I suspect that the gridding of the source term is at least partly responsible for this strange heat map:

Image

We may just be looking at some interpolation understandably doing a poor job. We should be giving CHERAB the source term values equispaced along field lines. I remember mentioing this a couple of years ago when this was being developed; there may have been a reason for not doing that, but I cannot remember it.

### Steps to reproduce

Run `examples/radiation_transport/radiation_calculation_solve_DEMO.py` with L160 `only_source = True` on branch: `develop`

### Tracebacks

```bash

```

### Version

2.6.2

### OS

### OS details

_No response_

je-cook:
## Linked Issues

Closes #3944

superceeds #3965 but discussion is between those two

## Description

Uses interpolation of coordinates to rediscretise source term

@ajpearcey fyi

## Checklist

### DIFF
--- a/bluemira/radiation_transport/flux_surfaces_maker.py
+++ b/bluemira/radiation_transport/flux_surfaces_maker.py
@@ -373,13 +373,11 @@ def _make_flux_surfaces_ibob(
         outboard flux surfaces
     """
     sign = 1 if outboard else -1
-    rnge = np.arange(x_sep_mp + (sign * dx_mp), x_out_mp - (sign * EPS), (sign * dx_mp))
-    return list(
-        zip(
-            *[
-                _make_flux_surfaces(x, o_point.z, equilibrium, o_point, yz_plane, dl)
-                for x in rnge
-            ],
-            strict=False,
+    flux_surfaces = [
+        _make_flux_surfaces(x, o_point.z, equilibrium, o_point, yz_plane, dl)
+        for x in np.arange(
+            x_sep_mp + (sign * dx_mp), x_out_mp - (sign * EPS), (sign * dx_mp)
         )
-    )
+    ]
+
+    return tuple(map(list, zip(*flux_surfaces, strict=True)))


bluemira/radiation_transport/flux_surfaces_maker.py
```diff
@@ -334,16 +342,25 @@ def _make_flux_surfaces(
...
-    coords = find_flux_surface_through_point(
+    coords_arr = find_flux_surface_through_point(
         equilibrium.x, equilibrium.z, equilibrium.psi(), x, z, equilibrium.psi(x, z)
     )
-    return OpenFluxSurface(Coordinates({"x": coords[0], "z": coords[1]})).split(
-        o_point, plane=yz_plane
-    )
+    coords = Coordinates({"x": coords_arr[0], "z": coords_arr[1]})
+    if dl is not None:
+        coords = coords.interpolate(dl=dl)
```
je-cook, line 350: this is the important line, I'm not currently preserving the points of the original fs but that can change

PR #4090, opened 2025-09-22
Beta_p approximation caution

### LINKED ISSUE #4020: Check if beta_p approximation is appropriate for Spherical Tokamaks
### Describe the Task

Approximation is ok for conventional aspect ratios but we should verify it works for STs. 

### Current and Future State

_No response_

geograham:
## Linked Issues

Closes [#4020](https://github.com/Fusion-Power-Plant-Framework/bluemira/issues/4020)

## Description

At higher plasma elongations, so for spherical tokamaks, the approximation used in our BetaIpProfile class becomes less good (for DEMO-like elongation there is less than a 1% difference between the approximate and actual values but for a STEP-like elongation this is closer to 10%).

There is now a caution message in the relevant functions and the approximation calculation has been refactored so that is only in one location in the code.

## Checklist

### DIFF
--- a/bluemira/equilibria/constants.py
+++ b/bluemira/equilibria/constants.py
@@ -16,6 +16,10 @@ PSI_NORM_TOL = 1e-2
 #     Used as a convergence criterion for Picard iterations
 PSI_REL_TOL = 2e-3
 
+# Default psi_norm value [n. a.]
+#    Used in flux surface finding and masking functions.
+PSI_NORM = 1.0
+
 # Absolute tolerance on position [m]
 #     Used to determine whether O- and X-points are the "same"
 #     Used as an offset to determine if a point is "on" the edge of a coil
--- a/bluemira/equilibria/constants.py
+++ b/bluemira/equilibria/constants.py
@@ -16,10 +16,6 @@ PSI_NORM_TOL = 1e-2
 #     Used as a convergence criterion for Picard iterations
 PSI_REL_TOL = 2e-3
 
-# Default psi_norm value [n. a.]
-#    Used in flux surface finding and masking functions.
-PSI_NORM = 1.0
-
 # Absolute tolerance on position [m]
 #     Used to determine whether O- and X-points are the "same"
 #     Used as an offset to determine if a point is "on" the edge of a coil
--- a/bluemira/equilibria/equilibrium.py
+++ b/bluemira/equilibria/equilibrium.py
@@ -1666,7 +1666,7 @@ class Equilibrium(CoilSetMHDState):
         psi = self.psi()
         return calc_psi_norm(psi, *self.get_OX_psis(psi))
 
-    def pressure_map(self, pn=PSI_NORM) -> npt.NDArray[np.float64]:
+    def pressure_map(self, pn: float =PSI_NORM) -> npt.NDArray[np.float64]:
         """
         Parameters
         ----------
@@ -1684,7 +1684,7 @@ class Equilibrium(CoilSetMHDState):
         p = self.pressure(np.clip(self.psi_norm(), 0, 1))
         return p * mask
 
-    def _get_core_mask(self, pn=PSI_NORM) -> npt.NDArray[np.float64]:
+    def _get_core_mask(self, pn: float =PSI_NORM) -> npt.NDArray[np.float64]:
         """
         Parameters
         ----------
--- a/bluemira/equilibria/equilibrium.py
+++ b/bluemira/equilibria/equilibrium.py
@@ -50,7 +50,7 @@ from bluemira.equilibria.limiter import Limiter
 from bluemira.equilibria.num_control import DummyController, VirtualController
 from bluemira.equilibria.physics import (
     EqSummary,
-    calc_li3minargs,
+    _calc_li3minargs,
     calc_psi_norm,
 )
 from bluemira.equilibria.plasma import NoPlasmaCoil, PlasmaCoil
@@ -1446,7 +1446,7 @@ class Equilibrium(CoilSetMHDState):
 
             plasma_psi = self._solver(rhs)
             self._update_plasma(plasma_psi, jtor_opt)
-            li = calc_li3minargs(
+            li = _calc_li3minargs(
                 self.x,
                 self.z,
                 self.psi(),
@@ -1666,18 +1666,19 @@ class Equilibrium(CoilSetMHDState):
         psi = self.psi()
         return calc_psi_norm(psi, *self.get_OX_psis(psi))
 
-    def pressure_map(self) -> npt.NDArray[np.float64]:
+    def pressure_map(self, pn=None) -> npt.NDArray[np.float64]:
         """
         Returns
         -------
         :
             Plasma pressure map.
         """
-        mask = self._get_core_mask()
-        p = self.pressure(np.clip(self.psi_norm(), 0, 1))
+        mask = self._get_core_mask(pn)
+        pn_max = pn or 1
+        p = self.pressure(np.clip(self.psi_norm(), 0, pn_max))
         return p * mask
 
-    def _get_core_mask(self) -> npt.NDArray[np.float64]:
+    def _get_core_mask(self, pn=None) -> npt.NDArray[np.float64]:
         """
         Returns
         -------
@@ -1685,9 +1686,14 @@ class Equilibrium(CoilSetMHDState):
             A 2-D masking array for the plasma core.
         """
         o_points, x_points = self.get_OX_points()
-        return in_plasma(
-            self.x, self.z, self.psi(), o_points=o_points, x_points=x_points
+        if pn is None:
+            return in_plasma(
+                self.x, self.z, self.psi(), o_points=o_points, x_points=x_points
+            )
+        zone = self.get_flux_surface(
+            pn, self.psi(), o_points=o_points, x_points=x_points
         )
+        return in_zone(self.x, self.z, zone.xz.T)
 
     def q(
         self,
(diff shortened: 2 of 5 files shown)


bluemira/equilibria/equilibrium.py
```diff
@@ -1666,28 +1666,34 @@ def psi_norm(self) -> npt.NDArray[np.float64]:
         psi = self.psi()
         return calc_psi_norm(psi, *self.get_OX_psis(psi))
 
-    def pressure_map(self) -> npt.NDArray[np.float64]:
+    def pressure_map(self, pn=None) -> npt.NDArray[np.float64]:
```
je-cook: pn needs a type and a docstring entry

bluemira/equilibria/physics.py
```diff
@@ -267,12 +280,12 @@ def calc_li3(eq: Equilibrium) -> float:
...
     where: Bp is the poloidal magnetic field and V is the plasma volume
     """  # noqa: DOC201
-    return calc_li3minargs(
+    return _calc_li3minargs(
         eq.x, eq.z, eq.psi(), eq.Bp(), eq.profiles.R_0, eq.profiles.I_p, eq.dx, eq.dz
     )
 
 
-def calc_li3minargs(
+def _calc_li3minargs(
```
je-cook, line 413: Why has this been made a private method?

bluemira/equilibria/physics.py
```diff
@@ -313,12 +331,18 @@ def calc_p_average(eq: Equilibrium) -> float:
...
     """
-    p = eq.pressure_map()
-    v_plasma = calc_volume(eq)
-    return volume_integral(p, eq.x, eq.dx, eq.dz) / v_plasma
+    v_plasma = calc_volume(fs)
+    return _calc_p_int(pressure_map, x, dx, dz) / v_plasma
 
 
-def calc_beta_t(eq: Equilibrium) -> float:
+def calc_beta_t(
```
je-cook: Why the change from the simpler signature, its also api breaking

bluemira/equilibria/physics.py
```diff
@@ -333,11 +357,19 @@ def calc_beta_t(eq: Equilibrium) -> float:
...
     """
-    p_avg = calc_p_average(eq)
-    return 2 * MU_0 * p_avg / eq._B_0**2
+    p_avg = calc_p_average(pressure_map, fs, x, dx, dz)
+    # return 2 * MU_0 * p_avg / profiles._B_0**2
+    return 2 * MU_0 * p_avg / B_0**2
 
 
-def calc_beta_p(eq: Equilibrium) -> float:
+def calc_beta_p(
```
je-cook: Same here, api breaking

je-cook (changes requested):
Not hugely happy with API breaking changes at the moment

bluemira/equilibria/physics.py
geograham: Because we use it within calc_li3 in physics.py and solve_li in equilibrium.py - I don't there is a reason a user would need _calc_li3minargs rather than one of the functions that calls it.

bluemira/equilibria/equilibrium.py
geograham: Added

bluemira/equilibria/physics.py
geograham: Now with wrapper

bluemira/equilibria/equilibrium.py
```diff
@@ -1666,28 +1666,47 @@ def psi_norm(self) -> npt.NDArray[np.float64]:
         psi = self.psi()
         return calc_psi_norm(psi, *self.get_OX_psis(psi))
 
-    def pressure_map(self) -> npt.NDArray[np.float64]:
+    def pressure_map(self, pn=PSI_NORM) -> npt.NDArray[np.float64]:
```
je-cook:
```suggestion
    def pressure_map(self, pn: float =PSI_NORM) -> npt.NDArray[np.float64]:
```

bluemira/equilibria/equilibrium.py
```diff
@@ -1666,28 +1666,47 @@ def psi_norm(self) -> npt.NDArray[np.float64]:
...
         :
             Plasma pressure map.
         """
-        mask = self._get_core_mask()
+        mask = self._get_core_mask(pn)
         p = self.pressure(np.clip(self.psi_norm(), 0, 1))
         return p * mask
 
-    def _get_core_mask(self) -> npt.NDArray[np.float64]:
+    def _get_core_mask(self, pn=PSI_NORM) -> npt.NDArray[np.float64]:
```
je-cook:
```suggestion
    def _get_core_mask(self, pn: float =PSI_NORM) -> npt.NDArray[np.float64]:
```

je-cook (approved):
Much happier now, two minors and just need to fix the tests. I leave the physics basis to @CoronelBuendia

bluemira/equilibria/constants.py
```diff
@@ -16,6 +16,10 @@
 #     Used as a convergence criterion for Picard iterations
 PSI_REL_TOL = 2e-3
 
+# Default psi_norm value [n. a.]
+#    Used in flux surface finding and masking functions.
+PSI_NORM = 1.0
```
CoronelBuendia: Do you mean BOUNDARY_PSI_NORM, or LCFS_PSI_NORM? Having a default value for all PSI_NORM seems very odd...

bluemira/equilibria/equilibrium.py
```diff
@@ -1666,28 +1666,47 @@ def psi_norm(self) -> npt.NDArray[np.float64]:
...
         :
             Plasma pressure map.
         """
-        mask = self._get_core_mask()
+        mask = self._get_core_mask(pn)
         p = self.pressure(np.clip(self.psi_norm(), 0, 1))
         return p * mask
 
-    def _get_core_mask(self) -> npt.NDArray[np.float64]:
+    def _get_core_mask(self, pn: float = PSI_NORM) -> npt.NDArray[np.float64]:
```
CoronelBuendia: could we have `psi_n` here? `p` could be many other things.

bluemira/equilibria/equilibrium.py
```diff
@@ -1666,28 +1666,47 @@ def psi_norm(self) -> npt.NDArray[np.float64]:
         psi = self.psi()
         return calc_psi_norm(psi, *self.get_OX_psis(psi))
 
-    def pressure_map(self) -> npt.NDArray[np.float64]:
+    def pressure_map(self, pn: float = PSI_NORM) -> npt.NDArray[np.float64]:
```
CoronelBuendia: could we have `psi_n` here as the argument? `pn` is potentially confusing, especially with pressure here.

bluemira/equilibria/equilibrium.py
CoronelBuendia: There is potentially a big problem with using 1.0 here. Here `None` will default to the LCFS, which I think is what one would want. But 1.0 goes to `find_flux_surface`  and there are no guarantees at present the the resulting flux surface at psi_norm = 1.0 is closed. 

As you know, finding the LCFS and the separatrix is treated differently, because of numerical issues in contour interpolation. Please treat 1.0 as a special case, as we do elsewhere.

bluemira/equilibria/physics.py
```diff
@@ -206,26 +208,120 @@ def calc_dx_sep(eq: Equilibrium) -> float:
...
+    return _calc_Bp2_int(
+        Bp=eq.Bp(),
+        mask=in_plasma(eq.x, eq.z, eq.psi()),
+        x=eq.x,
+        dx=eq.dx,
+        dz=eq.dz,
+    ) / (2 * MU_0)
+
+
+def _calc_Li_from_energy(p_energy: float, i_p: float) -> float:
```
CoronelBuendia: Again, `p` here could be read as pressure, perhaps `bp_energy`. I appreciate `p` is also used a "poloidal", as in the ubiquitous `beta_p` (which I am not proposing to change).

bluemira/equilibria/physics.py
```diff
@@ -333,8 +537,45 @@ def calc_beta_t(eq: Equilibrium) -> float:
     -------
     Ratio of plasma to toroidal magnetic pressure
     """
-    p_avg = calc_p_average(eq)
-    return 2 * MU_0 * p_avg / eq._B_0**2
+    return _calc_beta_t(
+        eq.pressure_map(PSI_NORM), eq.get_LCFS(), eq.x, eq.dx, eq.dz, eq.profiles._B_0
```
CoronelBuendia: PSI_NORM = 1.0 != LCFS

CoronelBuendia (changes requested):
Looks fine mostly. Just one important point of precision regarding the use of `PSI_NORM = 1.0` and to basically avoid using this as an ersatz for what is or isn't considered "inside" the plasma. See `find_LCFS_separatrix` for details, but basically psi_norm=1.0 is likely but not guranteed to be closed, and psi_norm=1.0 will almost certainly not describe the LCFS (exactly).

Elsewhere (in similar situations), we deliberately using `None` to default to the LCFS. And we specifically have the `in_plasma` function for similar purposes.

bluemira/equilibria/constants.py
geograham: I did, but removed as not needed anymore

bluemira/equilibria/equilibrium.py
geograham: Thanks for spotting, definitely need to keep it special case

CoronelBuendia (approved):

PR #4123, opened 2025-10-21
Beta_p profile classes: use full calculation rather than approximation

### LINKED ISSUE #4098: Beta_p - use full calculation rather than approx?

Currently we use our approximation for beta_p in the BetaIpProfile class, this is not a good approximation for high elongation plasmas (see #4090).

### Proposed solution

We should test using the full calculation rather than the approximation - noting that this may be too slow.

### Alternative solutions

Leave as is.

### Additional Context

_No response_

geograham:
## Linked Issues

Closes #4098

## Description

Added the full beta_p calculation in the profile class (previously used an approx. that was not good for high elongation plasmas).

~~Now we use the full calculation as default. Option to use the approximation is still available (although time saved is not likely to be significant)~~. -> Removed approximation from the profile class calculations as time saved not significant. 

beta_p calculation time for three test equilibria:

Device | Grid Size | Full calc time (s) | Approx. calc time (s)
-- | -- |-- | --
STEP-like | 129x257  | 0.10 | 0.06
STEP-like | 65x65 | 0.5 | 0.03
MAST-U-like | 165x165  | 0.07 | 0.04
DEMO | 65x65 | 0.01 | 0.01

## Checklist

### DIFF
--- a/bluemira/equilibria/profiles.py
+++ b/bluemira/equilibria/profiles.py
@@ -28,13 +28,12 @@ from bluemira.equilibria.error import EquilibriaError
 from bluemira.equilibria.find import (
     OPointCalcOptions,
     Opoint,
-    find_LCFS_separatrix,
     in_plasma,
     in_zone,
     o_point_fallback_calculator,
 )
 from bluemira.equilibria.grid import integrate_dx_dz
-from bluemira.equilibria.physics import _calc_beta_p, _calc_beta_p_approx
+from bluemira.equilibria.physics import _calc_beta_p
 from bluemira.equilibria.plotting import ProfilePlotter
 
 if TYPE_CHECKING:
@@ -599,11 +598,8 @@ class BetaIpProfile(Profile):
         R_0: float,
         B_0: float,
         shape: ShapeFunction | None = None,
-        *,
-        use_approx_beta_p: bool = False,  # noqa: FBT001, FBT002
     ):
         self.betap = betap
-        self.use_approx_beta_p = use_approx_beta_p
         self.I_p = I_p
         self._fvac = R_0 * B_0
         self.R_0 = R_0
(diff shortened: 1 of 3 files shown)


bluemira/equilibria/profiles.py
```diff
@@ -599,8 +599,10 @@ def __init__(
         R_0: float,
         B_0: float,
         shape: ShapeFunction | None = None,
+        use_approx_beta_p: bool = False,  # noqa: FBT001, FBT002
```
je-cook: So that a user has to type the variable name
```suggestion
        *,
        use_approx_beta_p: bool = False,  # noqa: FBT001, FBT002
```

bluemira/equilibria/profiles.py
```diff
@@ -749,8 +773,11 @@ def __init__(
         shape: ShapeFunction | None = None,
         li_rel_tol: float = 0.015,
         li_min_iter: int = 5,
+        use_approx_beta_p: bool = False,  # noqa: FBT001, FBT002
```
je-cook:
```suggestion
        *,
        use_approx_beta_p: bool = False,  # noqa: FBT001, FBT002
```

je-cook (approved):
From my side looks fine, 2 minors (that are the same).

CoronelBuendia:
Thanks for this @geograham, looks very interesting. Could we see a comparison of the computed jtor values using approx and full calculation? Also think we need to take a look at the speed comparison on those three types of equilibria at different discretisations, e.g. 64x128, 128x256, and 256x512.

Debating whether or not the approximation should even be offered as an option, is all.

Before that speed comparison though, it might be useful to see if just passing in the eq and calling Bx and Bz from there speeds things up compared to recreating RBS locally here. I think they should be the same... The values at the grid points are at least partially stashed, and if not are called using existing RBSs.

geograham:
> Thanks for this @geograham, looks very interesting. Could we see a comparison of the computed jtor values using approx and full calculation? Also think we need to take a look at the speed comparison on those three types of equilibria at different discretisations, e.g. 64x128, 128x256, and 256x512.
> 
> Debating whether or not the approximation should even be offered as an option, is all.
> 
> Before that speed comparison though, it might be useful to see if just passing in the eq and calling Bx and Bz from there speeds things up compared to recreating RBS locally here. I think they should be the same... The values at the grid points are at least partially stashed, and if not are called using existing RBSs.

Herer is a jtor comparison for DEMO: 

image

I have put the grid info in the table in the description above and added the results of a test for a STEP-like equilibria with a DEMO grid size. 

As we discussed in the meeting earlier today - I don't think using the eq directly makes sense structurally for the profile classes.

CoronelBuendia:
Thanks @geograham, I think the speed-up with the approximation is not worth the inaccuracy - especially for STs. Please remove the approximation altogether - no need for an additional option here.

geograham:
> Thanks @geograham, I think the speed-up with the approximation is not worth the inaccuracy - especially for STs. Please remove the approximation altogether - no need for an additional option here.

I have removed - thanks for checking the test

CoronelBuendia (approved):

PR #4145, opened 2025-11-12
Add remaining th tests and fix excluded coils

### LINKED ISSUE #4099: Add collocation test for TH approximation
### Describe the Task

Add tests for:
- [x] `toroidal_harmonics_to_positions`
- [x] `_separate_psi_contributions`
- [x] `_get_plasma_mask`
- [x] `_approximation_from_psi_fitting`
- [x] `plot_toroidal_harmonic_approximation`
- [x] `_approximation_direct_from_currents`

### Current and Future State

NA

clmould:
## Linked Issues

Closes #4099 
Ensure excluded coils are calculated properly instead of just checking if the centre of the coil is inside the approximation region
Also some additions to th fn docstrings that were missing
Add a json file to contain large arrays needed for TH tests
Add unit tests where appropriate

## Checklist


bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
```diff
@@ -340,7 +340,7 @@ def toroidal_harmonic_grid_and_coil_setup(
     c_names = np.array(th_coilset.control)
 
     # Find coils that can be used in TH approximation, and those that cannot be used
-    if min_tau < np.min(tau_c):
+    if min_tau > np.min(tau_c):
```
je-cook: this seems like a big bug no, what does it effect?

je-cook (changes requested):
Is there any way we can reduce the amount of test data here that just hard coded in the test? 
Is there a better way to test this because of that?

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
clmould: yes... with the demo files i was using before, there are no excluded coils so i hadn't noticed this previously

je-cook: ~also do you want >= which is the opposite to the previous state~ nope ignore that

clmould: i'm keeping it as > , but will modify so that excluded coils are calculated from the inner corner, rather than their centre

je-cook: this is harder than it sounds IIRC, also will be generically useful for position optimisation as we currently do it on their centres.

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
```diff
@@ -1088,6 +1122,8 @@ def _approximation_from_psi_fitting(
     Approximate psi by fitting for psi at collocation points using
     toroidal harmonic equations
 
+    Note: len(mode_id) must = n_degrees_of_freedom
```
clmould: Not sure if this is needed, was kind of a note to myself as I forgot this and it caused my test-writing issues! @je-cook @geograham

clmould:
> Is there any way we can reduce the amount of test data here that just hard coded in the test? Is there a better way to test this because of that?

I have made some unit tests and also put large test data in a json file. Some functions require an equilibrium to be passed in, so I kept those as regression tests, but any that did not need the equilibrium I have made into a unit test instead. @je-cook

tests/equilibria/test_harmonics.py
```diff
@@ -1156,6 +1050,94 @@ def test_toroidal_harmonic_grid_and_coil_setup(self, cc_all):
...
+
+        np.testing.assert_array_almost_equal(
+            expected_tau_c, tau_limit_coil_th_params.tau_c
+        )
+        np.testing.assert_array_almost_equal(
+            expected_sigma_c, tau_limit_coil_th_params.sigma_c
+        )
+
     @pytest.mark.parametrize("cc_all", [True, False])
     def test_coil_toroidal_harmonic_amplitude_matrix(self, cc_all):
```
clmould, line 1142: I have kept this as a regression test as well so it can test with the `cc_all` options, but I can remove this and just have a unit test if that would be better. I thought I'd leave it for now as I'd like to be sure it works as expected with different control coils @geograham @je-cook

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
je-cook: If its not a speed critical bit of code you could put a quick if check and raise. Otherwise a section like this is preferable

```
Notes
-----
notes about stuff
```

tests/equilibria/test_harmonics.py
geograham: I would be keen for there to be a regression test for the control coils set differently but a unit test and a test that it returns the expected result when no e.g. sin components are selected might also be useful.

clmould: The regression has the cc_all options so is testing different control coils. In the unit test (higher up in this file) I have added cases where no cos are used and where no sin are used, so I think that covers all of this comment?

je-cook:
Can you fix the new warnings in ^this comment?

tests/equilibria/test_harmonics.py
geograham: Oh, sorry I forgot you already added those! I think this is fine then.

geograham:
Hey, there are a couple of expected type warnings to look at as well.

clmould:
> Hey, there are a couple of expected type warnings to look at as well.

I have fixed one type error. i have checked the other regarding make_polygon, it's in this format in other locations in the code and is working fine

clmould:
> Can you fix the new warnings in ^this comment?

There were a few `nan` occurring in a test, I have fixed this now and the warning has gone

je-cook (approved):
Great thanks

PR #4154, opened 2025-11-19
Add optional scipy backend to optimiser

Hudsonbaker4:
## Description

Add interfaces to SciPy SLSQP, COBYLA, and COBYQA optimisation algorithms. Scope for extension going forwards.

## Checklist


bluemira/equilibria/run.py
```diff
@@ -477,6 +483,8 @@ def run_reference_equilibrium(self):
             relaxation=self.eq_config.relaxation,
             fixed_coils=True,
             diagnostic_plotting=self.eq_config.diagnostic_plotting,
+            keep_history=True,
```
CoronelBuendia: No point keeping history here it seems, the problem is not accessible a posteriori

bluemira/optimisation/_scipy/parameters.py
```diff
@@ -0,0 +1,43 @@
...
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+from collections.abc import Mapping
+from dataclasses import asdict, dataclass
+
+
+@dataclass
+class COBYLAParams:
+    rhobeg: float = 0.2
+    catol: float = 1e-4
```
CoronelBuendia: Have you checked that these parameters are the same defaults that NLOpt uses in its implementation?

bluemira/optimisation/_algorithm.py
```diff
@@ -13,22 +13,44 @@
...
+    NEWTON_CG = auto()
+    L_BFGS_B = auto()
+    TNC = auto()
+    COBYLA_SCIPY = auto()
+    COBYQA = auto()
+    SLSQP_SCIPY = auto()
+    TRUST_CONSTR = auto()
+    DOGLEG = auto()
+    TRUST_NCG = auto()
+    TRUST_EXACT = auto()
```
CoronelBuendia: You're adding quite a few algorithms here, but they are not (IIRC) all able to handle the same kinds of problems. Specifically:
* bounds
* equality constraints
* inequality constraints

The ones that are there (I think) are able to handle all of the above... I may be mistaken.

bluemira/optimisation/_algorithm.py
```diff
@@ -13,22 +13,44 @@
...
+    TRUST_KRYLOV = auto()
+
+    def __new__(cls, value):
+        """Create Enum and debug attribute"""
+        obj = object.__new__(cls)
+        obj._value_ = value
+        return obj
```
CoronelBuendia, line 52: Don't we need to add default conditions later in this file for the new algorithms?

CoronelBuendia (changes requested):
Hi @Hudsonbaker4,

Thank you for this, it's very clearly written. I'm well aware that getting different scipy optimisation algorithms to play nice together is a pain, as they have completely different interfaces (although this seems to have gotten a bit easier in recent times). So; good job. I have a few niggles and one big one for you.

My main concern at the moment relates to the different capabilities of the optimisation algorithms. The ones we map to in NLOpt have "known" capabilities, (i.e. they can handle bounds, equalities, and inequalities). IIRC, scipy's COBYLA is not capable of handling bounds natively. I don't known about the other ones you have added.

I'm going to submit a PR to this, adding a complicated constrained optimisation test problem, so you can see what I'm talking about. I think it will be a good litmus test for these different algorithms, and may pick up if some simply cannot handle the 3 things we invariably need. We may need to add some mappings for these capabilities, either converting bounds to inequality constraints (yucky but it works) or forbidding the use of certain algorithms if they cannot handle one of {b, eq, ineq}.

.github/workflows/ci.yml
```diff
@@ -67,7 +67,7 @@ jobs:
         with-process: ["v3.2.2", "process_disabled"]
         include:
           - os: "ubuntu-latest"
-            python-version: "3.10"
+            python-version: "3.11"
             with-process: "process_disabled"
```
je-cook: Just remove this block if we're dropping 3.10

.github/workflows/examples.yml
```diff
@@ -43,7 +43,7 @@ jobs:
 
       - name: Run examples
         env:
-          RUN_DAGMC_EX: ${{ matrix.python-version == '3.10' && '-e design/optimised_reactor' || '' }}
```
je-cook, line 46: This is only needed because one of our dependencies doesnt work on 3.10 so we can remove this and its use

je-cook (commented):
Needs a rebase so the tests run and a few small changes to the CI

bluemira/optimisation/_scipy/optimiser.py
```diff
@@ -0,0 +1,376 @@
...
+            result = minimize(
+                fun=self.f_objective,
+                x0=x0,
+                args=(),
+                method=SCIPY_ALG_MAPPING[self.algorithm],
+                jac=self.df_objective if self.algorithm in DF_SUPPORTED else None,
+                hess=None,  # algorithms that use this are not yet implemented
+                constraints=self._eq_constraints + self._ineq_constraints,
+                bounds=Bounds(lb=self.lower_bounds, ub=self.upper_bounds),
+                tol=None,  # ignore - provide specific tolerance in opt_conditions
```
CoronelBuendia, line 312: This needs to be treated more consistently in our interface to scipy. I guess scipy does not allow tolerances on constraints, but we do. Perhaps we could take some norm of all the constraint tolerances. Min, or L-2, for example

bluemira/optimisation/_scipy/parameters.py
```diff
@@ -0,0 +1,37 @@
...
+# SPDX-FileCopyrightText: 2021-present J. Morris, D. Short
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+
+from collections.abc import Mapping
+from dataclasses import asdict, dataclass
+
+
+@dataclass
+class COBYLAParams:
```
CoronelBuendia, line 82: We need the param lists and defaults of the other algorithms, otherwise it's a painful exercise in reading scipy docs.

tests/optimisation/test_constrained_problem.py
```diff
@@ -0,0 +1,320 @@
...
+                "f_constraint": f_inequalities,
+                "df_constraint": df_inequalities,
+                "tolerance": 1e-6 * np.ones(8),
+            }
+        ],
+        bounds=(ALKYLATION_DATA.lower_bounds, ALKYLATION_DATA.upper_bounds),
+        algorithm=algorithm,
+        opt_conditions={"max_eval": 5000, "ftol_rel": 1e-9},
+    )
+
```
CoronelBuendia, line 323: Need to add a test for `constraints_satisfied=True`

bluemira/optimisation/_algorithm.py
```diff
@@ -13,22 +13,42 @@
...
+    NELDER_MEAD = auto()
+    POWELL = auto()
+    L_BFGS_B = auto()
+    TNC = auto()
+    # Bounds + ineq constraints
+    COBYLA_SCIPY = auto()
+    # Bounds + ineq + eq constraints
+    COBYQA = auto()
+    SLSQP_SCIPY = auto()
+    TRUST_CONSTR = auto()
```
CoronelBuendia, line 45: Params + tests for all of these?

CoronelBuendia (changes requested):
Sorry some brief comments which might need some work..! All in all, this is definitely a good thing to have, even if the scipy implementation of COBYLA is less nice, and COBYQA does not appear to do well on our test problem. The SLSQP implementation does slightly better on our test problem, however!

bluemira/optimisation/_scipy/optimiser.py
Hudsonbaker4: SciPy trust-constr, cobyla, and cobyqa allow tolerances on constraints as they use the "new" Linear/NonlinearConstraint classes to define constraints. SLSQP does not have any way to define per constraint tolerances as it only accepts the legacy constraint definition.

bluemira/optimisation/_scipy/parameters.py
Hudsonbaker4: I'm not sure what the best approach is here. From what I understand, SciPy generally sets sensible default values in the backend - I was planning to remove the defaults I set as I think scipy's are better.

pyproject.toml
```diff
@@ -82,7 +82,7 @@ pinned = [
     "numba-scipy @ git+https://github.com/numba/numba-scipy@23c3b33440ea1fe0f84d05d269fb4a3df4b92787",
     "numpy==1.26.4",
     "matplotlib==3.10.8",
-    "scipy==1.15.3",
+    "scipy>=1.16.0",
```
je-cook: When we get towards completion can we put this as `==` some version (leave the one in dependencies as that is now our minimum version)

bluemira/optimisation/_optimise.py
```diff
@@ -332,11 +354,12 @@ def _check_constraint(
         :
             the items in the constraint vector that violate the condition.
         """
+        epsilon = 1e-15
         c_value = constraint["f_constraint"](x_star)
         # Deal with scalar constraints
         c_value = np.array([c_value]) if np.isscalar(c_value) else c_value
         tols = np.array(constraint["tolerance"])
-        indices = np.nonzero(condition(c_value, tols))[0]
+        indices = np.nonzero(condition(c_value, tols + epsilon))[0]
```
je-cook: A couple of questions:

1) why are we adding an epsilon here is that not covered by tols?
2) if we do need epsilon what happens if its true for tols - epsilon?

bluemira/equilibria/run.py
je-cook:
```suggestion
```

bluemira/optimisation/_optimise.py
```diff
@@ -250,13 +254,31 @@ def _make_optimiser(
...
     -------
     :
-        a new optimiser object.
+        Configured optimiser.
     """
-    opt = NloptOptimiser(
+    if (alg := Algorithm(algorithm)) in NLOPT_ALG_MAPPING:
+        optimiser = NloptOptimiser
+    elif alg in SCIPY_REGISTRY:
+        optimiser = ScipyOptimiser
```
je-cook, line 272: Right now keep_history doesnt do anything for scipy. Thats fine I'm not going to hold this up for that!
```suggestion
        optimiser = ScipyOptimiser
        if keep_history:
            bluemira_warn("History storage is not implemented for scipy optimisers")
```

je-cook (commented):
This looks really good and seems to function well theres a few small things I've commented. I've been playing with it using the below modified snippet from the rosenbrook example.
There are some runtime warning if you leave the df_objective on (because it doesnt work). I'm in two mind but it would be nice to catch the runtime warnings on nelder_mead, powell and cobyla or avoid passing in the df_objective early and pipe a warning to the user using our logging.

Also you need a rebase for the CI to run.

```python
import time

import numpy as np

from bluemira.optimisation import optimise, Algorithm

from pprint import pprint


def f_rosenbrock(x, a, b):
    """The Rosenbrock function."""
    return (a - x[0]) ** 2 + b * (x[1] - x[0] ** 2) ** 2


def df_rosenbrock(x, a, b):
    """Gradient of the Rosenbrock function."""
    grad = np.zeros(2)
    grad[0] = -2 * a + 4 * b * x[0] ** 3 - 4 * b * x[0] * x[1] + 2 * x[0]
    grad[1] = 2 * b * (x[1] - x[0] ** 2)
    return grad


a = 1
b = 100

for algorithm in range(1, 18):
    t1 = time.time()
    try:
        result = optimise(
        lambda x: f_rosenbrock(x, a, b),
        x0=np.array([0.5, -0.5]),
        algorithm=algorithm,
        df_objective=lambda x: df_rosenbrock(x, a, b),
        opt_conditions={"ftol_rel": 1e-12, "ftol_abs": 1e-12},
        keep_history=True,
        
        bounds=([-2, -2], [3, 3]),
        )
    except Exception as e:
        print('BAD', e)
    t2 = time.time()
    print(f"{Algorithm(algorithm)}: {result}, time={t2 - t1:.3f} seconds")
```

tests/optimisation/test_constrained_problem.py
```diff
@@ -0,0 +1,325 @@
...
+    "algorithm",
+    [
+        Algorithm.SLSQP,
+        Algorithm.SLSQP_SCIPY,
+        Algorithm.COBYLA,
+        Algorithm.COBYLA_SCIPY,
+        pytest.param(
+            Algorithm.COBYQA,
+            marks=pytest.mark.xfail(reason="poor handling of initial conditions"),
+        ),
```
je-cook: I think we are at the point where we xfail slsqp_scipy and cobyla_scipy. Just make sure to put the reasons

je-cook (commented):
Ok I think we're nearly there from my side. Can you resolve the previous comments that you've dealt with?

tests/optimisation/test_constrained_problem.py
Hudsonbaker4: Fixed slsqp so that it now passes. Not sure why cobyla isn't - it passes locally...

je-cook: They both (slsqp and cobyla) fail for me locally. I reckon its similar to https://github.com/ukaea/PROCESS/issues/4000. We could add the hilbert matrix check the Tim did or just xfail it for now up to you if you need to find a resolution to it.

For reference here are my failures:

SLSQP
```
E       bluemira.optimisation.error.OptimisationError: Positive directional derivative for linesearch      message: Positive
E       directional derivative for linesearch      success: False
E       status: 8          fun: -1768.8069649665013            x: [ 1.698e+03
E       1.582e+04  5.410e+01  3.031e+03  2.000e+03                 9.012e+01
E       9.500e+01  1.049e+01  1.562e+00  1.535e+02]          nit: 32
E       jac: [ 5.040e+00  3.500e-02  1.000e+01 -5.985e+00  3.360e+00
E       0.000e+00 -1.910e+02  0.000e+00  0.000e+00  0.000e+00]         nfev:
E       28         njev: 28  multipliers: [-4.209e+00  7.462e+01  5.943e+01
E       6.778e-01  0.000e+00                 2.296e+02  0.000e+00  0.000e+00
E       3.118e+02  6.992e+01                 0.000e+00]
```

COBYLA
```
E       bluemira.optimisation.error.OptimisationError: Did not converge to a solution satisfying the constraints. See `maxcv`
E       for the magnitude of the violation.  message: Did not converge to a
E       solution satisfying the constraints. See `maxcv` for the magnitude of
E       the violation.  success: False   status: 0      fun:
E       -1768.8077139884026        x: [ 1.698e+03  1.582e+04  5.410e+01
E       3.031e+03  2.000e+03             9.012e+01  9.500e+01  1.049e+01
E       1.562e+00  1.535e+02]     nfev: 886    maxcv: 1.0279564319307403e-15
```

Hudsonbaker4: What version of scipy are you using? I can replicate those errors locally using v1.15.3, but if I update to >= v1.16.3 then both pass. I'm not sure what's happening in the pipeline though because after I pushed the fix for slsqp it passed in the pipeline but cobyla still failed?

je-cook: 1.17.0 locally, CI is that too, I've just checked

tests/optimisation/test_constrained_problem.py
```diff
@@ -0,0 +1,331 @@
...
+        Algorithm.SLSQP,
+        pytest.param(
+            Algorithm.SLSQP_SCIPY,
+            marks=pytest.mark.xfail(reason="numerical instability between environments"),
+        ),
+        Algorithm.COBYLA,
+        pytest.param(
+            Algorithm.COBYLA_SCIPY,
+            marks=pytest.mark.xfail(reason="numerical instability between environments"),
+        ),
```
CoronelBuendia, line 282: @Hudsonbaker4 I thought scipy's SLSQP did OK on this when we first looked at it... what happened? I've just checked on my machine, and it indeed fails, whereas I could have sworn it did not previously.

What do you mean by "numerical instability between environments"?

CoronelBuendia: OK, I understand now that this is because this test fails on some environments. Upsetting... but by no means the first of its kind.

CoronelBuendia (approved):

je-cook (approved):

PR #4166, opened 2025-12-01
✨ Allow for offset fallback to be disabled

### LINKED ISSUE #3989: Add flag to prevent discretised offset

As a reactor modeller, I want the build of my reactor to be halted in some cases when performing a wire offset fails. 

Having the fallback to a discretised offset is often useful, but in some cases, it's important to subsequent build steps that an offset produces a spline. The `offset_wire` function does produce a warning, but it's all too easy to miss (or ignore) this.

### Proposed solution

Add a flag to `offset_wire` that prevents falling back to a discretised offset and instead raises a `GeometryError`.

### Alternative solutions

_No response_

### Additional Context

_No response_

je-cook:
## Linked Issues

Closes #3989 

## Checklist

### DIFF
--- a/bluemira/geometry/tools.py
+++ b/bluemira/geometry/tools.py
@@ -237,6 +237,8 @@ def fallback_to(fallback_func, exception):
             try:
                 return func(*args, **kwargs)
             except exception:
+                if not kwargs.get("allow_fallback", True):
+                    raise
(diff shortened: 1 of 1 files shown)


hsaunders1904 (approved):
Thanks for this! Should be useful for us

PR #4170, opened 2025-12-09
MagneticConstraintSet - remove and avoid a couple of pertinent deepcopies

geograham:
## Description

Constraints/Targets not updating - deepcopy in dummy eq set up, and if deepcopy of eq used between setting up optimisation problem and the Picard iterator resulted in disconnection.

## Interface Changes

PicardIterator no longer takes eq as input. Does not need to, as we have the COP (with relevant eq) as input.

## Checklist

### DIFF
--- a/bluemira/equilibria/optimisation/constraints.py
+++ b/bluemira/equilibria/optimisation/constraints.py
@@ -11,7 +11,6 @@ Equilibrium optimisation constraint classes
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from copy import deepcopy
 from typing import TYPE_CHECKING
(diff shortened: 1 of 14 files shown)


CoronelBuendia (changes requested):
So, this is API-breaking (not sure how @je-cook wants to play it), and there are presently a couple of places in eudemo where the inputs to `PicardIterator` are not updated (and should be as part of this PR).

Perhaps this should be scheduled for the next release, with e.g. the STEP people being notified beforehand?

EDIT: Aside from that looks fine

CoronelBuendia (approved):

PR #4172, opened 2025-12-10
More generic x-point check

geograham:
## Linked Issues

Closes #4161

Note for reviewer: I had thought this might be used in which_legs of Legflux but I am not sure. 

Image below, to show rough use - checking for intersections between the closed flux surface and a tangent line at the x-point.

image

## Checklist

I confirm that I have completed the following checks:

- [x] Tests run locally and pass `pytest tests --reactor`
- [ ] Code quality checks run locally and pass `pre-commit run --from-ref develop --to-ref HEAD`
- [ ] Documentation built locally and checked `sphinx-build -W documentation/source documentation/build`

Issue #4161, opened 2025-11-27
Generalise flux surface functions that rely on x-point z checks


geograham:
### Description of issue / requirement to address

Currently there are a few places in the code where we compare the max/min of a closed flux surface to an x-point z-location. We could make use of o-to x-point vectors and their tangents, for an approach that is more generally applicable.

je-cook:
Currently slightly broken

je-cook:
closed by #4176

PR #4176, opened 2025-12-11
Correct issue 4161

geograham:
## Linked Issues
#4161
## Description

Return a count to use in x point check. Tests added. Linked to PR #4172.

## Checklist


bluemira/geometry/coordinates.py
```diff
@@ -2072,7 +2072,7 @@ def _intersect_count(xz_inter: np.ndarray, xz_2: np.ndarray) -> np.ndarray:
 
 def join_intersect(
     tgt_poly: Coordinates, ref_poly: Coordinates, *, get_arg: bool = False
-) -> list[int] | None:
+) -> tuple[Coordinates, list[int] | None, int | None]:
```
je-cook:
```suggestion
) -> tuple[Coordinates, list[int] | None, int | None] | Coordinates:
```

bluemira/equilibria/find.py
```diff
@@ -898,8 +898,8 @@ def x_point_check(flux_surface: Coordinates, op: Opoint, xp: Xpoint):
     """
     length = np.hypot(np.max(flux_surface.x), np.max(np.abs(flux_surface.z)))
```
CoronelBuendia: I'm not sure this length is really relevant. In fact, the longer the length is, the higher the possibility of a false positive (though, unlikely, given that the closed flux surfaces are usually going to be ellipse-shaped).

Consider this, albeit contrived, example:

image

A better length might be something like 3 times the grid cell hypotenuse, or just some small-ish distance (e.g. 0.5 m). Both of which contain magic numbers, I know... Or if you want to scale it to the plasma size, then maybe something like 0.2 * the distance between the O and X points (rather than this odd plasma measure, which I imagine will be quite large in most cases ~ 8-15 m?).

bluemira/equilibria/find.py
```diff
@@ -898,8 +898,8 @@ def x_point_check(flux_surface: Coordinates, op: Opoint, xp: Xpoint):
...
     tanget_line = two_point_angled_line(op, xp, length)  # default theta is tangent
-    _, arg_inters = join_intersect(flux_surface, tanget_line, get_arg=True)
-    return arg_inters.sort()
+    _, _, intersection_count = join_intersect(flux_surface, tanget_line, get_arg=True)
```
CoronelBuendia: Surely only `get_intersect` is needed here? You're throwing away the other results from `join_intersect`. This function is going to get called a lot and speed is of the essence.

bluemira/geometry/coordinates.py
```diff
@@ -2118,7 +2122,7 @@ def join_intersect(
 
     if get_arg:
         insertion_locations = [dest_poly.argmin([x, 0, z]) for x, z in xz_inter]
-        return dest_poly, list(set(insertion_locations))
+        return dest_poly, list(set(insertion_locations)), len(insertion_locations)
```
CoronelBuendia: I don't think this additional return argument is actually helping anything here. You don't need to use `join_intersect` in the x point check, and something like `len(get_intersect(shape1.xz, shape2.xz).T)` (unchecked) should be enough. Like that the other places `join_intersect` is used do not need to be modified either.

CoronelBuendia (changes requested):

bluemira/equilibria/find.py
CoronelBuendia: Also, it appears this PR has killed the separatrix legs in SN plots:

image

bluemira/equilibria/find.py
geograham: Yep, you are correct, I assume the length I chose was due to pre-Christmas brain, I have changed to a much smaller number that scales with plasma size

bluemira/equilibria/find.py
geograham: Agree and changed

bluemira/geometry/coordinates.py
geograham: Will change back

bluemira/geometry/coordinates.py
```diff
@@ -2072,7 +2072,7 @@ def _intersect_count(xz_inter: np.ndarray, xz_2: np.ndarray) -> np.ndarray:
 
 def join_intersect(
     tgt_poly: Coordinates, ref_poly: Coordinates, *, get_arg: bool = False
-) -> list[int] | None:
+) -> tuple[Coordinates, list[int] | None, int | None] | Coordinates:
```
CoronelBuendia:
```suggestion
) -> tuple[Coordinates, list[int] | None] | Coordinates:
```

bluemira/equilibria/find.py
```diff
@@ -890,16 +895,23 @@ def x_point_check(flux_surface: Coordinates, op: Opoint, xp: Xpoint):
...
     arg_inters:
         Intersection indices
     """
-    length = np.hypot(np.max(flux_surface.x), np.max(np.abs(flux_surface.z)))
-    tanget_line = two_point_angled_line(op, xp, length)  # default theta is tangent
-    _, arg_inters = join_intersect(flux_surface, tanget_line, get_arg=True)
-    return arg_inters.sort()
+    if tangent_length is None:
+        tangent_length = (
+            np.hypot(np.max(flux_surface.x), np.min(flux_surface.x)) * 1e-2
```
CoronelBuendia:
```suggestion
            np.max(flux_surface.x) - np.min(flux_surface.x) * 1e-2
```
Doesn't really make sense otherwise. Also doesn't really work as is on a MAST-U case.

CoronelBuendia (approved):

PR #4219, opened 2026-02-09
♻️ Improve commutative unit reconstruction

### LINKED ISSUE #4052: Allow conversion of complex units to their standard SI representation

Under our current unit system we heavily enforce strict SI with exceptions for some well know cases (power etc):
https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/develop/bluemira/base/constants.py#L268-L275

These only function for standalone units eg W not W/m^2.
It would be nice to do standard conversion to any combination of the above combined units

### Proposed solution

Modify these function:
https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/develop/bluemira/base/parameter_frame/_frame.py#L645-L685

to use the below access to the constituent units to reconstruct the unit as SI:

```python
import pint
u = pint.UnitRegistry()
q = 1 * u.watt/(u.meter**2)
for a in q.units._units.items():
     print(repr(u.Unit(a[0]).dimensionality), f'multiplier={a[1]}')
# <UnitsContainer({'[length]': 2, '[mass]': 1, '[time]': -3})> multiplier=1
# <UnitsContainer({'[length]': 1})> multiplier=-2
```

### Alternative solutions

Leave the current solution and we get:
```python
>>> q.to_base_units()
<Quantity(1.0, 'kilogram / second ** 3')>
```

### Additional Context

_No response_

je-cook:
## Linked Issues

Closes #4052 

## Description

This PR simplifies and refactors the unit reconstruction in parameterframe to more closely follow the original intent of the input units.

pint's `to_preferred` which works the same as our previous construction sometimes over collapses units as detailed in the issue this closes. I've worked around this by extracting our non commutative units and trying to reconstruct the remaining units into a more standard SI representation.

This needs to wait until pint has its next release, 0.25.3 (otherwise `to_preferred` can crash) and the #4154 as the newer versions of pint drop support for python3.10 (which that PR does).

To test you'll need to install pint from the repo. Probably a few more edge cases can be found to test too.

## Checklist

### DIFF
--- a/bluemira/base/parameter_frame/_units.py
+++ b/bluemira/base/parameter_frame/_units.py
@@ -142,6 +142,7 @@ def _remake_units(quantity: Quantity) -> pint.Quantity:
             # is not commutative or is angle (as they're dimensionless in pint)
             fix_list.append(no)
         else:
+            # commumtative parts
             ind_list.append(no)
         unit_list.append(ureg.Quantity(f"{q.magnitude}({q.units})^{multiplier}"))
 
@@ -153,7 +154,12 @@ def _remake_units(quantity: Quantity) -> pint.Quantity:
 def _convert_non_commutative(
     unit_list: list[Quantity], filter_index: list[int]
 ) -> Quantity:
-    """Converts angle units and combines non commutative units"""  # noqa: DOC201
+    """Converts angle units and combines non commutative units
+
+    Notes
+    -----
+    Commutative here means units that we dont want to combine eg time and full power year
+    """  # noqa: DOC201
 
     # There is only one unit in _units of 'i' therefore extract the key/value
     def get_key(i: Quantity) -> str:
@@ -170,6 +176,7 @@ def _convert_non_commutative(
 
             filtered_list[no] = i.to(ureg.Unit(f"{ANGLE}**{get_exp(i)}"))
 
+    # multiplies all quantities together
     return math.prod(filtered_list)
 
 
@@ -182,15 +189,19 @@ def _combine_commutative(unit_list: list[Quantity], filter_index: list[int]) -> 
     Uses pints milp optimisation which can over combine eg W/m^2 -> kg/s^3.
     The optimisation minimises combinations eg the "knapsack problem".
     This function prioritise shorter original over new which could be improved in future
+
+    Commutative here means units that we dont want to combine eg time and full power year
     """  # noqa: DOC201
     filtered_list = [unit_list[i] for i in filter_index]
 
     quantity = math.prod(filtered_list)
     if not isinstance(quantity, ureg.Quantity):
+        # is quantity now a number
         return quantity
 
     # Prefer the users SI converted unit
     # if the number of constituent units is <= the number in the new unit
+    # this one line does a lot for us in the reconstruction
     pq = quantity.to_preferred()
     if len(quantity._units.keys()) <= len(pq._units.keys()):
         return quantity
--- a/bluemira/base/parameter_frame/_units.py
+++ b/bluemira/base/parameter_frame/_units.py
@@ -1,0 +1,153 @@
+# SPDX-FileCopyrightText: 2021-present M. Coleman, J. Cook, F. Franza
+# SPDX-FileCopyrightText: 2021-present I.A. Maione, S. McIntosh
+# SPDX-FileCopyrightText: 2021-present J. Morris, D. Short
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+from __future__ import annotations
+
+import math
+from typing import TYPE_CHECKING
+
+import pint
+
+from bluemira.base.constants import ANGLE, raw_uc, ureg
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    from pint import Quantity
+
+    from bluemira.base.parameter_frame._parameter import ParamDictT
+
+
+def _validate_units(param_data: ParamDictT, value_type: Iterable[type]):
+    try:
+        quantity = ureg.Quantity(param_data["value"], param_data["unit"])
+    except ValueError:
+        try:
+            quantity = ureg.Quantity(f"{param_data['value']}*{param_data['unit']}")
+        except pint.errors.PintError as pe:
+            if param_data["value"] is None:
+                quantity = ureg.Quantity(
+                    1 if param_data["unit"] in {None, ""} else param_data["unit"]
+                )
+            else:
+                raise ValueError("Unit conversion failed") from pe
+        else:
+            param_data["value"] = quantity.magnitude
+        param_data["unit"] = quantity.units
+    except KeyError as ke:
+        raise ValueError("Parameters need a value and a unit") from ke
+    except TypeError:
+        if param_data["value"] is None:
+            # dummy for None values
+            quantity = ureg.Quantity(
+                1 if param_data["unit"] in {None, ""} else param_data["unit"]
+            )
+        elif isinstance(param_data["value"], bool | str):
+            param_data["unit"] = "dimensionless"
+            return param_data
+        else:
+            raise
+
+    return _ensure_SI_unit_system(quantity, param_data, value_type)
+
+
+def _ensure_SI_unit_system(
+    quantity: Quantity, param_data: ParamDictT, value_type: Iterable[type]
+) -> ParamDictT:
+
+    if quantity.units != ureg.dimensionless:
+        unit_q = _remake_units(quantity)
+    else:
+        unit_q = quantity
+
+    if unit_q.units != quantity.units and param_data["value"] is not None:
+        val = raw_uc(quantity.magnitude, quantity.units, unit_q.units)
+        if isinstance(param_data["value"], int) and int in value_type:
+            val = int(val)
+        param_data["value"] = val
+
+    param_data["unit"] = f"{unit_q.units:~P}"
+    return param_data
+
+
+def _remake_units(quantity: Quantity) -> pint.Quantity:
+    """
+    Reconstruct unit from its dimensionality.
+
+    Parameters
+    ----------
+    quantity:
+        The quantity to reconstruct
+
+    Returns
+    -------
+    :
+        The quantity in new units
+
+    Notes
+    -----
+    The quantity of the conversion is not important here.
+    We just want the custom SI version of the input
+    The value conversion is done later.
+    """
+    unit_list, ind_list, fix_list = [], [], []
+    for no, (unit, multiplier) in enumerate(quantity.units._units.items()):
+        q = ureg.Quantity(unit)
+        if q.units not in {ureg.fpy, ureg.dpa}:
+            q = q.to_preferred()  # individually preferred (SI)
+        if q.units in {ureg.fpy, ureg.dpa} or not q.dimensionality:
+            # is not commutative or is angle (as their dimensionless in pint)
+            fix_list.append(no)
+        else:
+            ind_list.append(no)
+        unit_list.append(ureg.Quantity(f"{q.magnitude}{q.units}^{multiplier}"))
+
+    return _combine_commutative(unit_list, ind_list) * _convert_non_commutative(
+        unit_list, fix_list
+    )
+
+
+def _convert_non_commutative(
+    unit_list: list[Quantity], filter_index: list[int]
+) -> Quantity:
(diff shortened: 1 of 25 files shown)


CoronelBuendia (approved):
Just approving for the minor changes in geometry, equilibria, etc.

tests/base/parameterframe/test_parameterframe.py
```diff
@@ -621,6 +634,15 @@ def test_wtf_units(self):
         assert frame.wtf3.value == pytest.approx(0.01388888)
         assert frame.wtf3.unit == "m⁵/deg/dpa/fpy/s"
 
+    def test_combined_units(self):
+        frame = CombinedFrame.from_dict(self.COMBINED_FRAME_DATA)
```
jonmaddock, line 648: Is it worth saying that part of the purpose of this test is to check that this new unit reconstruction works? i.e. what the real purpose of this test is?

bluemira/base/parameter_frame/_units.py
```diff
@@ -0,0 +1,153 @@
...
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    from pint import Quantity
+
+    from bluemira.base.parameter_frame._parameter import ParamDictT
+
+
+def _validate_units(param_data: ParamDictT, value_type: Iterable[type]):
```
jonmaddock: I would ask for docstrings here, but hey, it's your code!

bluemira/base/parameter_frame/_units.py
```diff
@@ -0,0 +1,153 @@
...
+    try:
+        quantity = ureg.Quantity(param_data["value"], param_data["unit"])
+    except ValueError:
+        try:
+            quantity = ureg.Quantity(f"{param_data['value']}*{param_data['unit']}")
```
jonmaddock, line 53: Does this actually ever make a difference over the previous try statement?

bluemira/base/parameter_frame/_units.py
```diff
@@ -0,0 +1,153 @@
...
+        except pint.errors.PintError as pe:
+            if param_data["value"] is None:
+                quantity = ureg.Quantity(
+                    1 if param_data["unit"] in {None, ""} else param_data["unit"]
+                )
+            else:
+                raise ValueError("Unit conversion failed") from pe
```
jonmaddock, line 61: Could be worth commenting what the various try/except cases are and when they might be hit: some obvious, some not.

bluemira/base/parameter_frame/_units.py
```diff
@@ -0,0 +1,153 @@
...
+        else:
+            param_data["value"] = quantity.magnitude
+        param_data["unit"] = quantity.units
+    except KeyError as ke:
+        raise ValueError("Parameters need a value and a unit") from ke
+    except TypeError:
+        if param_data["value"] is None:
```
jonmaddock, line 69: This looks similar to above: when would this be hit instead of `PintError`?

bluemira/base/parameter_frame/_units.py
```diff
@@ -0,0 +1,153 @@
...
+            # dummy for None values
+            quantity = ureg.Quantity(
+                1 if param_data["unit"] in {None, ""} else param_data["unit"]
+            )
+        elif isinstance(param_data["value"], bool | str):
+            param_data["unit"] = "dimensionless"
+            return param_data
```
jonmaddock, line 76: Why return here, instead of running `_ensure_SI_unit_system()`?

bluemira/base/parameter_frame/_units.py
```diff
@@ -0,0 +1,153 @@
...
+        else:
+            raise
+
+    return _ensure_SI_unit_system(quantity, param_data, value_type)
+
+
+def _ensure_SI_unit_system(
+    quantity: Quantity, param_data: ParamDictT, value_type: Iterable[type]
+) -> ParamDictT:
```
jonmaddock, line 85: Docstring please.

bluemira/base/parameter_frame/_units.py
```diff
@@ -0,0 +1,153 @@
...
+    if quantity.units != ureg.dimensionless:
+        unit_q = _remake_units(quantity)
+    else:
+        unit_q = quantity
+
+    if unit_q.units != quantity.units and param_data["value"] is not None:
+        val = raw_uc(quantity.magnitude, quantity.units, unit_q.units)
+        if isinstance(param_data["value"], int) and int in value_type:
+            val = int(val)
+        param_data["value"] = val
```
jonmaddock, line 110: Don't understand what this is doing.

bluemira/base/parameter_frame/_units.py
```diff
@@ -0,0 +1,153 @@
...
+
+    param_data["unit"] = f"{unit_q.units:~P}"
```
jonmaddock, line 112: ?

bluemira/base/parameter_frame/_units.py
```diff
@@ -0,0 +1,153 @@
...
+    The quantity of the conversion is not important here.
+    We just want the custom SI version of the input
+    The value conversion is done later.
+    """
+    unit_list, ind_list, fix_list = [], [], []
+    for no, (unit, multiplier) in enumerate(quantity.units._units.items()):
+        q = ureg.Quantity(unit)
+        if q.units not in {ureg.fpy, ureg.dpa}:
+            q = q.to_preferred()  # individually preferred (SI)
+        if q.units in {ureg.fpy, ureg.dpa} or not q.dimensionality:
```
jonmaddock, line 141: Combine with previous if statement?

bluemira/base/parameter_frame/_units.py
```diff
@@ -0,0 +1,153 @@
...
+            # is not commutative or is angle (as their dimensionless in pint)
+            fix_list.append(no)
+        else:
+            ind_list.append(no)
```
jonmaddock, line 146: What case is this for?

bluemira/base/parameter_frame/_units.py
```diff
@@ -0,0 +1,153 @@
...
+
+    filtered_list = [unit_list[i] for i in filter_index]
+    for no, i in enumerate(filtered_list):
+        if not i.dimensionality and get_key(i) != ureg.dpa:
+            if get_key(i) in {ureg.steradian, ureg.square_degree}:
+                raise NotImplementedError("Solid angle conversion not supported")
+
+            filtered_list[no] = i.to(ureg.Unit(f"{ANGLE}**{get_exp(i)}"))
+
+    return math.prod(filtered_list)
```
jonmaddock, line 180: ?

bluemira/base/parameter_frame/_units.py
```diff
@@ -0,0 +1,153 @@
...
+
+    return _combine_commutative(unit_list, ind_list) * _convert_non_commutative(
+        unit_list, fix_list
+    )
+
+
+def _convert_non_commutative(
+    unit_list: list[Quantity], filter_index: list[int]
+) -> Quantity:
+    """Converts angle units and combines non commutative units"""  # noqa: DOC201
```
jonmaddock: Better definition of commutative would help here I think, especially an example.

bluemira/base/parameter_frame/_units.py
```diff
@@ -0,0 +1,153 @@
...
+
+    Notes
+    -----
+    Uses pints milp optimisation which can over combine eg W/m^2 -> kg/s^3.
+    The optimisation minimises combinations eg the "knapsack problem".
+    This function prioritise shorter original over new which could be improved in future
+    """  # noqa: DOC201
+    filtered_list = [unit_list[i] for i in filter_index]
+
+    quantity = math.prod(filtered_list)
```
jonmaddock, line 197: Clearly I'm missing something...

bluemira/base/parameter_frame/_units.py
```diff
@@ -0,0 +1,153 @@
...
+    if not isinstance(quantity, ureg.Quantity):
+        return quantity
```
jonmaddock, line 200: When would this be the case? Hasn't this been checked already?

bluemira/base/parameter_frame/_units.py
```diff
@@ -0,0 +1,153 @@
...
+
+    pq = quantity.to_preferred()
+    if len(quantity._units.keys()) <= len(pq._units.keys()):
+        return quantity
```
jonmaddock, line 207: Wouldn't we always want preferred units?

bluemira/base/parameter_frame/_units.py
```diff
@@ -0,0 +1,153 @@
...
+    The optimisation minimises combinations eg the "knapsack problem".
+    This function prioritise shorter original over new which could be improved in future
+    """  # noqa: DOC201
+    filtered_list = [unit_list[i] for i in filter_index]
+
+    quantity = math.prod(filtered_list)
+    if not isinstance(quantity, ureg.Quantity):
+        return quantity
+
+    pq = quantity.to_preferred()
```
jonmaddock, line 205: Maybe highlight this: this one line does almost all of the work of unit conversion using pint: the rest is handling the stuff pint doesn't handle, right?

jonmaddock (changes requested):
No problem with code, just questions, some of the answers to which might be worth turning into comments. This is quite tricky to get right (as you've done) and to understand working around pint's limitations for edge cases: some good examples as comments might help remove ambiguity and work out what's being done at each step.

bluemira/base/parameter_frame/_units.py
je-cook: yes the first fails whereas the second works:
```python
from pint import Quantity

try:
    Quantity(1, '1e-19m^2')
except ValueError as v:
    print(v)

Quantity('1*1e-19m^2')
```

bluemira/base/parameter_frame/_units.py
je-cook: see additional comments

bluemira/base/parameter_frame/_units.py
je-cook: It shortcircuits a load of logic that will never be used. Whereas in one case where we collapse a unit to dimensionless the quantity may now be different and therefore modify the value in some instances

bluemira/base/parameter_frame/_units.py
je-cook: I've added a comment

bluemira/base/parameter_frame/_units.py
je-cook: this is a short string-ified representation of the unit, removes pint objects from the user

bluemira/base/parameter_frame/_units.py
je-cook: q has possibly been updated in the first if statement and we need to deal with also this needs running if q has angles in it irrelevant of the previous step

bluemira/base/parameter_frame/_units.py
je-cook: commutative parts of the unit which we just want to pass through

bluemira/base/parameter_frame/_units.py
je-cook: multiplies all quantities together in the list

bluemira/base/parameter_frame/_units.py
je-cook: see previous comment on the same code

bluemira/base/parameter_frame/_units.py
je-cook: this covers if the multiplication became a raw number

bluemira/base/parameter_frame/_units.py
je-cook: preffered units sometimes overcollapse the unit and in some cases the original unit was "better" eg W/m² can resolve to kg/s3

tests/base/parameterframe/test_parameterframe.py
je-cook: docstring updated

bluemira/base/parameter_frame/_units.py
je-cook: I've added some notes here that should help the definition

josieapeters:
LGTM! tests all work well for me

josieapeters (approved):

jonmaddock (approved):
I'm happy with these comments: they answer all my queries.

PR #4221, opened 2026-02-09
Bugfix: LCFS and separatrix finding for different equilibrium configurations

CoronelBuendia:
## Linked Issues

Closes #{ID}

## Checklist


bluemira/equilibria/find.py
```diff
@@ -819,23 +818,17 @@ def get_flux_loop(psi_norm):
...
 
     while delta > psi_n_tol:
-        middle = low + delta / 2
+        middle = low + 0.5 * delta
         flux_surface = get_flux_loop(middle)
 
-        intersection_count = x_point_check(flux_surface, primary_op, primary_xp)
-        if flux_surface.closed and intersection_count <= 1:
-            # Middle flux surface is still closed, shift search bounds
+        if flux_surface.closed and flux_surface.length < 1.1 * perimeter:
```
geograham, line 828: Think there should be a note about the number 1.1 choice, so future people know it was chosen.

geograham (changes requested):
I have tested with the STEP eq and is working as expected. However, when I experimented with larger grid size, I also find that the 'delta_start' value needs adjusting to avoid choosing a starting 'low' value (and therefore a perimeter value) that loops around the coils. I think it would be worth putting the 'psi_n_tol' and 'delta_start' into the eq set up params.

geograham:
> I have tested with the STEP eq and is working as expected. However, when I experimented with larger grid size, I also find that the 'delta_start' value needs adjusting to avoid choosing a starting 'low' value (and therefore a perimeter value) that loops around the coils. I think it would be worth putting the 'psi_n_tol' and 'delta_start' into the eq set up params.

Just double checked - and was plotting the wrong flux surface - defaults working fine :)

geograham (approved):

PR #4242, opened 2026-02-24
Updated template builder for process v330

### LINKED ISSUE #4240: Update template builder for PROCESS v3.3.0
### Describe the bug

PROCESS v3.3.0 requires equality constraints to be listed first, and neqns stated - otherwise PROCESS will fail to converge. 

Add a separate function for adding equality constraints, to make it more fail safe that neqns will be set automatically and equalities at the start.

run_process_example.ex.py will need to be updated to reflect this, and need to re-enable the process example (in examples.yml) to run in the CI again. 

### Steps to reproduce

_No response_

### Tracebacks

```bash

```

### Version

develop

### OS

### OS details

_No response_

josieapeters:
## Linked Issues

Closes #4240 

## Description

User now needs to state when adding a constraint whether they intend for it to be an equality constraint or not. (i.e. template_builder.add_constraint(constraint, equality=bool))

add_constraint (bluemira/codes/process/template_builder.py) now calculates neqns, and automatically sorts the equalities to be at the top of the list, which PROCESS requires.

This has been tested using examples/codes/run_process_example.ex.py, which finds a feasible solution. 

## Checklist

### DIFF
--- a/bluemira/codes/process/template_builder.py
+++ b/bluemira/codes/process/template_builder.py
@@ -123,7 +123,7 @@ class PROCESSTemplateBuilder:
             bluemira_warn(f"Over-writing model choice {model_choice}.")
         self._models[model_choice.switch_name] = model_choice
 
-    def add_constraint(self, constraint: Constraint, *, equality: bool= False):
+    def add_constraint(self, constraint: Constraint, *, equality: bool = False):
         """
         Add a constraint to the PROCESS run
 
--- a/bluemira/codes/process/template_builder.py
+++ b/bluemira/codes/process/template_builder.py
@@ -123,16 +123,19 @@ class PROCESSTemplateBuilder:
             bluemira_warn(f"Over-writing model choice {model_choice}.")
         self._models[model_choice.switch_name] = model_choice
 
-    def add_constraint(self, constraint: Constraint):
+    def add_constraint(self, constraint: Constraint, *, equality: bool):
         """
         Add a constraint to the PROCESS run
+
+        It must be stated whether the constraint is intended as an equality constraint
         """
         if constraint in self._constraints:
             bluemira_warn(
                 f"Constraint {constraint.name} is already in the constraint list."
             )
-        elif constraint.equality:
+        elif equality:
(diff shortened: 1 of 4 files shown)


bluemira/codes/process/template_builder.py
```diff
@@ -123,14 +123,19 @@ def set_model(self, model_choice: PROCESSModel):
             bluemira_warn(f"Over-writing model choice {model_choice}.")
         self._models[model_choice.switch_name] = model_choice
 
-    def add_constraint(self, constraint: Constraint):
+    def add_constraint(self, constraint: Constraint, *, equality: bool):
```
je-cook: I would default to inequality because thats often what is wanted. If a user wants equality they can state it explicitly
```suggestion
    def add_constraint(self, constraint: Constraint, *, equality: bool= False):
```

examples/codes/run_process_example.ex.py
```diff
@@ -106,32 +106,34 @@
...
-    Constraint.TF_DUMP_VOLTAGE_UPPER_LIMIT,
-    Constraint.TF_CURRENT_DENSITY_UPPER_LIMIT,
-    Constraint.TF_T_MARGIN_LOWER_LIMIT,
-    Constraint.CS_T_MARGIN_LOWER_LIMIT,
-    Constraint.CONFINEMENT_RATIO_LOWER_LIMIT,
-    Constraint.DUMP_TIME_LOWER_LIMIT,
-    Constraint.CS_STRESS_UPPER_LIMIT,
-    Constraint.DENSITY_PROFILE_CONSISTENCY,
-    Constraint.PSEPB_QAR_UPPER_LIMIT,
+for constraint, equality in (
```
je-cook, line 115: this is good for the example though, its good to be explicit here

je-cook (commented):
Looks great, just one minor

bluemira/codes/process/template_builder.py
josieapeters: for sure, excellent, thank you :)

je-cook (approved):

Issue #4260, opened 2026-03-19
CoilGroupPlotter incorrectly carries kwargs for colours


kj5248:
### Describe the bug

When passing kwargs into a CoilGroup.plot() they are incorrectly passed through the CoilGroupPlotter class. For example within plot_coil there is self._plot_coil where it calls color for edgecolor when it should be edgecolor so that when it looks for edgecolor it can find it. This also occurs for facecolor too. Means that when trying to pass colours into coil plotting you end up with black squares with a grey fill.

Fix would involve changing this:
```
self._plot_coil(
    x_b,
    z_b,
    ct,
    color=self.edgecolor,
    linewidth=self.linewidth,
    **kwargs,
)
```

<img width="1178" height="981" alt="Image" src="https://github.com/user-attachments/assets/e8414364-909d-496f-9951-49b5da5c0842" />

to this:
```
self._plot_coil(
    x_b,
    z_b,
    ct,
    edgecolor=self.edgecolor,
    linewidth=self.linewidth,
    facecolor=self.colors,
    **kwargs,
)
```

<img width="1178" height="981" alt="Image" src="https://github.com/user-attachments/assets/1481ca73-f782-481b-92d7-c4783f13e671" />

which means that these are correctly identified in _plot_coil here:
```
fcolor = kwargs.pop("facecolor", PLOT_DEFAULTS["coil"]["facecolor"][ctype.name])
color = kwargs.pop("edgecolor", PLOT_DEFAULTS["coil"]["edgecolor"])
linewidth = kwargs.pop("linewidth", PLOT_DEFAULTS["coil"]["linewidth"])
```
For reference the dicts used for kwargs input are:
```
kwargs1 = {"facecolor": "white", "edgecolor": "green"}
kwargs2 = {"facecolor": "white", "edgecolor": "red"}
```

Just need to check this fix would work for all cases and needs to be made for other uses of self._plot_coil
### Version

2.11.1

### OS

WSL

PR #4261, opened 2026-03-19
Addition of method to create coils from a wire for use in RZIp

### LINKED ISSUE #4254: Add methods to turn a shell into passive coils for use with RZIp implementation

A method that allows for the conversion of a shell (such as the VV) into a set of passive coils would be beneficial for the RZIp implementation and allow for creating a more accurate system in a flexible/integrated manner.

### Proposed solution

Simple initial implementation could be just in 2D, ie turn it from a shell into face. Then to reduce complexity can take a wire at the plasma facing side and convert that to a series of coils. 

Later this can be built upon to fully fill the face area with coils, however this will drastically increase the computational cost of calculating the inductance matrices so might not be wanted/feasible.

Could also consider options such as partial and full VV walls, as most of the stabilisation occurs within certain angles from the midplane and could be a solution to reducing the complexity.

kj5248:
## Linked Issues

Closes #4254 

## Description

- Added function to create a series of coils with a given thickness from a wire.
- Added custom discretisation function to handle the breaking down of the wire in a more ideal way to minimise gaps between square coils and remove overlapping.
- Added some simple tests to check how shape changes with discretisation and compare new discretisation to previous method for 3 different shapes.
- Had to create separate file for the functions due to circular imports and named temp until resolved.

## Still needed

- Add necessary changes for function to make use of materials
- Needs change to how CoilGroupPlotter handles kwargs (issue #4260)
- Test using these coils in RZIp implementation to see how handles inductance matrix calculations
- Consider addition of filling face rather than just following wire for coil creation
- Resolve circular imports so functions can be in tools within magnetostatics (or move to _tools in equilibria.coils)

## Checklist


bluemira/magnetostatics/temp.py
```diff
@@ -0,0 +1,129 @@
...
+    Returns
+    -------
+    float:
+        Length value for a wire discretisation step.
+    """
+    theta = np.arctan2(p2[0] - p1[0], p2[2] - p1[2])
+    return 0.5 * delta * ((np.sqrt(2) + 1) - (np.sqrt(2) - 1) * np.cos(4 * theta))
+
+
+def get_coil_points(wire: BluemiraWire, thickness: float) -> np.ndarray:
```
CoronelBuendia: make private

bluemira/magnetostatics/temp.py
```diff
@@ -0,0 +1,129 @@
...
+        p = point
+        if current_length + g_val < wire.length - thickness:
+            points = np.append(points, [p], axis=0)
+            current_length += g_val
+        else:
+            continue
+    return points
+
+
+def make_coils_from_wire(
```
CoronelBuendia: what about `make_coils_along_wire`?

bluemira/magnetostatics/temp.py
```diff
@@ -0,0 +1,129 @@
...
+    wire: BluemiraWire,
+    thickness: float,
+    material: Material | None = None,
```
CoronelBuendia: Hmm, I know I said this, but perhaps we should keep the interface even simpler. The dominant use case for this will be for passive coils, so perhaps we should just past in a resistivity (more on this later)

bluemira/magnetostatics/temp.py
```diff
@@ -0,0 +1,129 @@
...
+            n_turns=1,
+            discretisation=np.nan,
+        )
+        for point in coil_points
+    ]
+    cg = CoilGroup(*coils)
+    if material:
+        j_max = material.j_max  # how to get?
+        b_max = material.b_max  # how to get?
+        cg.assign_material(ctype=None, j_max=j_max, b_max=b_max)
```
CoronelBuendia: remove

bluemira/magnetostatics/temp.py
```diff
@@ -0,0 +1,129 @@
...
+    for _ in range(int(n_max)):
+        p2 = wire.value_at(distance=current_length + dl).T
+        g_val = _length_step(p, p2, thickness)
+        point = wire.value_at(distance=current_length + g_val).T
+        p = point
+        if current_length + g_val < wire.length - thickness:
+            points = np.append(points, [p], axis=0)
+            current_length += g_val
+        else:
+            continue
```
CoronelBuendia: This feels like it could be slow, but it works so no need to improve.

CoronelBuendia (changes requested):
Looks good and works out the box. Nice.

Please move it to equilibria/coils/_tools.py and make this branch go into the RZIp branch.

We need to have a chat about how to incorporate coil resistance in here which is R [Ohm] = rho [Ohm.m] * length [m] / area [m^2]. Need to check which units are actually required in RZIp.

I think we should just take the resistivity (rho) as an input for this function, rather than a Material. I know I said that originally, but really we should avoid too much pollution; we can always wrap the resistivity calc in a Designer when we actually use it.

bluemira/magnetostatics/temp.py
CoronelBuendia: We should also pass in some default coil creation args, in particular ctype and some naming prefix

CoronelBuendia (approved):

PR #4273, opened 2026-03-27
Pin github actions uses to specific hashes

josieapeters:
Added the specific hashes and commented which versions these refer to


je-cook (approved):

je-cook:
Oh looks like you've missed one or two

je-cook (approved):

PR #4297, opened 2026-04-15
Cadquery backend

s-oli:
## Description
Work-in-progress implementation of the CadQuery geometry backend for bluemira.

This draft PR tracks the ongoing migration from FreeCAD to CadQuery as an 
alternative geometry backend. The goal is to mirror the existing `_freecadapi.py` 
API surface in `_cadqueryapi.py` so that the backend can be selected at runtime 
via `BLUEMIRA_GEOMETRY_BACKEND`.

### Current state
- Core geometry primitives implemented in `_cadqueryapi.py`
- Backend switching infrastructure in place
- ~100 tests still failing due to API differences between FreeCAD and CadQuery

### Known gaps / TODO
- Several FreeCAD-specific operations not yet fully mapped to CadQuery equivalents
- Some geometry operations produce slightly different results due to kernel differences

## Interface Changes
None — backend selection is additive, existing FreeCAD path unchanged.

## Checklist


tests/geometry/test_data/test_circ.stp
```diff
@@ -1,8 +1,9 @@
 ISO-10303-21;
 HEADER;
-FILE_DESCRIPTION(('FreeCAD Model'),'2;1');
-FILE_NAME('Open CASCADE Shape Model','2023-09-04T10:11:59',('Bluemira'),
-  ('Bluemira'),'Open CASCADE STEP processor 7.6','FreeCAD','Unknown');
+FILE_DESCRIPTION(('Open CASCADE Model'),'2;1');
+FILE_NAME('Open CASCADE Shape Model','2026-04-09T18:07:50',('Author'),(
```
je-cook: Future nice to have is to encode bluemira somewhere in the metadata, not for this PR

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+    """
+    if not inner:
+        try:
+            builder = BRepBuilderAPI_MakeFace(outer.wrapped, False)
+            builder.Build()
+            if builder.IsDone():
+                return cq.Face(builder.Face())
+        except Exception:  # noqa: BLE001, S110
+            pass  # fall through to planar paths below
+
```
Hudsonbaker4, line 163:
```suggestion
    else:  # we have inner wire(s), check if they are valid
        o_bounds = np.array(bounding_box(outer))
        
        for hole_wire in inner:
            h_bounds = np.array(bounding_box(hole_wire))
            
            is_outside_mins = np.any(h_bounds[:3] < o_bounds[:3] - _OCC_DEFAULT_TOL)
            is_outside_maxs = np.any(h_bounds[3:] > o_bounds[3:] + _OCC_DEFAULT_TOL)
            
            if is_outside_mins or is_outside_maxs:
                raise GeometryError(
                    "Topological error: inner wire is located partially or completely outside "
                    "the bounds of the outer wire."

```

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+    edge and the start-tangent of the outgoing edge must agree (within a
+    dot-product tolerance). OCCT's ``BRepOffsetAPI_MakePipeShell`` does
+    not enforce this itself and will happily sweep along a kinked polyline,
+    producing a self-intersecting / kinked solid. FreeCAD raises on this
+    case, so we do too.
+    """
+    edges = ordered_edges(path)
+    if len(edges) < 2:  # noqa: PLR2004
+        return
+    for e_prev, e_next in pairwise(edges):
```
Hudsonbaker4: Pairwise doesn't compare the first and last elements, which is potentially a problem for closed wires. We should add an `isClosed` check, like in `freecadapi.py > _wire_edges_tangent`.

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+    profiles: Iterable[apiWire],
+    path: apiWire,
+    *,
+    solid: bool = True,
+    frenet: bool = True,
+    transition: int = 0,
+) -> apiShell | apiSolid:
+    """Sweep one or more *profiles* along *path*."""
+    if not isinstance(profiles, list):
+        profiles = list(profiles)
```
Hudsonbaker4: If profiles is a single apiWire object rather than a list, calling `list(profiles` will invoke the `__iter__` method, which has been overwritten to return an ordered list of the edges composing the wire. We should use `profiles = [profiles]` here instead to avoid this.

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+                [],
+                path,
+                makeSolid=solid,
+                isFrenet=frenet,
+                transitionMode=transition_mode,
+            )
+        else:
+            result = cq.Solid.sweep_multi(
+                profiles, path, makeSolid=solid, isFrenet=frenet
+            )
```
Hudsonbaker4: Should `sweep_multi` have `mode` passed as a parameter so that the user can specify transition modes like they can for a single profile?

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+        Wire to offset (must be planar and non-straight).
+    thickness:
+        Offset distance. Positive = outward, negative = inward.
+    join:
+        Corner style: "arc", "intersect", or "tangent".
+    open_wire:
+        For open wires, keep offset open (True) or close it (False).
+    """
+    if not thickness:
+        return wire
```
Hudsonbaker4: Should we return a copy of the wire here, rather than the original?

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+        result_wires = wire.offset2D(thickness, kind=kind)
+    except Exception as exc:
+        raise FreeCADError(
+            f"CadQuery was unable to make an offset of wire: {exc}"
+        ) from exc
+
+    if not result_wires:
+        raise FreeCADError("offset_wire: no result produced")
+
+    result = result_wires[0]
```
Hudsonbaker4, line 623: Is it safe to assume that the result of an offset, specifically an inward offset, will produce only one wire? If it's possible that inwardly offsetting could produce multiple shapes, for example inwardly offsetting a figure 8 to produce two separate circles, we should either return a compound object or warn the user we are discarding geometry.

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+    cos=0) while accepting smooth parametric paths whose junction tangents may
+    differ slightly due to numerical evaluation or moderate curvature changes.
+    """
+    try:
+        t1 = _vector_to_numpy(edge1.tangentAt(edge1.paramAt(1.0)))
+        t2 = _vector_to_numpy(edge2.tangentAt(edge2.paramAt(0.0)))
+        cos_angle = np.dot(t1, t2) / (np.linalg.norm(t1) * np.linalg.norm(t2) + EPS)
+        return bool(cos_angle > 1.0 - tol)
+    except Exception:  # noqa: BLE001
+        return True  # conservative: don't block sweep on uncertainty
```
Hudsonbaker4: Is this function used anywhere?

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+        t_after = wire.tangentAt(d_after, mode="length")
+        v1 = np.array([t_before.x, t_before.y, t_before.z])
+        v2 = np.array([t_after.x, t_after.y, t_after.z])
+        n1, n2 = np.linalg.norm(v1), np.linalg.norm(v2)
+        if n1 < EPS or n2 < EPS:
+            continue
+        cos_a = np.dot(v1, v2) / (n1 * n2)
+        angle = math.acos(max(-1.0, min(1.0, cos_a)))
+        if not np.isclose(angle, 0.0, rtol=0.0, atol=atol):
+            return False
```
Hudsonbaker4: Should we consider the case when the wire is closed and we have a final edge tangency pair to check?

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+    if isinstance(obj, cq.Face):
+        inner = obj.innerWires()
+        if inner:
+            outer_area = _occ_face_area(cq.Face.makeFromWires(obj.outerWire()).wrapped)
+            hole_area = sum(
+                _occ_face_area(cq.Face.makeFromWires(w).wrapped) for w in inner
+            )
+            return outer_area - hole_area
+        return obj.Area  # property (monkey-patched) → float
+    return obj.Area()  # method on Wire/Solid/Shell/Edge → float
```
Hudsonbaker4, line 733: Should this function be able to handle non-planar faces? IIRC, FreeCAD's .Area property works on non-planar faces, so we should use `_face_from_wires_tolerant` instead of `makeFromWires` as you correctly point out before that it cannot handle non-planar faces.

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+def is_closed(obj: apiShape) -> bool:
+    """True if the shape is closed.
+
+    For compounds (e.g. the result of boolean wire fuse), a compound is
+    considered closed if it contains exactly one wire and that wire is closed.
+    """
+    if isinstance(obj, cq.Compound):
+        wires = obj.Wires()
+        return len(wires) == 1 and wires[0].IsClosed()
+    return obj.IsClosed()
```
Hudsonbaker4, line 762: Not all `apiShape` instances have `.IsClosed()` defined (i.e., faces) so it will crash.

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+    reconstructs the sharing and tightens tolerances.
+
+    The CadQuery wrapper's ``.wrapped`` attribute is reassigned so callers that
+    already hold a reference to *shape* see the fixed geometry.
+    """
+    fixer = ShapeFix_Shape(shape.wrapped)
+    fixer.SetPrecision(precision)
+    fixer.SetMinTolerance(min_length)
+    fixer.Perform()
+    shape.wrapped = fixer.Shape()
```
Hudsonbaker4, line 786: Should we cast at the end? `shape.wrapped = cq.Shape.cast(fixer.Shape())` - otherwise it becomes a generic `topoDS_Shape` rather than the specific type it should be (e.g., `topoDS_Face), which will cause issues downstream.

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+        explorer = BRepTools_WireExplorer(obj.wrapped)
+        result = []
+        while explorer.More():
+            result.append(cq.Edge(explorer.Current()))
+            explorer.Next()
+        if result:
+            return result
+    except Exception:  # noqa: BLE001, S110
+        pass  # fall back to storage-order edges below
+    return obj.Edges()
```
Hudsonbaker4, line 822: The `Wire` class in CadQuery overwrites `__iter__` to yield a wire's edges in order. We could use this here and replace the logic in the `try` block simply with `list(obj)`.

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+    """Create a wire from a list of edges."""
+    return cq.Wire.assembleEdges(edge_list)
+
+
+def wire_from_wires(wire_list: list[apiWire]) -> apiWire:
+    """Create a single wire from a list of connected wires."""
+    if len(wire_list) == 1:
+        return wire_list[0]
+    result = cq.Wire.combine(wire_list)
+    return result[0] if isinstance(result, list) else result
```
Hudsonbaker4, line 873: We should probably warn the user if the wires they passed in were not joining as otherwise we are silently throwing away part of their geometry.

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+    dss.Perform()
+    dist = dss.Value()
+    vectors = []
+    for i in range(1, dss.NbSolution() + 1):
+        p1 = dss.PointOnShape1(i)
+        p2 = dss.PointOnShape2(i)
+        vectors.append((
+            np.array([p1.X(), p1.Y(), p1.Z()]),
+            np.array([p2.X(), p2.Y(), p2.Z()]),
+        ))
```
Hudsonbaker4, line 929: Could we wrap `p1` and `p2` as `cq.Vectors` and then use `_vector_to_numpy`? That way we get the nice correction if it's near the geometry tolerance.

bluemira/codes/_cadqueryapi/_display.py
```diff
@@ -0,0 +1,165 @@
...
+    _orig_collect_verts = _orig_cadapi.collect_verts_faces
+    _orig_collect_wires = _orig_cadapi.collect_wires
+
+    try:
+        _orig_cadapi.collect_verts_faces = collect_verts_faces
+        _orig_cadapi.collect_wires = collect_wires
+        ps_backend.show_cad(parts, part_options, labels, **kwargs)
+    finally:
+        _orig_cadapi.collect_verts_faces = _orig_collect_verts
+        _orig_cadapi.collect_wires = _orig_collect_wires
```
Hudsonbaker4, line 231: Can we remove the monkey patching here? The polyscope script calls `cadapi.collect_verts_face` or `collect_wires` so it will get to the correct backend without monkey patching. This also avoids importing FreeCAD when it isn't needed.

bluemira/codes/_cadqueryapi/_placement.py
```diff
@@ -0,0 +1,376 @@
...
+    """Shape-(3,) or (N, 3) array of coordinates; TypeError on wrong types.
+
+    Mirrors FreeCAD's ``@check_data_type`` contract used by
+    :func:`vector_to_numpy` / :func:`vertex_to_numpy`. *get_xyz* maps one
+    element to an object exposing ``.x .y .z``.
+    """
+    if isinstance(obj, types):
+        c = get_xyz(obj)
+        return np.array([c.x, c.y, c.z])
+    if isinstance(obj, (list, tuple)) and obj and all(isinstance(x, types) for x in obj):
```
Hudsonbaker4, line 275: We could remove the `and obj` so that we return an empty array if the user passes in an empty array, rather than raising an error. I believe FreeCAD will return an empty array rather than crash.

bluemira/codes/_cadqueryapi/_curves.py
```diff
@@ -0,0 +1,333 @@
...
+        surface = Geom_BSplineSurface(
+            tcol_poles,
+            _real_array(knot_vector_u),
+            _real_array(knot_vector_v),
+            _int_array(mults_u),
+            _int_array(mults_v),
+            degree_u,
+            degree_v,
+        )
+    return surface
```
Hudsonbaker4, line 224: This does not return the correct type. It returns a `Geom_BSplineSurface`, whereas FreeCAD returns an `apiShape`. This crashes downstream when `tools.py` tries to convert the `apiShape` into a specific Bluemira object (`BluemiraWire`, etc...).

bluemira/codes/_cadqueryapi/_curves.py
```diff
@@ -0,0 +1,333 @@
...
+        )
+        edge = cq.Edge(BRepBuilderAPI_MakeEdge(arc.Value()).Edge())
+    return cq.Wire.assembleEdges([edge])
+
+
+def make_circle_arc_3P(
+    p1,
+    p2,
+    p3,
+    axis=None,
```
Hudsonbaker4, line 281: Should we use the axis parameter, so that if the user passes in an axis, it is used as the normal?

bluemira/codes/_cadqueryapi/_curves.py
```diff
@@ -0,0 +1,333 @@
...
+        raise FreeCADError(str(e)) from e
+    return cq.Wire.assembleEdges([edge])
+
+
+def make_ellipse(
+    center: list = (0.0, 0.0, 0.0),
+    major_radius: float = 2.0,
+    minor_radius: float = 1.0,
+    major_axis: list = (1.0, 0.0, 0.0),
+    minor_axis: list = (0.0, 1.0, 0.0),
```
Hudsonbaker4, line 296:
```suggestion
    center: tuple = (0.0, 0.0, 0.0),
    major_radius: float = 2.0,
    minor_radius: float = 1.0,
    major_axis: tuple = (1.0, 0.0, 0.0),
    minor_axis: tuple = (0.0, 1.0, 0.0),
```

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+
+
+def wire_value_at(wire: apiWire, distance: float) -> np.ndarray:
+    """Return the point a given arc-length distance along the wire."""
+    total = wire.Length()
+    if distance <= 0.0:
+        return start_point(wire)
+    if distance >= total:
+        return end_point(wire)
+    return _vector_to_numpy(wire.positionAt(distance / total))
```
Hudsonbaker4, line 1219: FreeCAD also warns the user if they request a value < `start_point` or > `end_point`.

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+
+
+def mirror_shape(
+    shape: apiShape,
+    base,
+    direction,
+) -> apiShape:
+    """Return a mirrored copy of the shape about the plane defined by base+direction."""
+    pnt = gp_Pnt(*[float(x) for x in base])
+    drc = gp_Dir(*[float(x) for x in direction])
```
Hudsonbaker4, line 1353: Should we normalise the direction here before passing it to `gp_Dir` (as is done above in `rotate_shape`)?

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+        groups.append(current)
+
+    wires: list = []
+    for g in groups:
+        wires.extend(_assemble_wires_from_edges(g))
+    wires.sort(key=lambda w: w.Length(), reverse=True)
+    return wires
+
+
+def boolean_cut(shape: apiShape, tools: list, *, split: bool = True) -> list[apiShape]:
```
Hudsonbaker4, line 1595: Does the split parameter need to be applicable to faces and solids as well, or is only wires okay? Also, do we need to handle shells here explicitly? FreeCAD does.

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+    *,
+    solid: bool = False,
+    ruled: bool = False,
+    closed: bool = False,
+) -> apiShape:
+    """Loft through a sequence of profiles."""
+    result = cq.Solid.makeLoft(list(profiles), ruled=ruled)
+    if not solid:
+        return cq.Shell(result.Shells()[0].wrapped)
+    return result
```
Hudsonbaker4, line 1739: The closed parameter is not used, FreeCAD uses it so we probably should too. Also, if the user passes in non-closed profiles, then I believe cq.Solid.makeLoft will crash, whereas FreeCAD will create an open loft.

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+    pnt = gp_Pnt(*[float(x) for x in point])
+
+    if isinstance(shape, apiFace):
+        # Project point onto the face's surface, then classify in UV space
+        surf = BRep_Tool.Surface_s(shape.wrapped)
+        projector = GeomAPI_ProjectPointOnSurf(pnt, surf)
+        projector.Perform(pnt)
+        if projector.NbPoints() == 0:
+            return False
+        u, v = projector.LowerDistanceParameters()
```
Hudsonbaker4, line 1899: Do we need to account for 3D distance? For example, if we have a 2D plane lying on z=0 and a point at z=100, that point would be projected onto the plane and shown to be contained inside it, even though it lies 100 above the plane itself.

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+                    "checkRational": bsp.IsRational(),
+                }
+            })
+        elif ctype == GeomAbs_BezierCurve:
+            bez = adaptor.Bezier()
+            poles = [
+                [bez.Pole(i).X(), bez.Pole(i).Y(), bez.Pole(i).Z()]
+                for i in range(1, bez.NbPoles() + 1)
+            ]
+            serialised.append({"BezierCurve": {"Poles": poles}})
```
Hudsonbaker4, line 1969: The start and end points of splines are not saved to the deserialisation. This could cause problems when a spline is trimmed because I believe it only creates a view, it doesn't change the underlying maths for the spline construction, so serialising the trimmed spline, and then deserialising would result in the full spline being reconstructed rather than the trimmed version.

bluemira/codes/_cadqueryapi/_core.py
```diff
@@ -0,0 +1,2319 @@
...
+
+    # Open wire: remove the dummy closing edge from the result.
+    # The dummy edge goes from end_pt to start_pt and was not at any filleted corner,
+    # so it should appear unchanged in the result.
+    result_edges = ordered_edges(result_wire)
+    keep = []
+    for e in result_edges:
+        sp = _vector_to_numpy(e.startPoint())
+        ep = _vector_to_numpy(e.endPoint())
+        tol = radius + 1e-3
```
Hudsonbaker4, line 2144: Is `tol = radius + 1e-3` too large? Shouldn't it be a very small absolute tolerance to just check around the start and end points, otherwise we risk removing geometry that wasn't temporary?

Hudsonbaker4 (changes requested):
Hi @s-oli,

Thanks for the PR. This is a significant chunk of work! You have clearly put in a lot of effort to this and completed it to a high standard, so thank you very much.

I have left a few comments and suggestions below. Once these are addressed, we can get this merged in and then continue working together on any outstanding features and further testing.

Thanks again.

bluemira/codes/_cadqueryapi/_core.py
s-oli: yep, makes sense, I added your suggestion to the code

bluemira/codes/_cadqueryapi/_core.py
s-oli: Good catch, fixed — also checks (edges[-1], edges[0]) when is_closed(path), mirroring        
  _freecadapi._wire_edges_tangent.

bluemira/codes/_cadqueryapi/_core.py
s-oli: Good catch — fixed. Now branches on isinstance(profiles, apiWire) first to wrap a single wire, then falls back to list(profiles) for genuine iterables (tuple/generator), so cq.Wire's edge-iterating __iter__ doesn't get triggered.

bluemira/codes/_cadqueryapi/_core.py
s-oli: Did you mean transitionMode rather than mode? They're separate things in CadQuery — mode is the auxiliary direction (Vector/Wire/Edge for trihedron control), transitionMode is what FreeCAD's transition int maps to. I went with the latter: multi-profile now drives BRepOffsetAPI_MakePipeShell directly so the transition argument finally has effect there too.

bluemira/codes/_cadqueryapi/_core.py
s-oli: we should - FreeCAD does it, now we do it, too. return wire.copy(), and don't create side effects.

bluemira/codes/_cadqueryapi/_core.py
s-oli: Ok, yes, fixed. Now sorts by length and keeps the longest piece (typically the dominant intended boundary), warning the caller when more than one disjoint wire results so they don't silently lose geometry. Full multi-wire support would mean widening the return type to apiCompound, which would ripple through  downstream callers, but for now the warn-and-take-the-largest matches the implicit assumption every existing caller has.

bluemira/codes/_cadqueryapi/_core.py
s-oli: Nope, that's dead code. Removed it.

bluemira/codes/_cadqueryapi/_core.py
s-oli: Yes — fixed, and at the same time used the closed-seam refactor as a chance to fix a related bug I'd noticed earlier: the old wire.tangentAt(arc_length)-based loop smoothed across edge boundaries, so it returned True even for a square wire (4× 90° corners) that should obviously have failed. Replaced with per-edge BRepAdaptor_Curve.D1 via two shared helpers — _edge_pair_cos_angle and _edge_junction_pairs — that both _wire_edges_tangent and _check_path_tangent_continuity now 
drive off, so the closed-seam handling and the kink detection are consistent.

Hudsonbaker4 (approved):
Thanks for these changes. I am happy to merge now, then we can address any outstanding comments on the new feature branch.

bluemira/codes/_cadqueryapi/_core.py
s-oli: Fixed in 1982482f — switched to `_face_from_wires_tolerant` for the
outer/hole sub-faces. `cq.Face.makeFromWires` assumes planarity, which
broke hole subtraction on BSpline-derived Faces.

bluemira/codes/_cadqueryapi/_core.py
s-oli: Done in c3b08e3f. Dispatched per kind: 
Wire/Edge -> IsClosed; 
Solid -> True; 
Shell -> BRep_Tool::IsClosed; 
Face -> True;
Previously fell back to obj.IsClosed() and would have raised AttributeError on Solids.

bluemira/codes/_cadqueryapi/_core.py
s-oli: Fixed in 41c878fe — `cq.Shape.cast(fixer.Shape()).wrapped` so the
concrete cq subclass survives the in-place fix.

bluemira/codes/_cadqueryapi/_core.py
s-oli: Ok, done in 8a120cb1. `cq.Wire.__iter__` already yields
edges in connectivity order, so the WireExplorer + try/except fallback
was redundant.

bluemira/codes/_cadqueryapi/_core.py
s-oli: Done in d04e3fdb — `bluemira_warn` when `cq.Wire.combine` returns
multiple disjoint pieces, then return the longest.

bluemira/codes/_cadqueryapi/_core.py
s-oli: Fixed in 0ea8077e — points now go through `cq.Vector` + `_vector_to_numpy` so they pass through the same coincidence-tolerance snap as the rest of the API.

bluemira/codes/_cadqueryapi/_core.py
s-oli: Done in 15418dd1. Equality cases short-circuit with `math.isclose` (ruff RUF069 forbids bare float `==`), out-of-range cases warn and clamp. Mirrors FreeCAD's behaviour.

bluemira/codes/_cadqueryapi/_core.py
s-oli: Fixed in 64181865 — direction is normalised with numpy before `gp_Dir`. Lets callers pass any non-zero vector and rejects zero explicitly via `np.linalg.norm`.

bluemira/codes/_cadqueryapi/_core.py
s-oli: Shell case added in ee466c60 (Shell → `_collect_subshapes` + entry in `_TOPABS_FOR_KIND`).

For Solid/Face I'd push back on adding a separate splitter: OCC's raw `shape.cut(*tools)` already partitions correctly for those kinds - FreeCAD's `BOPTools.SplitAPI.slice` is defensive parity, not
functionally required. Wire-vs-wire is the one case where raw cut does nothing (no dimensional overlap), and that's already covered. 
Happy to add the splitter if a concrete failing case turns up.

bluemira/codes/_cadqueryapi/_core.py
s-oli: Done in caac0ec2 — drives `BRepOffsetAPI_ThruSections(solid, ruled)` directly. `cq.Solid.makeLoft` hard-codes IsSolid=True and has no closed-loft option, so we now thread `solid`, `ruled`, and `closed`
through (closed re-adds the first profile to cycle the loft).

bluemira/codes/_cadqueryapi/_core.py
s-oli: Good catch — fixed in 2eefb21d. `GeomAPI_ProjectPointOnSurf` returns the closest surface point even for far-away inputs, so a point above or below a planar face was classifying as inside on its in-plane
projection. Now bails out when projection distance exceeds `_OCC_DEFAULT_TOL`.

bluemira/codes/_cadqueryapi/_core.py
s-oli: Done in 62f8ceb6. Persist `FirstParameter` / `LastParameter` for both BSpline and Bezier in `serialise_shape`; 
on read-back they drive the `BRepBuilderAPI_MakeEdge(curve, u1, u2)` overload via new 
optional kwargs on `make_bspline` / `make_bezier`. This actually round-trips trimmed 
edges — FreeCAD serialises the parameters but doesn't use them on read-back, so this 
is a small improvement on parity.

bluemira/codes/_cadqueryapi/_core.py
s-oli: Fixed in 7fa62425 — tightened to `_OCC_DEFAULT_TOL`. The open-wire endpoints are in `skip_keys` (never filleted), so the dummy edge's endpoints survive unchanged and a tight absolute tolerance is enough. 
The previous `radius + 1e-3` also matched short real edges adjacent to the endpoints.

bluemira/codes/_cadqueryapi/_display.py
s-oli: Done in 9e09f374. The monkey-patch was a no-op once `_polyscope` started calling through the `cadapi` dispatcher; dropped it entirely (also avoids forcing a FreeCAD import when the cadquery backend is
selected).

bluemira/codes/_cadqueryapi/_placement.py
s-oli: Fixed in e7588a1e — dropped the `and obj` truthiness gate so empty list/tuple yields an empty (0, 3) array instead of falling into the TypeError branch.

bluemira/codes/_cadqueryapi/_curves.py
s-oli: Fixed in 77a3222a — wrapped the parametric surface into a `cq.Face` via `BRepBuilderAPI_MakeFace` so callers receive an apiShape (matches FreeCAD's `bsplinesurface.toShape()`).

bluemira/codes/_cadqueryapi/_curves.py
s-oli: Done in 78b17d88. When `axis` is given, rebuild the circle on (centre, radius, axis), map `p1`/`p3` to parameters via `ElCLib::Parameter`, and reassemble the arc with that range. `axis=None` keeps the existing fast path.

bluemira/codes/_cadqueryapi/_curves.py
s-oli: Fixed in 50992fa5 - `list` -> `tuple` on the default-arg annotations.

PR #4317, opened 2026-05-06
Use better caching of Green's functions

CoronelBuendia:
## Linked Issues

Closes #{ID}

## Checklist

### DIFF
--- a/bluemira/equilibria/equilibrium.py
+++ b/bluemira/equilibria/equilibrium.py
@@ -623,7 +623,6 @@ class CoilSetMHDState(MHDState):
             self._psi_cache_valid = True
         return self._psi_green_cache
 
-    # --- Bx ---
     @property
     def _bx_green(self):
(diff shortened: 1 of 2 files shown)


CoronelBuendia:
Note I haven't actually tested for speed... dB only used for offline stability analysis at present, so that's a no-brainer. Bx and Bz caches should theoretically only be used mid-solve when using a BetaLiIp profile, in other situations it should make things faster.

kj5248 (commented):
Seems like a good improvement and will hopefully cut down on the amount of time recalculation occurs. You have changed 3 of the 4 params to properties but just got rid of _db_green, is this because it isn't necessary for the single use case? If so I'll approve it as everything looks good. Missed the comment you left, so will approve it.

kj5248 (approved):

je-cook (approved):

PR #4324, opened 2026-05-11
Remove trivial TH sin mode and fix TH indexing

### LINKED ISSUE #4329: Update toroidal harmonics example notebook
Update the toroidal harmonic example notebook to include an example of using toroidal harmonic constraints to alter the divertor leg shape

### LINKED ISSUE #4501: Toroidal harmonics doesn't work with a max harmonic mode > n degrees of freedom
Running a TH approximation with max harmonic mode > n degrees of freedom fails:

Image

Image

Fix this and add a test to cover this case

clmould:
## Linked Issues

Closes #{ID}
The sin 0 mode is always zeros and so should never be selected. This PR fixes the selection of modes, and updates the tests accordingly.
Adds in an example notebook showing leg movement (Closes #4329) 
Closes #4501 - use max harmonic mode in `_approximation_from_psi_fitting` instead of number of DOF so you can select a higher max mode than DOF

## Checklist


bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
```diff
@@ -1008,7 +1008,7 @@ def toroidal_harmonic_approximation(
     n_degrees_of_freedom = _set_n_degrees_of_freedom(
         n_degrees_of_freedom, max_harmonic_mode, len(th_params.th_coil_names)
     )
-    collocation = collocation_points(eq.get_LCFS(), PointType.GRID_POINTS)
+    collocation = collocation_points(eq.get_LCFS(), PointType.RANDOM, n_points=11)
```
CoronelBuendia: Why this change? Underlying behaviour will change that is presumably unrelated to the remove of the trivial sin mode and the bugfix which this PR addresses.

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
```diff
@@ -1018,39 +1018,42 @@ def toroidal_harmonic_approximation(
...
-                mode_id,
-                max_harmonic_mode,
-                collocation_psi,
-                mask,
-                true_coilset_psi,
+        cos_m_chosen = mode_values[mode_id[mode_id <= max_harmonic_mode]]
+        sin_m_chosen = mode_values[mode_id[mode_id > max_harmonic_mode]]
+
+        # Discard sin mode 0 as this is always 0s
+        if 0 not in sin_m_chosen:
```
CoronelBuendia: This is a very odd way of doing things...

examples/equilibria/Toroidal_Harmonics_Optimisation_Set_Up.ex.py
```diff
@@ -117,109 +103,3 @@
 )
 ax.set_title("Comparison of bluemira coilset psi to TH approx.")
 plt.show()
-# %% [markdown]
```
CoronelBuendia, line 120: Why has this example been cut short? It now no longer does what it says in the filename... no?

tests/equilibria/test_harmonics.py
```diff
@@ -1215,7 +1236,11 @@ def test_toroidal_harmonic_approximation(self):
             plasma_mask=True,
         )
 
-        assert result.error < 10
+        # Check sin mode 0 not selected
+        assert 0 not in result.sin_m
+
+        # Check error is suitably small
+        assert result.error < 20
```
CoronelBuendia, line 1243: What is 20 here? Why is it bigger than in the previous test? I thought we were fitting better with the bugfix..?

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
```diff
@@ -1018,39 +1018,42 @@ def toroidal_harmonic_approximation(
         eq=eq, th_params=th_params, psi_norm=psi_norm, plasma_mask=plasma_mask
     )
 
-    dof_id = np.arange(0, 2 * max_harmonic_mode)
-    mode_values = np.tile(np.arange(max_harmonic_mode), 2)
+    dof_id = np.arange(0, 2 * (max_harmonic_mode + 1))
+    mode_values = np.tile(np.arange(max_harmonic_mode + 1), 2)
```
CoronelBuendia: WHy not just:

```python
dof_id = np.arange(0, 2 * (max_harmonic_mode + 1) - 1)
np.r_[np.arange(0, max_harmonic_mode+1), np.arange(1, max_harmonic_mode+1)]
```
so you can get rid of this odd if statement further down?

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
```diff
@@ -1094,14 +1102,14 @@ def _approximation_direct_from_currents(
...
-    th_params,
-    n_degrees_of_freedom,
-    collocation,
-    mode_id,
-    max_harmonic_mode,
-    collocation_psi,
-    mask,
-    true_coilset_psi,
+    th_params: ToroidalHarmonicsParams,
+    n_deg_of_freedom: int,
```
CoronelBuendia, line 1119: is this now not just `len(cos_m_chosen) + len(sin_m_chosen)`?

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
```diff
@@ -743,7 +743,7 @@ def _set_n_degrees_of_freedom(
         appropriate n_dof to return
     """
     if n_dof is None:
-        n_dof = min(max_n_dof, 2 * max_harmonic_mode)
+        n_dof = min(max_n_dof, 2 * max_harmonic_mode - 1)
```
CoronelBuendia: Here we have `2*max_harmonic_mode - 1` but later on we have `dof_id = .. 2 * (max_harmonic_mode + 1)`. Are you sure this is consistent?

If we say that max_harmonic_mode e.g. = 5, then we are searching [0, 1, 2, 3, 4, 5] which is len == 6. If we throw away sin = 0, then that leaves us with cos[0,1,2,3,4,5] and sin[1,2,3,4,5] which is len == 11. But here if I specify max_n_dof = 11, I will get n_dof = 9.

CoronelBuendia (changes requested):

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
clmould: We wanted to use fewer points because we found that the fitting was better for a more square matrix

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
clmould: That was supposed to be temporary, have fixed now, thanks for spotting

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
clmould: Using this now, thank you!

tests/equilibria/test_harmonics.py
clmould: This is due to the change to PointType.RANDOM, and this was choosing a bad selection of points. We tested this out and found on develop found that using RANDOM collocation points we get an error of 33.62147368986065, but on this sin-0 branch we get an error of 16.874106576868474, so the sin -0-removal is giving a better approximation. It might be that the collocation points are not well distributed in random sample for the seed we have chosen. We were using random samples to test using smaller numbers of collocation points. Have tested now with grid spacing using a smaller sample of grid points. If we use a more square harmonics matrix (similar number of coils and collocation points), we get a better fit to the LCFS. If we use regularly spaced points, then we get lower errors.
 
In summary: the sin fix improves the error but we should use a smaller grid of collocation points for the fitting.

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
clmould: Update: see comment below on test_harmonics file

examples/equilibria/Toroidal_Harmonics_Optimisation_Set_Up.ex.py
clmould: The example currently on main wasn't showing a useful optimisation. I've made an issue to work on this #4329 to include a leg movement, but didn't want to hold up this PR trying to get that to work.

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
clmould: Have fixed this in the new commit

bluemira/equilibria/optimisation/harmonics/harmonics_approx_functions.py
```diff
@@ -201,12 +202,74 @@ class Collocation:
...
+        "z",
+        "z_1d",
+        "z_max",
+        "z_min",
+    )
+
+    def __init__(
+        self, x_min: float, x_max: float, z_min: float, z_max: float, nx: int, nz: int
+    ):
+        if x_min == x_max or z_min == z_max:
```
je-cook: I think this should be an isclose or some less than a tolerance otherwise you might get some weirdness later on

bluemira/equilibria/optimisation/harmonics/harmonics_approx_functions.py
```diff
@@ -201,12 +202,74 @@ class Collocation:
     z: np.ndarray
 
 
+class CollocationGrid:
```
CoronelBuendia, line 205: There is a fair amount of duplicate code here with `Grid.__init__`, perhaps we could split that out into a function that both call?

bluemira/equilibria/optimisation/harmonics/harmonics_approx_functions.py
clmould: Changed in new commit

geograham:
Hi @CoronelBuendia, I have gone through the changes in person with Clair. 
Hi @clmould, I am ok for this to get merged in as soon as you are happy that the review comments have been addressed.

clmould:
Here is a comparison between our old indexing result and our new indexing result. Please note: both use the updated collocation GRID_POINTS contained in this PR to make the comparison fair.

| |Old index  new grid|New index, new grid |
| --- | --- | -- |
|cos modes | [0 1 2 3 4] |[0 1 2 3 4]|
| sin modes | [0] | [2] |
|error from approx | 11.81 |8.86 |
| fit metric | 0.02 | 0.01|
| img| image | image |


Error is the error of L2 norm when comparing approximated coilset psi to desired coilset psi.

Fit metric is a measure of how 'good' the approximation is: fit_metric_value = total area within one but not both FSs / (input FS area + approximation FS area). We have chosen to compare a flux surface with a psi norm of 0.95.

Note that the cos modes selected by our method are the same but now sin mode 2 is chosen rather than 0.

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
CoronelBuendia, line 819: I guess this is not particularly related to this PR, but this collocation stuff has crept in and I now no longer fully understand why this is not somehow part of `ToroidalHarmonicParams` where there seems to be a fair  amount of coordinate stuff going on too.

In this function for example, we are passing in `ToroidalHarmonicParams`, but also using `Collocation` and then doing two almost identical things ever so slightly differently if `Collocation` is specified. Should `Collocation` not just be part of `ToroidalHarmonicParams` if it is shadowing some of the variables.

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
```diff
@@ -1005,10 +1010,9 @@ def toroidal_harmonic_approximation(
         R_0, Z_0 = eq.effective_centre()
         th_params = toroidal_harmonic_grid_and_coil_setup(eq=eq, R_0=R_0, Z_0=Z_0)
 
-    n_degrees_of_freedom = _set_n_degrees_of_freedom(
-        n_degrees_of_freedom, max_harmonic_mode, len(th_params.th_coil_names)
+    collocation = collocation_points(
+        eq.get_LCFS(), PointType.GRID_POINTS, n_points=5, use_mask=True
     )
-    collocation = collocation_points(eq.get_LCFS(), PointType.GRID_POINTS)
```
CoronelBuendia, line 1016: So now collocation is hard-coding n_points explicitly. Again, should this not be part of the "setup" in `ToroidalHarmonicsParams`?

Again, not really part of this PR, but it's quite hard to untangle where the tweaking/input parameters are (psi_norm, n_points, PointType, Collocation, etc.), and then the resulting data (x, z, r, phi) are actually stashed. The user-facing API seems quite fragmented. An unassuming user who just wants to pick up TH and use them would struggle, as I do every time I drop out of the loop for a couple of months on this.

The example disappearing in this PR does not give me great visibility on the user-facing API, and lots of the underlying options seem to be preset such that a normal user is unaware of them. This is fine if there is only one algorithm, but if the user won't use them because they are not part of the user-facing API, why have them?

CoronelBuendia:
Hi @clmould, as discussed yesterday, please put back in a reasonable(-ish) working example of the use of TH constraints in the existing example.

CoronelBuendia (commented):
Hi @clmould @geograham, I've given this a look over and it's probably OK, but without a working optimisation example I can't really play around with it (. There are a couple of general comments in this PR that don't really relate to this PR per se, sorry!

CoronelBuendia:
Previous review comments to be addressed in https://github.com/Fusion-Power-Plant-Framework/bluemira/issues/4471

examples/equilibria/Toroidal_Harmonics_Optimisation.ex.py
```diff
@@ -0,0 +1,243 @@
...
+    constraints=[
+        th_constraint,
+    ],
+    gamma=1e-8,
+)
+
+program = PicardIterator(
+    th_move_legs,
+    fixed_coils=True,
+    convergence=DudsonConvergence(limit=5e-3),
```
CoronelBuendia: 1e-3 works for me too

examples/equilibria/Toroidal_Harmonics_Optimisation.ex.py
```diff
@@ -0,0 +1,243 @@
...
+
+# %%
+diag_ops = EqDiagnosticOptions(
+    psi_diff=PsiPlotType.PSI_ABS_DIFF,
+    split_psi_plots=EqSubplots.XZ_COMPONENT_PSI,
+)
+eq_analysis = EqAnalysis(
+    input_eq=th_move_legs_eq, diag_ops=diag_ops, reference_eq=ref_eq
+)
+eq_analysis.plot_compare_psi()
```
CoronelBuendia, line 257: plt.show() ?

CoronelBuendia (changes requested):
Thanks @clmould, some improvements to the example (e.g. showing plots progressively) and fixing the notebook so it passes the CI checks, but otherwise good. As discussed API and architecture to be addressed in separate issue.

examples/equilibria/Toroidal_Harmonics_Optimisation.ex.py
clmould: I had tested and also found this, but had left it as 5e-3 as this works for more point placements for the inner leg, but I will change it to 1e-3 since that works here, with an added note that if experimenting with the optimisation the user may need to change some settings in the optimiser or iterator

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
```diff
@@ -1005,10 +1010,13 @@ def toroidal_harmonic_approximation(
         R_0, Z_0 = eq.effective_centre()
         th_params = toroidal_harmonic_grid_and_coil_setup(eq=eq, R_0=R_0, Z_0=Z_0)
 
-    n_degrees_of_freedom = _set_n_degrees_of_freedom(
-        n_degrees_of_freedom, max_harmonic_mode, len(th_params.th_coil_names)
+    # TODO @clmould: this will be updated in #4471
+    collocation = collocation_points(
+        eq.get_LCFS(),
+        PointType.GRID_POINTS,
+        n_points=max_harmonic_mode,
```
CoronelBuendia: This should `n_degrees_of_freedom`

CoronelBuendia (changes requested):
Hi @clmould, thanks for this. Just one point where I think we should either stick to the hard-coded 5, or after discussion with @geograham , n_degrees_of_freedom

bluemira/equilibria/optimisation/harmonics/harmonics_approx_functions.py
CoronelBuendia: #4471

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
CoronelBuendia: #4471

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
CoronelBuendia: #4471

bluemira/equilibria/optimisation/harmonics/toroidal_harmonics_approx_functions.py
clmould:
```suggestion
        n_points=5,
```

will put back to being hard-coded as 5

clmould: Have changed back to 5 for this PR, but have a new issue and plan to experiment with this #4516

CoronelBuendia (approved):

PR #4326, opened 2026-05-14
Enable caching of greens numba functions

hsaunders1904:
## Description

Apologies for the out-of-the-blue PR. Hopefully someone's OK to review this. 

I've been using the `ChargedParticleSolver` a fair bit recently and was curious why 1) `Equilibrium.from_eqdsk()`, and 2) the `ChargedParticleSolver.analyse()` took so long to execute during debugging cycles. 

I profiled a crude benchmark and found that a large part of the time taken was compiling Numba functions.

<details><summary>Benchmark</summary>

```python
from pathlib import Path
from time import time

from bluemira.base.file import get_bluemira_path
from bluemira.equilibria import Equilibrium
from bluemira.geometry.coordinates import Coordinates
from bluemira.radiation_transport.advective_transport import ChargedParticleSolver

read_path = get_bluemira_path("equilibria", subfolder="data")
eq_name = "DN-DEMO_eqref.json"
eq_name = Path(read_path, eq_name)

start = time()
eq = Equilibrium.from_eqdsk(eq_name, from_cocos=3, qpsi_positive=False)
print(f"Equilibrium.from_eqdsk() took {time() - start} s")
# -> Equilibrium.from_eqdsk() took 9.195956468582153 s

params = {
    "P_sep_particle": 150,
    "f_p_sol_near": 0.50,
    "fw_lambda_q_near_omp": 0.01,
    "fw_lambda_q_far_omp": 0.05,
    "f_lfs_lower_target": 0.75,
    "f_hfs_lower_target": 0.25,
    "f_lfs_upper_target": 0,
    "f_hfs_upper_target": 0,
}
read_path = get_bluemira_path("radiation_transport/test_data", subfolder="tests")
fw_name = "DN_fw_shape.json"
fw_name = Path(read_path, fw_name)
fw_shape = Coordinates.from_json(str(fw_name))
solver = ChargedParticleSolver(params, eq, dx_mp=0.001)

start = time()
x, z, hf = solver.analyse(first_wall=fw_shape)  # 
print(f"solver.analyse() took {time() - start} s")
# -> solver.analyse() took 9.254453182220459 s
```

</details>

```
Equilibrium.from_eqdsk() took 9.195956468582153 s
solver.analyse() took 9.254453182220459 s
```

image

You can see in this profile that the script takes 26.8 seconds, over 16 seconds of which is spent in Numba's compiler lock. You can drill down into the profile and work out that a lot of the compilations are run during functions that are calling Green's functions. 

The Green's functions in the `greens.py` module that are compiled with Numba are not cached. The reason they can't be cached is because they're using `scipy` functions: `ellipk` and `ellipe`. If you try caching, you get a warning like:

```
.../greens.py:328: NumbaWarning: Cannot cache compiled function "greens_psi" as it uses dynamic globals (such as ctypes pointers and large global arrays)
```

I don't know if there's another way around this, but one solution is to just implement the elliptic functions ourself. So that's what I've done here. That allows us to cache the compiled Green's functions.

The first run of the benchmark script will obviously take a similar amount of time as before as it compiles and caches all the functions. But all subsequent runs are significantly faster:

```
Equilibrium.from_eqdsk() took 0.17005300521850586 s
solver.analyse() took 1.0648384094238281 s
```

## Interface Changes

None.

## Checklist

### DIFF
--- a/tests/magnetostatics/test_greens.py
+++ b/tests/magnetostatics/test_greens.py
@@ -4,9 +4,12 @@
 #
 # SPDX-License-Identifier: LGPL-2.1-or-later
 
+
 import numba as nb
 import numpy as np
 import pytest
+from scipy.special import ellipe as scipy_ellipe
+from scipy.special import ellipk as scipy_ellipk
 
 from bluemira.base.constants import EPS, MU_0_2PI, MU_0_4PI
 from bluemira.magnetostatics.greens import (
@@ -127,6 +130,52 @@ def test_greens_vs_greens_all():
     np.testing.assert_allclose(Bx, Bx2)
     np.testing.assert_allclose(Bz, Bz2)
     np.testing.assert_allclose(psi, psi2)
+
+
+ellip_params = pytest.mark.parametrize(
+    "m",
+    [
+        # nans
+        np.nan,
+        # in valid range [0, 1]
+        0.314,
+        np.linspace(0.05, 0.95, 21),
+        np.full((4, 3, 2), 0.42),
+        # bottom of valid range
+        0.0,
+        np.zeros((3,)),
+        np.zeros((1, 2)),
+        # top of valid range
+        1.0,
+        np.ones((3,)),
+        np.ones((1, 2)),
+        # positive and out of valid range
+        1.5,
+        np.full((3,), 1.5),
+        np.full((2, 3), 2.0),
+        # small negative
+        -0.314,
+        np.full((4, 3, 2), -0.42),
+        # large negative
+        -100.0,
+        np.full((3,), -100.0),
+        # approaching range bounds
+        np.array([1e-9, 1e-22]),
+        np.array([1.0 - 1e-9, 1 - 1e-16]),
+        # non-C-contiguous
+        np.asfortranarray(np.full((4, 3), 0.5)),
+    ],
+)
+
+
+@ellip_params
+def test_ellipe_nb_matches_scipy_implementation(m):
+    np.testing.assert_allclose(ellipe_nb(m), scipy_ellipe(m))
+
+
+@ellip_params
+def test_ellipk_nb_matches_scipy_implementation(m):
+    np.testing.assert_allclose(ellipk_nb(m), scipy_ellipk(m))
 
 
 class TestGreenFieldsRegression:
--- a/tests/magnetostatics/test_greens.py
+++ b/tests/magnetostatics/test_greens.py
@@ -137,9 +137,9 @@ ellip_params = pytest.mark.parametrize(
     [
         # nans
         np.nan,
-        # in valid range [0, 1]
+        # in valid range [0, 1)
         0.314,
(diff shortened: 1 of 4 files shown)


CoronelBuendia:
Thanks Harry! Something happened along the way with numba and nb_scipy and I'm not quite sure when, but I have also noticed a massive slowdown since a couple of years ago. I thought at one point we had them cached, but clearly no longer.

Happy to take a look - will perhaps give you a call next week

tests/magnetostatics/test_greens.py
```diff
@@ -129,6 +132,52 @@ def test_greens_vs_greens_all():
...
+@ellip_params
+def test_ellipe_nb_matches_scipy_implementation(m):
+    np.testing.assert_allclose(ellipe_nb(m), scipy_ellipe(m))
+
+
+@ellip_params
+def test_ellipk_nb_matches_scipy_implementation(m):
+    np.testing.assert_allclose(ellipk_nb(m), scipy_ellipk(m))
+
+
```
CoronelBuendia, line 182:
```suggestion
class TestEllipticalFunctionsRegression:
    rng = np.random.default_rng(846023420)
    fixtures = []  # noqa: RUF012
    for _ in range(5):  # Tested with 80000, no failures
        fixtures.append(  # noqa: PERF401
            [
            2.0 * rng.random(np.random.randint(1, 10)) - 1.0
            ]
        )

    @pytest.mark.parametrize(("m",), fixtures)
    def test_ellipe_nb(self, m):
        self.runner(ellipe_nb, scipy_ellipe, m)

    @pytest.mark.parametrize(("m",), fixtures)
    def test_ellipk_nb(self, m):
        self.runner(ellipk_nb, scipy_ellipk, m)
    
    def runner(self, new_ellip_func, old_ellip_func, m):
        new = new_ellip_func(m)
        old = old_ellip_func(m)
        np.testing.assert_allclose(1e7 * new, 1e7 * old, atol=EPS)
```

CoronelBuendia: This was for my sanity, more than anything else, don't think it hurts though... Thoughts?

hsaunders1904: Added :+1: 

I've also updated the other test cases I had to test over the full valid range

je-cook:
I think this might allow us to remove numba-scipy as a dependency, the elliptic integrals were the main use IIRC.

hsaunders1904:
> I think this might allow us to remove numba-scipy as a dependency, the elliptic integrals were the main use IIRC.

Sounds good, always nice to remove a dependency! Can I leave that to you guys for another PR?

je-cook:
> > I think this might allow us to remove numba-scipy as a dependency, the elliptic integrals were the main use IIRC.
> 
> Sounds good, always nice to remove a dependency! Can I leave that to you guys for another PR?

yeah thats fine!

CoronelBuendia (approved):

PR #4345, opened 2026-05-29
Miscellaneous changes from demo

je-cook:
## Description

Changes to bluemira from demo branch

## Checklist


bluemira/display/auto_config.py
```diff
@@ -114,7 +114,7 @@ def get_figure_scale_factor(figsize):
     return sf
 
 
-def plot_defaults(*, force=False):
+def plot_defaults(*, force=False, override_scale=None):
```
je-cook: needs typing

bluemira/display/plotter.py
```diff
@@ -716,7 +716,7 @@ def _create_plotters(self, comp: Component) -> Iterator[BasePlotter]:
                 if self.options._external_options:
                     options = comp.plot_options
                 else:
-                    options = self.options
+                    options = comp.plot_options
```
je-cook, line 722: this feels wrong

bluemira/equilibria/profiles.py
```diff
@@ -111,7 +111,7 @@ def doublepowerfunc(x: float, *args) -> float:
     return np.sign(f) * (np.abs(f)) ** n
 
 
-@nb.jit(cache=True, forceobj=True)
+# @nb.jit(cache=True, forceobj=True)
```
je-cook: why was this done?

je-cook (commented):
A few defaults changes to look out for and some bugs that we should try and fix properly

bluemira/display/plotter.py
CoronelBuendia: Yes, I remember this, and I commented at the time. This was one of the fixes that I had to do to get the plasma to be pink again :')

bluemira/geometry/tools.py
```diff
@@ -1210,7 +1210,7 @@ def sweep_shape(
     *,
     solid: bool = True,
     frenet: bool = True,
-    transition: SweepShapeTransition | int = SweepShapeTransition.DEFAULT,
+    transition: SweepShapeTransition | int = SweepShapeTransition.ROUND_CORNER,
```
CoronelBuendia: Should drop this I think. It's accessible, changing the default could mess with other people's stuff.

bluemira/display/plotter.py
je-cook: ok I think I've fixed it now

bluemira/equilibria/profiles.py
CoronelBuendia: No idea now... You can turn it back on.

je-cook:
```suggestion
@nb.jit(cache=True, forceobj=True)
```

CoronelBuendia (approved):

bluemira/display/plotter.py
```diff
@@ -892,6 +889,29 @@ def plot_3d(
...
+    """Wraps default plot setting
+
+    Yields
+    ------
+    plotter:
+        the plotting object
+    part:
+        the part to be plotted
+    """
+    _user_options = options is not None or kwargs
```
josieapeters, line 906: just double checking my understanding, so in the case if someone inputted options but no kwargs -> on line 911 plotter.options._user_options = true -> would mean don't use default plotter settings, and any option info would have been given in the line before with the plotter class?

josieapeters (approved):

bluemira/display/plotter.py
je-cook: essentially yes. This allows the user to override any plotting options stored on a given object (inc the defaults  or those already on a component)

PR #4354, opened 2026-06-03
Update for get workflow origin

### LINKED ISSUE #3604: Fixup PR comments
## Describe the bug

There are a few problems that have surfaced with our warnings comment CI.

1) for some reason the report.json doesnt seem to be generated anymore?
2) The action used is archived  https://github.com/potiuk/get-workflow-origin
    something here will be useful to get it going again https://stackoverflow.com/questions/59077079/how-to-get-pull-request-number-within-github-actions-workflow
3) the actual warning message will have the same number of warnings for new and existing even if there are no new errors.

- OS: 
- Version: 
- Branch:

josieapeters:
## Linked Issues

Closes #3604 

## Description

Old action we used is no longer maintained, this uses github-script and adds a small action to specify the same behaviour we have before. 

## Checklist

### DIFF
--- a/.github/workflows/actions/get-workflow-origin/action.yml
+++ b/.github/workflows/actions/get-workflow-origin/action.yml
@@ -1,0 +1,44 @@
+name: "Get workflow origin"
+description: "Get PR number and target branch from pull_request or workflow_run"
+
+outputs:
+  pr_number:
+    description: "Pull request number"
+  target_branch:
+    description: "Target (base) branch"
+
+runs:
+  using: "composite"
+  steps:
+    - id: compute
+      uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7
+      with:
+        script: |
+          const event = context.payload;
+
+          let pullRequestNumber = null;
+          let targetBranch = null;
+
+          if (event.pull_request) {
+            prNumber = event.pull_request.number;
+            targetBranch = event.pull_request.base.ref;
+          }
+          else if (event.workflow_run) {
+            const run_id = event.workflow_run.id;
+
+            // Find PR linked to the workflow run
+            const prs = await github.rest.actions.listWorkflowRunPullRequests({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              run_id
+            });
+
+            if (prs.data.length > 0) {
+              const pr = prs.data[0];
+              pullRequestNumber = pr.number;
+              targetBranch = pr.base.ref;
+            }
+          }
+
+          core.setOutput("pr_number", pullRequestNumber);
+          core.setOutput("target_branch", targetBranch);
--- a/.github/workflows/actions/get-workflow-origin/action.yml
+++ b/.github/workflows/actions/get-workflow-origin/action.yml
@@ -43,7 +43,7 @@ runs:
               let pr;
               if (num_prs > 1) {
                 core.warning('${num_prs} PRs found.');
-                core.info('Proceeding with first PR');
+                core.info('Proceeding with first PR: ${prs.data[0].number}');
               }
 
               if (num_prs > 0) {
--- a/.github/workflows/actions/get-workflow-origin/action.yml
+++ b/.github/workflows/actions/get-workflow-origin/action.yml
@@ -2,43 +2,59 @@ name: "Get workflow origin"
 description: "Get PR number and target branch from pull_request or workflow_run"
 
 outputs:
-  pr_number:
+  pullRequestNumber:
     description: "Pull request number"
-  target_branch:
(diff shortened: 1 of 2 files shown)


josieapeters:
ok, it is definitely fetching the PR number and target branch now (as were the things we got from the potiuk action before) . I turned off a bunch of if statements in comments.yml to see how it would progress, but the CI stuff doesn't run fully outside of bluemira proper. 

This was the latest run, : https://github.com/josieapeters/bluemira-test-ci/actions/runs/27358516333/job/80839266342

je-cook (approved):
Ok we will merge and see after the below reverts (if @clmould is happy)

.github/workflows/comments.yml
```diff
@@ -72,14 +69,14 @@ jobs:
...
-      - name: Find warning report comment
+      - name: Find warning report comment TEST !
         if: env.RESULT == 'success' && steps.workflow-run-info.outputs.pullRequestNumber != ''
         uses: peter-evans/find-comment@b30e6a3c0ed37e7c023ccd3f1db5c6c0b0c23aad # v4.0.0
         id: find-warning-report-comment
         with:
           issue-number: ${{ steps.workflow-run-info.outputs.pullRequestNumber }}
           comment-author: "github-actions[bot]"
-          body-includes: ⚠️ Warning Report
+          body-includes: ⚠️ Howdy pal, here's your warnings
```
je-cook: this needs reverted as does the TEST addition a few lines above

josieapeters: thank you! have done

.github/workflows/actions/get-workflow-origin/action.yml
```diff
@@ -0,0 +1,60 @@
...
+                commit_sha: head_sha,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+              });
+
+              core.info(`Found ${prs.data.length} PRs`);
+
+              let pr;
+              if (prs.data.length > 0) {
+                pr = prs.data[0];
```
clmould, line 50: Can we guarantee that the first one is the one we want?

josieapeters: Not sure. There's a print statement included above in case there are multiple PRs associated with the commit, so it would flag up at least, but not 100% definite with the workflow. 

There is an [example output](https://docs.github.com/en/rest/commits/commits?apiVersion=2026-03-10#list-pull-requests-associated-with-a-commit).

I thought about this [other action](
https://docs.github.com/en/rest/actions/workflow-runs?apiVersion=2026-03-10#get-a-workflow-run), it's about getting a workflow run and it does mention PRs but I can't see PR number references in the example.

Could work to combine them? But we have the head commit associated with the workflow from l32, so I think it should be alright.

These are all from [this page](https://octokit.github.io/rest.js/v22/), and the 'See also: GitHub Developer Guide documentation' opens up the extra info on the other two links.

josieapeters: will make the "number of PRs" message a bit more of a big warning! :)

josieapeters:
Added a warning for if there are more than one PRs associated with the head commit. 

Realised [the style isn't in keeping with the other action ymls](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands). Let me know if you'd like me to change it to match (apologies, this started as a vibe-coded project that got heavily edited!)

.github/workflows/actions/get-workflow-origin/action.yml
clmould: Have had a think and a chat and it seems very unlikely that more than 1 pr would be selected - since it relies on the head commit, we think this could only happen if there was a duplicate pr (which seems unlikely), but even in this case the original should've been closed in favour of opening the newer one so only 1 pr should be selected here anyway. Even if the old one wasn't closed by the time the new one opens, they should contain the same information (perhaps just rebased) so the warnings-comment would still be valid if it selected the old pr.
I think the warning message is good to flag up if this scenario occurs, and if it does we can investigate again, but hard to test this out until it is merged!

.github/workflows/actions/get-workflow-origin/action.yml
```diff
@@ -0,0 +1,64 @@
...
+                commit_sha: head_sha,
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+              });
+
+              let num_prs = prs.data.length;
+              let pr;
+              if (num_prs > 1) {
+                core.warning('${num_prs} PRs found.');
+                core.info('Proceeding with first PR');
```
clmould:
```suggestion
                core.info('Proceeding with first PR: ${prs.data[0].number}');
```

clmould (approved):

PR #4373, opened 2026-06-14
LAR improvements: miscellaneous

s-oli:
## Description

eudemo/reactor.py now runs all the way through with BLUEMIRA_GEOMETRY_BACKEND=cadquery (PROCESS 3.4.1). Almost all the crashes came from one root cause, not from CadQuery geometry being wrong — both backends use the same OCCT 7.8.1.

The root cause. The wall-silhouette optimiser ended up with a different wall on CadQuery — one that didn't enclose the X-point — and that broke everything after it. But the two walls were actually the same shape (bounding box matches to 1e-7). The difference was in how the keep-out-zone constraint reads the wall: it checks each point of the wall against the zone, and discretise() starts at the wire's first edge, which isn't the same edge on both backends. So each backend starts from a different point — same shape, just sampled in a different order around the loop. That shifted the numbers by up to 6.6 m per point and sent the optimiser down a different path. The fix samples the wall the same way every time (same start point, same direction, evenly spaced), so both backends see the same thing. The wall difference between backends drops from 5.3% to 0.012%.

What's in here
- Sample the keep-out-zone constraint the same way on both backends (the main fix).
- length() uses OCCT BRepGProp instead of the rougher cq.Edge.Length().
- When a hole cut leaves slivers, keep the main face and warn instead of erroring.
- Penalise the optimiser when it tries an invalid shape, instead of crashing.
- WallSilhouette now requires the wall to enclose the X-point.
- PF-coil split tolerance set to EPS_FREECAD.
- OIS designer copes with keep-out zones that don't touch each other.
- save_reactor no longer needs a neutronics config folder to exist.

Everything works the same on FreeCAD, and there are tests.

## Checklist

### DIFF
--- a/bluemira/geometry/optimisation/_tools.py
+++ b/bluemira/geometry/optimisation/_tools.py
@@ -7,6 +7,7 @@ from dataclasses import dataclass
 
 import numpy as np
 
+from bluemira.geometry.error import OutOfBoundsError
 from bluemira.geometry.optimisation.typed import (
     GeomClsOptimiserCallable,
     GeomConstraintT,
@@ -52,10 +53,20 @@ def to_objective(
     :
         The objective function converted from a geometry objective function.
     """
+    worst = [0.0]  # largest finite objective value seen so far
 
     def f(x):
         geom.variables.set_values_from_norm(x)
-        return geom_objective(geom)
+        try:
+            value = geom_objective(geom)
+        except OutOfBoundsError as exc:
+            # No constructable geometry here. Penalise above the worst feasible
+            # value seen, growing with how far out of bounds we are, so a
+            # gradient-based optimiser is pushed back to the feasible region
+            # instead of crashing on the geometry build.
+            return (abs(worst[0]) + 1.0) * (1.0 + exc.excess)
+        worst[0] = max(worst[0], value)
+        return value
 
     return f
 
@@ -73,10 +84,22 @@ def to_optimiser_callable(
     :
         The optimiser function converted from a geometry optimiser function.
     """
+    state = {"shape": None, "worst": 0.0}  # last output shape, largest magnitude
 
     def f(x):
         geom.variables.set_values_from_norm(x)
-        return geom_callable(geom)
+        try:
+            value = geom_callable(geom)
+        except OutOfBoundsError as exc:
+            # See ``to_objective``: penalise (a constraint reads as strongly
+            # violated) rather than crash on the geometry build.
+            penalty = (state["worst"] + 1.0) * (1.0 + exc.excess)
+            return penalty if not state["shape"] else np.full(state["shape"], penalty)
+        arr = np.asarray(value, dtype=float)
+        state["shape"] = arr.shape
+        if arr.size:
+            state["worst"] = max(state["worst"], float(np.max(np.abs(arr))))
+        return value
 
     return f
 
--- a/bluemira/geometry/optimisation/_tools.py
+++ b/bluemira/geometry/optimisation/_tools.py
@@ -164,9 +164,76 @@ def calculate_signed_distance(
     shape = parameterisation.create_shape()
     # Note that we do not discretise by edges here, as the number of
     # points must remain constant so the size of constraint vectors
-    # remain constant.
-    s = shape.discretise(n_shape_discr, byedges=False).xz
+    # remain constant. Sample densely, then resample to the requested count
+    # from a backend-independent anchor (see _canonical_loop_points).
+    dense = shape.discretise(max(8 * n_shape_discr, 200), byedges=False).xz
+    s = _canonical_loop_points(dense, n_shape_discr)
     return signed_distance_2D_polygon(s.T, zone_points.T).T
+
+
+def _canonical_loop_points(loop: np.ndarray, n: int) -> np.ndarray:
+    """Resample a closed loop to ``n`` points in a backend-independent way.
+
+    ``discretise`` starts at the wire's first edge and follows the wire's
+    stored direction, both of which depend on the backend-specific edge
+    ordering. For an otherwise identical loop this only phase-shifts (and can
+    reverse) the samples, yet it scrambles the per-element keep-out-zone
+    constraint vector between backends and sends the optimiser into a different
+    basin. Pinning the winding, anchoring at the outer-midplane point and
+    resampling at uniform arc length makes the constraint deterministic and
+    smooth in the design variables.
+
+    Parameters
+    ----------
+    loop:
+        ``(2, M)`` densely sampled closed loop in the xz plane.
+    n:
+        Number of points to return.
+
+    Returns
+    -------
+    :
+        ``(2, n)`` points at uniform arc length, measured from the anchor.
+    """
+    pts = loop.T
+    if np.allclose(pts[0], pts[-1]):
+        pts = pts[:-1]
+    # Consistent winding (counter-clockwise) via the shoelace signed area.
+    x, z = pts[:, 0], pts[:, 1]
+    if np.sum(x * np.roll(z, -1) - np.roll(x, -1) * z) < 0:
+        pts = pts[::-1]
+    # Cumulative arc length around the closed loop.
+    closed = np.vstack([pts, pts[:1]])
+    cum = np.concatenate([
+        [0.0],
+        np.cumsum(np.linalg.norm(np.diff(closed, axis=0), axis=1)),
+    ])
+    length = cum[-1]
+    # Anchor where the loop crosses the outboard horizontal ray from its
+    # centroid (the outer-midplane point). Unlike an axis-aligned or radial
+    # extremum, this crossing is unique, sits on the smooth outer wall and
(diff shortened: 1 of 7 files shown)


eudemo/eudemo/ivc/wall_silhouette.py
```diff
@@ -183,6 +185,20 @@ def f_objective(geom: GeometryParameterisation) -> float:
...
+        z_margin = 0.05  # the wall must clear the x-point by this much, in metres
+
+        def f_x_point(geom: GeometryParameterisation) -> np.ndarray:
+            # Inequality constraint, satisfied when <= 0: the wall's lowest point
+            # sits at least ``z_margin`` below the separatrix x-point, so the
+            # first wall encloses it. Enforced here rather than only validated
+            # after the fact (see ``execute``).
+            z_min = geom.create_shape().bounding_box.z_min
+            return np.array([z_min - (x_point_z - z_margin)])
+
```
je-cook, line 201: @CoronelBuendia I'd like your opinion here

je-cook (dismissed):
Phew that was heavy. Well done for finding all of that. I just want a second opinion from @CoronelBuendia on the point below but other good to go I think.

I will cherry pick to the require locations after we've merged this into the test branch

bluemira/geometry/optimisation/_tools.py
```diff
@@ -52,10 +53,20 @@ def to_objective(
...
+        try:
+            value = geom_objective(geom)
+        except OutOfBoundsError as exc:
+            # No constructable geometry here. Penalise above the worst feasible
+            # value seen, growing with how far out of bounds we are, so a
+            # gradient-based optimiser is pushed back to the feasible region
+            # instead of crashing on the geometry build.
+            return (abs(worst[0]) + 1.0) * (1.0 + exc.excess)
+        worst[0] = max(worst[0], value)
+        return value
```
CoronelBuendia, line 69: You are applying an additional penalty in the objective function for a constraint that should be correctly captured in a gradient-based optimiser. I have recently noticed issues with the application of this constraint in the optimisation algorithms (in particular since NLopt was bumped from 2.7.0).

Ideally I'd like to find a state where the gradient-based optimiser is running correctly (i.e. never violating its constraints). I assume this has cropped up on this run... so just for future context:
* with the strange ripple bug we presently have
* with this version of NLOpt

We will probably need to revisit this.

bluemira/geometry/optimisation/_tools.py
```diff
@@ -141,11 +164,78 @@ def calculate_signed_distance(
...
-    # remain constant.
-    s = shape.discretise(n_shape_discr, byedges=False).xz
+    # remain constant. Sample densely, then resample to the requested count
+    # from a backend-independent anchor (see _canonical_loop_points).
+    dense = shape.discretise(max(8 * n_shape_discr, 200), byedges=False).xz
+    s = _canonical_loop_points(dense, n_shape_discr)
     return signed_distance_2D_polygon(s.T, zone_points.T).T
 
 
+def _canonical_loop_points(loop: np.ndarray, n: int) -> np.ndarray:
```
CoronelBuendia, line 174: This is effectively a discretisation of the CCW convex hull of points. Am I right in understanding it is only required are present because of the wire ordering issue in cadquery?

bluemira/geometry/parameterisations.py
```diff
@@ -1442,9 +1443,13 @@ def _get_centres(
...
     if a_start >= 180 and vertical_symmetry:  # noqa: PLR2004
-        raise GeometryParameterisationError("The total angles should add up to <180°.")
+        raise OutOfBoundsError(
+            "The total angles should add up to <180°.", excess=a_start / 180 - 1
+        )
     if a_start >= 360 and not vertical_symmetry:  # noqa: PLR2004
-        raise GeometryParameterisationError("The total angles should add up to <360°.")
+        raise OutOfBoundsError(
+            "The total angles should add up to <360°.", excess=a_start / 360 - 1
+        )
```
CoronelBuendia, line 1453: As previously commented, I thought this should be handled by the constraints on the GeometryOptimisationProblem

eudemo/eudemo/ivc/wall_silhouette.py
CoronelBuendia: It will work, but only because we are minimising length, and because there are other things "centering" the shape towards the top. Formally it should be an equality constraint (if we want the shape to "touch" the x_point_z - z_margin. Otherwise it could happily satisfy this inequality constraint and let the shape drift far away from the touching point. I think it's fine as is though because an equality constraint would be a pain to satisfy and probably cause errors.

bluemira/codes/cadapi/_cadquery/core.py
```diff
@@ -733,17 +733,22 @@ def _wire_is_planar(wire: apiWire) -> bool:
...
-    if isinstance(obj, cq.Edge):
-        return obj.Length()
-    return sum(e.Length() for e in obj.Edges())
+    edges = [obj] if isinstance(obj, cq.Edge) else obj.Edges()
+    total = 0.0
+    for edge in edges:
+        props = GProp_GProps()
+        BRepGProp.LinearProperties_s(edge.wrapped, props)
+        total += props.Mass()
+    return total
```
CoronelBuendia: Wow, was not aware of this subtlety...

CoronelBuendia (commented):
Thanks @s-oli, I have a few comments. No clear steer from my side on some of the changes yet, there are some things I find are surprising in terms of getting this just to work for cadquery, and some things that I think are probably good to have but I'd want to investigate them a bit further.

Part of the problem is that the present run of this EUDEMO thing is actually broken (for some reason unrelated to CAD).

bluemira/geometry/optimisation/_tools.py
s-oli: Thanks, that's a fair point and I agree with where you want to get to. Just to clarify what this guard actually does: it only catches OutOfBoundsError, i.e. the case where the geometry literally cannot be built at that point (e.g. arc angles past the symmetry limit). It does not add a penalty for normal constraint violations — inside the feasible region the objective value is unchanged.
The reason it's here: a gradient-based optimiser takes small finite-difference steps, and those steps can land just outside the buildable region, where create_shape() raises and the whole run crashes. The penalty just pushes it back instead of dying.
So I think we're actually agreeing: if the constraint were always respected, this would never fire — it's a safety net for exactly the NLopt-since-2.7.0 behaviour you mention. Happy to dig into that constraint/NLopt issue with you, and once it's solid we can drop the net.

bluemira/geometry/optimisation/_tools.py
s-oli: Good question, and let me correct one thing: it isn't a convex hull. It resamples the same loop along its real arc length (np.interp over the actual perimeter), so every concavity is kept — a hull would throw those away and change the shape. The "CCW" bit is only fixing the winding direction so it's identical on both backends.

And yes, this is correct: discretise starts at the wire's first edge and follows the wire's stored direction, and FreeCAD vs CadQuery order/orient the edges differently. That just phase-shifts (and can reverse) the samples. For the keep-out-zone constraint — one signed distance per sample point — that shift means element k is a different physical point on each backend, so the constraint vector and its gradients diverge and the optimiser ends up in a different basin.

I'd frame it as a general robustness fix rather than a CadQuery patch, though: anything that moves the discretisation start (a re-mesh, a different parameterisation) would bite the same way. CadQuery just surfaced it. The resample makes the constraint deterministic regardless of the backend.

bluemira/geometry/parameterisations.py
s-oli: Yep - same story as the objective comment: with the constraint reliably enforced this should never trigger.

CoronelBuendia:
@s-oli thanks for this great work. I am progressively cherry-picking certain commits and submitting them to the feature branch as discussed. I think we should probably get all of these in, but the fixes to the parameterisation and optimisation stuff will take me longer to look and I need to check them with both CAD backends.

I'm leaving this open for the time being, but I will get round to it after the cadquery feature branch is in.

CoronelBuendia:
As a summary for all involved, @s-oli , @je-cook , @Hudsonbaker4 over the past few weeks we have progressively cherry-picked various improvements in this PR (rebased now, so the diff looks less significant).

There is one outstanding commit here which is promising but needs to be evaluated more: https://github.com/Fusion-Power-Plant-Framework/bluemira/pull/4495
It is broad-reaching and we need to study the implications in a variety of situations before proceeding. We already have a couple  of other approaches to treating the same problem, so we also need to discuss whether and how these approaches co-exist.

The remaining commits are:
* 1b25a948d6cd8804c88bee1f2bc93557c5faa684, which was only required for freecad/cadquery matching, and should have been made redundant by wire-ordering conventions as addressed by @Hudsonbaker4 
* ad1a372d3218cfb0ff6b17bf3592228a6a057c72, which can in some situations be considered an improvement, but was only ever required in the first place due some very strange TF shapes being produced (from a totally unrelated and now fixed bug)

Thanks again to @s-oli for all this great work.

PR #4378, opened 2026-06-15
🎨 Viewer backend selection robustness

je-cook:
## Description

Sets the default viewer backend to the default cad backend

## Checklist

### DIFF
--- a/bluemira/display/displayer.py
+++ b/bluemira/display/displayer.py
@@ -62,22 +62,26 @@ class ViewerBackend(Enum):
                 name = self.name.lower()
                 bluemira_warn(
                     f"Unable to import {name.capitalize()} viewer\n"
-                    f"Please 'pip install {name}' to use, falling back to FreeCAD."
+                    f"Please 'pip install {name}' to use,"
+                    f" falling back to {self.DEFAULT.name.lower()}."
                 )
-                return get_module(type(self).DEFAULT.value)
+                return get_module(self.DEFAULT.value)
             raise
 
     @classmethod
     def _missing_(cls, value):
         if isinstance(value, str):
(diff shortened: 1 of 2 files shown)


bluemira/display/displayer.py
```diff
@@ -52,17 +58,29 @@ def get_module(self):
         try:
             return get_module(self.value)
         except (ModuleNotFoundError, FileNotFoundError):
-            if self.name != "FREECAD":
+            if self.name != self.DEFAULT.name:
                 name = self.name.lower()
                 bluemira_warn(
                     f"Unable to import {name.capitalize()} viewer\n"
                     f"Please 'pip install {name}' to use, falling back to FreeCAD."
```
Hudsonbaker4: Being pedantic, I suppose there is a case where the user sets their default backend to CadQuery, then requests Polyscope in the code. If the import were to fail here, it would fallback to CadQuery rather than FreeCAD, so the warning message would be incorrect.

bluemira/display/displayer.py
```diff
@@ -52,17 +58,29 @@ def get_module(self):
...
+                return get_module(type(self).DEFAULT.value)
             raise
 
+    @classmethod
+    def _missing_(cls, value):
+        if isinstance(value, str):
+            try:
+                backend = ViewerBackend[value.upper()]
+            except KeyError:
+                bluemira_warn(f"Unknown viewer backend '{value}' defaulting to FreeCAD")
```
Hudsonbaker4: Same here.

bluemira/display/displayer.py
je-cook: yeah ok thanks

PR #4381, opened 2026-06-15
Seperate seperatrix legs when not off grid

geograham:
## Description

Put a break (at max z) in the separatrix legs if they do not extend off the equilibrium grid. 
This allows divertor leg identification and separation in the case where we have a flux surface that wraps around the lower/upper coils (see #4176 and associated).

~~Have also added functions to the coordinates class that allow you to shift the start  point of a coordinates array and to split the coordinates at a given x, y, z-point.~~

POST REVIEW: have modified LegFlux class with the above separatrix break and modified so that the class has a legs property which is found on initialisation. 

POST REVIEW: Moved changes out of coordinates class and into legflux. Now have split function that can be use on DN or SN. Added test. 

## Checklist

### DIFF
--- a/bluemira/equilibria/find.py
+++ b/bluemira/equilibria/find.py
@@ -875,22 +875,7 @@ def find_LCFS_separatrix(
 
         coords = find_flux_surfs(x, z, psi, high, o_points=o_points, x_points=x_points)
         loops = [Coordinates({"x": c.T[0], "z": c.T[1]}) for c in coords]
-        if flux_surface.closed and len(loops) == 1:
-            # Split in two if we have one separatrix that loops
-            # around the coils within the grid.
-            loop = loops[0]
-            x, z = (
-                loop.x[loop.z == np.max(loop.z)][0],
-                loop.z[loop.z == np.max(loop.z)][0],
-            )
-            loop.shift_start(x=x, z=z)
-            x, z = (
-                loop.x[loop.z == np.min(loop.z)][0],
-                loop.z[loop.z == np.min(loop.z)][0],
-            )
-            loops = loop.split_open(x=x, z=z)
         loops.sort(key=lambda loop: -loop.length)
-        [lp.set_ccw() for lp in loops]
         separatrix = loops[:2]
 
     return lcfs, separatrix
--- a/bluemira/equilibria/find.py
+++ b/bluemira/equilibria/find.py
@@ -875,7 +875,22 @@ def find_LCFS_separatrix(
 
         coords = find_flux_surfs(x, z, psi, high, o_points=o_points, x_points=x_points)
         loops = [Coordinates({"x": c.T[0], "z": c.T[1]}) for c in coords]
+        if flux_surface.closed and len(loops) == 1:
+            # Split in two if we have one separatrix that loops
+            # around the coils within the grid.
+            loop = loops[0]
+            x, z = (
+                loop.x[loop.z == np.max(loop.z)][0],
+                loop.z[loop.z == np.max(loop.z)][0],
+            )
+            loop.shift_start(x=x, z=z)
+            x, z = (
+                loop.x[loop.z == np.min(loop.z)][0],
+                loop.z[loop.z == np.min(loop.z)][0],
+            )
+            loops = loop.split_open(x=x, z=z)
         loops.sort(key=lambda loop: -loop.length)
+        [lp.set_ccw() for lp in loops]
         separatrix = loops[:2]
 
     return lcfs, separatrix
--- a/bluemira/equilibria/find_legs.py
+++ b/bluemira/equilibria/find_legs.py
@@ -117,12 +117,12 @@ class LegFlux:
                     loop.x[loop.z == np.max(loop.z)][0],
                     loop.z[loop.z == np.max(loop.z)][0],
                 )
-                loop.shift_start(x=x, z=z)
+                loop.shift_start([x, 0.0, z])
                 x, z = (
                     loop.x[loop.z == np.min(loop.z)][0],
                     loop.z[loop.z == np.min(loop.z)][0],
                 )
-                loops = loop.split_open(x=x, z=z)
+                loops = loop.split_open([x, 0.0, z])
                 loops.sort(key=lambda loop: -loop.length)
                 [lp.set_ccw() for lp in loops]
                 self.separatrix = loops
--- a/bluemira/equilibria/find_legs.py
+++ b/bluemira/equilibria/find_legs.py
@@ -90,10 +90,10 @@ class LegFlux:
         self.o_point = o_points[0]
         self.x_points = x_points[:2]
         # Check we are using the x-points nearest the LCFS max and/or min z-coord.
-        (xlow,) = np.nonzero(
+        xlow = np.flatnonzero(
             np.isclose(np.min(lcfs.z), [xp.z for xp in self.x_points], rtol=delta_start)
         )
-        (xup,) = np.nonzero(
+        xup = np.flatnonzero(
             np.isclose(np.max(lcfs.z), [xp.z for xp in self.x_points], rtol=delta_start)
         )
         check_dn = isinstance(self.separatrix, list) and (len(xlow) + len(xup) < 2)  # noqa: PLR2004
--- a/bluemira/equilibria/find_legs.py
+++ b/bluemira/equilibria/find_legs.py
@@ -110,8 +110,15 @@ class LegFlux:
         # We have not introduced a first wall or divertor silhouette,
         # so sometimes this can happen if the flux surface wraps around the
         # lower divertor coils rather that extending off the grid.
-        if eq.is_double_null and len(self.separatrix) != 2:  # noqa: PLR2004
-            self.separatrix = split(self.separatrix)
+        if (eq.is_double_null and len(self.separatrix) != 2) or (  # noqa: PLR2004
+            not eq.is_double_null and self.separatrix.closed
+        ):
+            self.separatrix = split(
+                self.separatrix
+                if isinstance(self.separatrix, Coordinates)
+                else self.separatrix[0],
+                self.n_null,
+            )
 
         self.rtol = rtol
         self.x_range_lcfs = [min(lcfs.x), max(lcfs.x)]
@@ -124,45 +131,6 @@ class LegFlux:
         if n_layers > 1:
             legs = self._get_leg_offsets(legs, eq)
         self._legs = legs
-
-    def _split(self):
-        """
-        Method used to split a separatrix in two if we have one that loops
-        around the coils within the grid and thus appears 'closed'.
-        """
-        loop = (
-            self.separatrix
-            if isinstance(self.separatrix, Coordinates)
-            else self.separatrix[0]
-        )
-
-        max_point = np.array([
-            loop.x[loop.z == np.max(loop.z)][0],
-            0.0,
-            loop.z[loop.z == np.max(loop.z)][0],
-        ])
-        min_point = np.array([
-            loop.x[loop.z == np.min(loop.z)][0],
-            0.0,
-            loop.z[loop.z == np.min(loop.z)][0],
-        ])
-
-        # Move start of array to top of upper divertor region and open coords
-        idx = len(loop._array[0]) - loop.argmin(max_point)
-        for i in [0, 1, 2]:
-            loop._array[i] = np.roll(loop._array[i], idx)
-
-        # Split into two at bottom of lower divertor region
-        idx = loop.argmin(min_point) + 1
-        loops = [
-            Coordinates(self._array[:, :idx]),
-            Coordinates(self._array[:, idx:]),
-        ]
-
-        # Sort, set direction and update
-        loops.sort(key=lambda loop: -loop.length)
-        [lp.set_ccw() for lp in loops]
-        self.separatrix = loops
 
     @property
     def legs(self):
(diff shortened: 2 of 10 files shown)


bluemira/equilibria/find_legs.py
```diff
@@ -89,6 +89,23 @@ def __init__(
         )
         self.o_point = o_points[0]
         self.x_points = x_points[:2]
+        # Check we are using the x-points nearest the LCFS max and/or min z-coord.
+        (xlow,) = np.nonzero(
```
je-cook: maybe use np.flatnonzero?https://numpy.org/doc/stable/reference/generated/numpy.flatnonzero.html#numpy-flatnonzero

bluemira/equilibria/find.py
```diff
@@ -867,7 +867,22 @@ def get_flux_loop(psi_norm):
...
+                loop.z[loop.z == np.max(loop.z)][0],
+            )
+            loop.shift_start(x=x, z=z)
+            x, z = (
+                loop.x[loop.z == np.min(loop.z)][0],
+                loop.z[loop.z == np.min(loop.z)][0],
+            )
+            loops = loop.split_open(x=x, z=z)
         loops.sort(key=lambda loop: -loop.length)
+        [lp.set_ccw() for lp in loops]
```
CoronelBuendia: This logic should be moved to leg classification. This function is complex enough as it is, and changing the resulting data will affect many other things, not just the leg classification (e.g. now all plots of faux-closed separatrices will have a funny blank spot on them).

bluemira/equilibria/find_legs.py
```diff
@@ -89,6 +89,23 @@ def __init__(
...
+        check_dn = isinstance(self.separatrix, list) and (len(xlow) + len(xup) < 2)  # noqa: PLR2004
+        check_sn = not isinstance(self.separatrix, list) and (len(xlow) + len(xup) < 1)
+        if check_dn or check_sn:
+            bluemira_print(
+                "Seperatrix shape is not as expected. Best keep an eye on it, "
+                "as seperatrix leg identification may not work."
+            )
+            self._red_flag = True
+        else:
+            self._red_flag = False
```
CoronelBuendia: This is not really a robust way of identifying SN/DN, and depends on the "opinionated" data being returned by find_LCFS_separatrix (in this PR).

Is there a reason `eq.is_double_null` does not work for this use case?

This is really where the treatment of the case of faux-closed flux surfaces should be handled.

bluemira/geometry/coordinates.py
```diff
@@ -1588,6 +1588,61 @@ def open(self):
...
+            & np.isclose(self._array[2], z)
+        )
+        if len(idx) > 1:
+            bluemira_warn(
+                "There is more than one point at this location, returning first value."
+            )
+            return [idx[0]]
+        if len(idx) < 1:
+            bluemira_warn("There is no point at this location.")
+        return idx
```
CoronelBuendia: Why not use `self.argmin(point)` and `self.distance_to(point)`. This looks to be shadowing that behaviour somewhat, but with np's isclose tolerance argument defaults.

bluemira/geometry/coordinates.py
```diff
@@ -1588,6 +1588,61 @@ def open(self):
...
+        """
+        if self.closed:
+            self.open()
+        idx = self._get_loc_idx(x, y, z)
+        if len(idx) != 1:
+            bluemira_warn("Doing nothing to coordinates start location.")
+        else:
+            self._array = np.concatenate(
+                [self._array[:, idx[0] :], self._array[:, : idx[0]]], axis=1
+            )
```
CoronelBuendia: Won't this mean that an open polygon [A, B, C, D], when shift_started to C, now read [C, D, A,B], thereby making it a self-intersecting polygon, and also messing with CW/CCW, and general length/COM properties?

np.roll for closed polygons here

bluemira/geometry/coordinates.py
```diff
@@ -1588,6 +1588,61 @@ def open(self):
         if self.closed:
             self._array = self._array[:, :-1]
 
+    def shift_start(self, x=0.0, y=0.0, z=0.0):
```
CoronelBuendia: we use `point` for all other methods in this class I think

bluemira/geometry/coordinates.py
```diff
@@ -1588,6 +1588,61 @@ def open(self):
...
+        if len(idx) > 1:
+            bluemira_warn(
+                "There is more than one point at this location, returning first value."
+            )
+            return [idx[0]]
+        if len(idx) < 1:
+            bluemira_warn("There is no point at this location.")
+        return idx
+
+    def split_open(self, x=0.0, y=0.0, z=0.0):
```
CoronelBuendia: point

bluemira/geometry/coordinates.py
```diff
@@ -1588,6 +1588,61 @@ def open(self):
...
+        -------
+        :
+            Coordinates from start of input coordinates up to
+            (but not including) user chosen point
+        :
+            Coordinates from user chosen point to the end of the
+            input coordinates
+        """
+        if self.closed:
+            self.open()
```
CoronelBuendia: This is going to mutate the instance, and then in any case returns new objects. Please leave the original data unaffected.

bluemira/geometry/coordinates.py
```diff
@@ -1588,6 +1588,61 @@ def open(self):
         if self.closed:
             self._array = self._array[:, :-1]
 
+    def shift_start(self, x=0.0, y=0.0, z=0.0):
+        """
+        Change the start location of your coordinates array.
+        Will set to open if coords are closed.
+        """
+        if self.closed:
+            self.open()
```
CoronelBuendia: Why do this? I would not expect `shift_start` to open an instance of `Coordinates`. Do you need to do this for this function to work?

CoronelBuendia (changes requested):
Hi @geograham, sorry several big ones here... You're modifying / adding a lot of key things here, all to achieve something which is quite specific. We should aim to keep the generic as all-purpose as possible.

bluemira/equilibria/find.py
geograham: Makes sense and have moved.

geograham:
Thanks again for the review @CoronelBuendia, I've made the changes you suggested.

bluemira/equilibria/find_legs.py
```diff
@@ -85,18 +102,128 @@ def __init__(
...
+                loop = self.separatrix
+                x, z = (
+                    loop.x[loop.z == np.max(loop.z)][0],
+                    loop.z[loop.z == np.max(loop.z)][0],
+                )
+                loop.shift_start([x, 0.0, z])
+                x, z = (
+                    loop.x[loop.z == np.min(loop.z)][0],
+                    loop.z[loop.z == np.min(loop.z)][0],
+                )
```
CoronelBuendia: Is `shift_start` even required here? It looks redundant... if `split_open` at a given point works, why does one need to shift the starting point?

bluemira/geometry/coordinates.py
```diff
@@ -1588,6 +1588,36 @@ def open(self):
...
             self._array = self._array[:, :-1]
 
+    def shift_start(self, point: npt.ArrayLike):
+        """
+        Change the start location of your coordinates array.
+        Will set to open if coords are closed.
+        """
+        point = np.array(point)
+        idx = len(self._array[0]) - self.argmin(point) - 1
+        self._array = np.roll(self._array, idx)
```
CoronelBuendia: This will still not work for open Coordinates. Case in point:
```python
from copy import deepcopy
from bluemira.geometry.coordinates import Coordinates
c = Coordinates({"x":[0,1,2,3,4], "y":[0,1,2,3,4]})
c2 = deepcopy(c)
c2.shift_start([2, 2, 0])
print(c.length)  # 5.656854249492381
print(4*np.sqrt(2)) # 5.656854249492381
print(c2.length)  # 9.899494936611667
```
I'm also not convinced it is required...

bluemira/geometry/coordinates.py
```diff
@@ -1588,6 +1588,36 @@ def open(self):
...
+        :
+            Coordinates from user chosen point to the end of the
+            input coordinates
+        """
+        point = np.array(point)
+        idx = self.argmin(point)
+        return [
+            Coordinates(self._array[:, : idx[0]]),
+            Coordinates(self._array[:, idx[0] :]),
+        ]
```
CoronelBuendia: Sonarcloud is flagging this and I can see why. `self.argmin` should return an int.

```python
c.split_open([2,2,0])
  File "/home/matti/code/bluemira/bluemira/geometry/coordinates.py", line 1617, in split_open
    Coordinates(self._array[:, : idx[0]]),
                                 ~~~^^^
IndexError: invalid index to scalar variable.

```

Possibly why the tests are failing, but haven't checked

bluemira/equilibria/find_legs.py
```diff
@@ -85,18 +102,128 @@ def __init__(
...
-            delta_start=delta_start,
+            delta_start=delta_lcfs,
         )
-        self.o_point = o_points[0]
-        self.x_points = x_points[:2]
+        self._set_oxpts_and_check(lcfs, delta_lcfs, o_points, x_points)
+
+        if eq.is_double_null and isinstance(self.separatrix, Coordinates):
+            if not self.separatrix.closed:
+                self._print_warning_set_flag()
```
CoronelBuendia: In a "normal" DN, with pure up-down symmetry, the separatrix will consist of two open Coordinates, no?

CoronelBuendia (changes requested):
Partial review, sorry for the hold-up. I still need to properly understand what is going in find_legs.

bluemira/equilibria/find_legs.py
```diff
@@ -85,18 +102,128 @@ def __init__(
...
+        xlow = np.flatnonzero(
+            np.isclose(np.min(lcfs.z), [xp.z for xp in self.x_points], rtol=delta_lcfs)
+        )
+        xup = np.flatnonzero(
+            np.isclose(np.max(lcfs.z), [xp.z for xp in self.x_points], rtol=delta_lcfs)
+        )
+        if self.n_null == NumNull.SN and (len(xlow) + len(xup) < 1):
+            self._print_warning_set_flag()
+        if self.n_null == NumNull.DN and (len(xlow) + len(xup) < 2):  # noqa: PLR2004
+            self._print_warning_set_flag()
```
CoronelBuendia, line 217: The X-points are already sorted by normalised flux space, with the most relevant (active, closest in normalised flux to LCFS) being the first two entries, followed by any inactive X-points.

Here you are checking again for proximity, but this time in physical space - why exactly? The more natural  way to classify a DN/SN is in terms of dR_sep at the midplane, that is to say the distance between the flux surfaces passing through the two active/inactive X-points. We used to do this in BLUEPRINT, IIRC, but now I don't think it is done so explicitly anymore.

So, is there a reason why you need a different check for the number of "active" nulls in an equilibrium? Is it that you still need the "DN" separatrix in a SN case? Right now all this seems to do is warn and set a private flag.

geograham: I was getting some cases (not good eq) where the secondary x-point near the divertor target for a short inner leg was getting chosen rather than the expected. I have added this check + flag just so that I can tell if this is what is causing issues in the div designer further down the line.

CoronelBuendia: Ah, right - yes, the present approach will not properly sort X-points XD configurations I guess.

I appreciate we need to move on with this, so I've written up an issue [here](https://github.com/Fusion-Power-Plant-Framework/bluemira/issues/4463) for us to treat the general case more properly. If you have a picture of this situation you describe to add to that issue, that would be great. Please add a TODO with this issue name here, too.

The use of private flags effectively means you will be the only one using this (or else we have a bigger problem).

bluemira/equilibria/find_legs.py
```diff
@@ -205,41 +336,61 @@ def get_legs(
...
-            self.separatrix, self.delta, self.x_points, self.o_point
+        return get_legs_double_null_xsplit(
+            self.separatrix, self.delta_legs, self.x_points, self.o_point
         )
-        if self.dx_offsets is not None:
-            return self.get_leg_offsets(leg_dict)
-        return leg_dict
+
+
+def split(separatrix):
```
CoronelBuendia: This seems to be a duplicate from the above?

bluemira/equilibria/find_legs.py
```diff
@@ -85,18 +102,152 @@ def __init__(
...
-        self.o_point = o_points[0]
-        self.x_points = x_points[:2]
+        self._set_oxpts_and_check(lcfs, delta_lcfs, o_points, x_points)
+        # If we have a 'closed' flux surface then we need to
+        # artificially open it at the top and bottom.
+        # We have not introduced a first wall or divertor silhouette,
+        # so sometimes this can happen if the flux surface wraps around the
+        # lower divertor coils rather that extending off the grid.
+        if eq.is_double_null and len(self.separatrix) != 2:  # noqa: PLR2004
+            self.separatrix = split(self.separatrix)
```
CoronelBuendia: I know it's not a problem now, but what about single nulls with the same "looping around a coil" issue?

Happy for an issue to go up with a TODO here too, and kick it down the line (rather it being potentially forgotten about!)

geograham: I have added the SN option and put in a test.

CoronelBuendia (approved):

PR #4382, opened 2026-06-15
Add an unfix coil size option

geograham:
It is sometimes useful to be able to unfix coil sizes even if you have previously chosen dimensions. Creating a new option to unfix the coil sizes without having to create a whole new coilset and equilibria. 

This will be done via a boolean property called 'fix_coils' that can be set to True or False (or an array of T/F values if you want a mix of coils with fix and unfixed sizes). Default is None, which will behave as it did before, automatically fixing the coil sizes if they are specified or unfixing them if they are not specified.  

## Checklist

### DIFF
--- a/bluemira/equilibria/coils/_coil.py
+++ b/bluemira/equilibria/coils/_coil.py
@@ -15,7 +15,7 @@ from typing import TYPE_CHECKING
 import numpy as np
 
 from bluemira.base.constants import CoilType
-from bluemira.base.look_and_feel import bluemira_debug
+from bluemira.base.look_and_feel import bluemira_debug, bluemira_print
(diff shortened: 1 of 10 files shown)


je-cook (changes requested):
Can we have a property called `fix_size`
```
coil.fix_size = True # or False
```
We can also add a flag to init to override the default state with the same name by default it will be None to keep the api the same

geograham:
> Can we have a property called `fix_size`
> 
> ```
> coil.fix_size = True # or False
> ```
> 
> We can also add a flag to init to override the default state with the same name by default it will be None to keep the api the same

This is a better idea and I have changed to address.

je-cook (approved):
Looks good, it looks like you've missed some fix_size() -> fix_size in demo

PR #4386, opened 2026-06-18
Fix for upper port test

je-cook:
## Description

This is not a great fix but it works locally with both backends.

We're try to match the sorting of both the inner and outer wires and check they are the right distance apart. This was reliably done by index before but now that doesnt seem to work (at least in my naiive attempts).

The sum of the y coordinate of the start and end and the length reliably gets the parallel wires but separating the distances is not trivial. Ideas welcomed..
## Checklist


CoronelBuendia:
@Hudsonbaker4 I think I messed up a rebase here, sorry...

je-cook:
If you remove the change on the bounding box function itself back to the 'better' one everything passes locally for me. Pushed a commit to see if its just me

je-cook:
also ignore the docs failure we need to rebase the stack on develop where the fix is

PR #4398, opened 2026-06-25
Update CadQuery wire_from_wires behaviour to mirror FreeCAD

Hudsonbaker4:
## Linked Issues

Closes #{ID}

## Checklist

### DIFF
--- a/bluemira/codes/cadapi/_cadquery/core.py
+++ b/bluemira/codes/cadapi/_cadquery/core.py
@@ -997,6 +997,10 @@ def wire_from_wires(wire_list: list[apiWire]) -> apiWire:
         return cq.Wire(builder.Wire())
 
     # If the sequential builder failed, fallback to combine algorithm.
+    bluemira_warn(
+        "Failed to join wires sequentially in the order provided. "
+        "Falling back to spatial assembly to heal the shape."
+    )
     result = cq.Wire.combine(wire_list)
     if isinstance(result, list) and len(result) > 1:
         bluemira_warn(
--- a/bluemira/codes/cadapi/_cadquery/core.py
+++ b/bluemira/codes/cadapi/_cadquery/core.py
@@ -974,11 +974,29 @@ def wire_from_edges(edge_list: list[apiEdge]) -> apiWire:
 def wire_from_wires(wire_list: list[apiWire]) -> apiWire:
     """Create a single wire from a list of connected wires.
 
+    Uses BRepBuilderAPI_MakeWire builder to preserve the exact memory order
+    of the input list, perfectly matching FreeCAD's behavior for closed loops.
+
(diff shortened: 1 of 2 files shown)


bluemira/codes/cadapi/_cadquery/core.py
```diff
@@ -733,22 +733,17 @@ def _wire_is_planar(wire: apiWire) -> bool:
...
-    edges = [obj] if isinstance(obj, cq.Edge) else obj.Edges()
-    total = 0.0
-    for edge in edges:
-        props = GProp_GProps()
-        BRepGProp.LinearProperties_s(edge.wrapped, props)
-        total += props.Mass()
-    return total
+    if isinstance(obj, cq.Edge):
+        return obj.Length()
+    return sum(e.Length() for e in obj.Edges())
```
CoronelBuendia: Is this reversion back to the old behaviour required for this new ordering behaviour to work? The length calculation changes @s-oli brought in matched FreeCAD's behaviour to machine precision, would be a shame to lose it (particularly for optimisation purposes)

bluemira/codes/cadapi/_cadquery/core.py
```diff
@@ -974,11 +969,29 @@ def wire_from_edges(edge_list: list[apiEdge]) -> apiWire:
...
+
+    # Build the wire sequentially.
+    builder = BRepBuilderAPI_MakeWire()
+    for e in all_edges:
+        builder.Add(e.wrapped)
+
+    if builder.IsDone():
+        return cq.Wire(builder.Wire())
+
+    # If the sequential builder failed, fallback to combine algorithm.
```
CoronelBuendia, line 999: Could perhaps do with a warning here. Do you have any idea what sort of situations the fallback occurs in?

CoronelBuendia (changes requested):
Thanks @Hudsonbaker4 for expediting this, just a couple of comments.

bluemira/codes/cadapi/_cadquery/core.py
Hudsonbaker4: Whoops, no that's not supposed to be there. I must have messed something up cherry picking commits onto this new branch.

bluemira/codes/cadapi/_cadquery/core.py
Hudsonbaker4: Yes, I'll add that. I believe cq.Wire.combine can stitch together an unordered input list of wires, whereas the first approach takes the wires strictly in the order they were given. The fallback occurs if the input list is handed out of order, or if there are tiny gaps between the segments, which forces CadQuery to heal the shape.

CoronelBuendia (approved):

CoronelBuendia:
Can confirm this matches the FW, blanket and divertor build as in Freecad (to visual precision)

PR #4418, opened 2026-07-10
proposal to solve boolean fuse problems for faces with flipped orientation

s-oli:
The CadQuery backend's boolean_fuse produced a multi-face compound ("gives
more than one face") for coplanar faces whose normals are anti-parallel.

Mirror the FreeCAD backend's _make_shapes_coaxis: reverse anti-parallel faces
to a common normal before fusing, then collapse the overlap via clean().
The sew pass stays as a fallback for edge-only-touching faces.

Includes regression tests and updates the stale _unify_same_domain exception
test to the clean() path.


je-cook (approved):
Thanks very much. I think this is the first stage of fixes for boolean fuse, I think there may need to be a few more!

PR #4436, opened 2026-07-20
fix(cad): don't let splitter removal corrupt spline-bounded solids

s-oli:
## Description

boolean_fuse tidies its result with ShapeUpgrade_UnifySameDomain (via cadquery's clean()). On a solid bounded by B-spline surfaces that can merge faces it should not, returning a shape that fails BRepCheck and reports a nonsense volume -- observed: -1146 m^3 for a 31 m^3 solid -- while the shallow isValid() still passes. boolean_fuse then extracts that solid from the broken compound, where isValid() passes again, and hands it to the caller. The next boolean either yields garbage or dies on a null shape.

Move the tidy-up into _remove_splitter, which refuses a result that is invalid when the input was not. Splitter faces are cosmetic; a corrupt solid is not. Repairing was not an option: the unified shape's volume is wrong (32.83 vs 31.01), so geometry had been merged away and ShapeFix would only have made it valid-and-still-wrong. The guard is solids-only -- on faces the same tidy-up doubles as a repair step and an unscoped guard breaks StraightOISDesigner. The bare contextlib.suppress(Exception) becomes a warning: still never raises, but no longer silent.

Surfaced by EUDEMO's VacuumVessel.add_ports, which joins port after port onto the accumulated result. Cylindrical and polygonal vessels never showed it; the real vessel wall is a spline (600-point discretise plus force_wire_to_spline), which is why this only ever bit complex geometry.

Also raise a clear GeometryError in pipe_pipe_join when the target and tool do not intersect, instead of an unguarded target_fragments[0] IndexError.

## Checklist

### DIFF
--- a/bluemira/codes/cadapi/_cadquery/core.py
+++ b/bluemira/codes/cadapi/_cadquery/core.py
@@ -1606,15 +1606,34 @@ def _merge_fused_faces(result: apiShape, shapes: list) -> apiShape:
     unified_faces = _collect_subshapes(unified, cq.Face)
     if len(unified_faces) == 1:
         return unified_faces[0]
-    out = shapes
-    for _ in range(2):
-        out = _sew_shapes(out)
-        if isinstance(out, cq.Face):
-            return out
-        sewn_faces = _collect_subshapes(out, cq.Face)
-        if len(sewn_faces) == 1:
-            return sewn_faces[0]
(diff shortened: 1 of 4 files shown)


CoronelBuendia (commented):
LGTM! Needs a rebase - appears a different solution was developed in the meantime somehow.

je-cook:
rebased

PR #4439, opened 2026-07-22
Add simple carabiner parameterisation

CoronelBuendia:
## Linked Issues

Closes #{ID}

## Checklist


bluemira/geometry/parameterisations.py
```diff
@@ -2949,3 +2949,430 @@ def _make_arc(center, a, b, start, end):
...
+        description="Inner limb radius",
+    )
+    z1: OptVariable = ov(
+        "z1",
+        3.0,
+        lower_bound=0,
+        upper_bound=5,
+        description="Vertical position of the inner limb start point",
+    )
+    dz: OptVariable = ov(
```
josieapeters, line 3169: (just for my understanding :) )  what is vertical offset for? if you have the coords of the start point of the straight edhe / inner limb, why not absorb it into there?

bluemira/geometry/parameterisations.py
```diff
@@ -2949,3 +2949,430 @@ def _make_arc(center, a, b, start, end):
...
+        wires, center, start, end, _, p2, p3, _, _ = self._preamble()
+        wires.sort(key=lambda w: w.center_of_mass[0])
+
+        arc_3 = make_circle(
+            r3,
+            center=(center[0], 0, center[1]),
+            start_angle=start,
+            end_angle=end,
+            axis=(0, -1, 0),
+            label="arc_2",
```
josieapeters:
```suggestion
            label="arc_3",
```

bluemira/geometry/parameterisations.py
```diff
@@ -2949,3 +2949,430 @@ def _make_arc(center, a, b, start, end):
             start_angle=start,
             end_angle=end,
         )
+
+
+def solve_tangent_arc(p1, t1, p3, t3):
```
CoronelBuendia: Not used

josieapeters (changes requested):
Looks good! :)

Just needs a simple constraint that r3 is smaller than r1 and r2, (if someone is inputting r3 rather than letting the function find it) 

I played around with making r3 bigger than r1 and/or r2 the carabiner inverts itself. Maybe to do with the wire ordering based on the centre of mass? Not sure!
with r1 = 1, r2 = 2, r3 = 6
Image
with r1 = 6, r2 = 5, r3 = 6
Image
r1 = 3, r2 = 7, r3 = 6
Image

bluemira/geometry/parameterisations.py
CoronelBuendia: This is something we use in most if not all parameterisations, and just lets the optimiser drift the shape up and down with a single variable. This is helpful sometimes, similar to how offset is helpful here too.

CoronelBuendia:
> Looks good! :)
> 
> Just needs a simple constraint that r3 is smaller than r1 and r2, (if someone is inputting r3 rather than letting the function find it)
> 
> I played around with making r3 bigger than r1 and/or r2 the carabiner inverts itself. Maybe to do with the wire ordering based on the centre of mass? Not sure! with r1 = 1, r2 = 2, r3 = 6 Image with r1 = 6, r2 = 5, r3 = 6 Image r1 = 3, r2 = 7, r3 = 6 Image

Yeah this is a wee bit tricky, because the optimiser in this case can explore places where the constraints are violated, so we don't really want to error. I've added a check upon initialisation that performs this check on input variable combinations.

The relevant constraints are that the center of the fillet arc must have a z-value that is lower than that of the point at which it joins the straight line or arc.

bluemira/geometry/parameterisations.py
```diff
@@ -2949,3 +2949,407 @@ def _make_arc(center, a, b, start, end):
...
+                axis=(0, -1, 0),
+                label=f"arc_{i + 1}",
+            )
+            wires.append(arc)
+
+        p1 = [x1, 0.0, z1]
+        t1 = [0, 0, -1.0]
+        p2 = arc.start_point().xyz.T[0]
+        t2 = arc._shape.Edges[0].tangentAt(arc._shape.Edges[0].FirstParameter)
+        t2 = -np.array([t2.x, t2.y, t2.z])
```
je-cook:
```suggestion
        t2 = -arc.tangent_at(0)
```

josieapeters (approved):

PR #4447, opened 2026-07-27
🧪 Test boolean split with wires

### LINKED ISSUE #3914: Possibly redundant branch in `_freecadapi.boolean_cut`
### Describe the bug

Continuing from the investigation of #3897 : it appears that the `if split` [branch inside the](https://github.com/Fusion-Power-Plant-Framework/bluemira/blob/develop/bluemira/codes/_freecadapi.py#L2321) `_freecadapi.boolean_cut` function does nothing, at least on the cases that I tested. Perhaps someone more enlightened than I am would need to look at this and explain if this conditional branch need to exist at all.

### Steps to reproduce

```python
# script to investigate the boolean_cut in _freecadapi.

from bluemira.geometry.tools import make_polygon, make_circle, extrude_shape, revolve_shape
from bluemira.geometry.solid import BluemiraSolid
from bluemira.codes._freecadapi import boolean_cut
from bluemira.display import plot_2d, show_cad
from bluemira.geometry.face import BluemiraFace

# boudning box 
arc = make_circle(1.0, start_angle=-90, end_angle=90, axis=[0,1,0])
sphere = BluemiraSolid(revolve_shape(arc, degree=360))
xi = -0.7
xa = 0.7
zi = -0.7
za = 0.7
small_face = BluemiraFace(make_polygon([[xi, 0, zi],
[xa, 0, zi],
[xa, 0, za],
[xi, 0, za],], closed=True))
small_cuboid = extrude_shape(small_face, [0, 1, 0]) # Solid

# split = True
out = boolean_cut(small_cuboid.shape, sphere.shape, split=False)
solid = BluemiraSolid._create(out[0])
show_cad(solid)

out = boolean_cut(small_cuboid.shape, sphere.shape, split=True)
solid = BluemiraSolid._create(out[0])
show_cad(solid)
```
Note that the outputs in both cases are exactly the same.

### Tracebacks

```bash

```

### Version

develop

### OS

### OS details

_No response_

je-cook:
## Linked Issues

Closes #3914

## Description

In cadquery this works so I've added a test. in freecad this fails with a bad coplanar check (as in the check to me seems bad)

## Checklist

### DIFF
--- a/tests/codes/_shared/backend_api_tests.py
+++ b/tests/codes/_shared/backend_api_tests.py
@@ -573,6 +573,19 @@ class BackendApiTestsBase:
         sliced_coil = boolean_cut(boolean_cut_princeton, cut_tool)[0]
         assert sliced_coil.is_valid_deep()  # similar to ``test_boolean_cut_hollow_coil``
 
+    def test_boolean_cut_split_arg(self):
+
+        wire1 = make_polygon([[-1, 0, -1], [1, 0, 1]])
+        wire2 = make_polygon([[-1, 0, 1], [1, 0, -1]])
+
+        result = self.cadapi.boolean_cut(wire1.shape, wire2.shape, split=True)
+        assert len(result) == 2
+        assert all(r.Length() == pytest.approx(np.sqrt(2)) for r in result)
+
+        result = self.cadapi.boolean_cut(wire1.shape, wire2.shape, split=False)
+        assert len(result) == 1
+        assert result[0].Length == pytest.approx(2 * np.sqrt(2))
+
     def test_wire_tangent_at_complex(self):
         e1 = self.cadapi.make_polygon([(0, 0, 0), (10, 0, 0)])
 
--- a/tests/codes/_shared/backend_api_tests.py
+++ b/tests/codes/_shared/backend_api_tests.py
@@ -584,7 +584,7 @@ class BackendApiTestsBase:
 
         result = self.cadapi.boolean_cut(wire1.shape, wire2.shape, split=False)
         assert len(result) == 1
-        assert result[0].Length == pytest.approx(2 * np.sqrt(2))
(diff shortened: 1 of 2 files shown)


tests/codes/_shared/backend_api_tests.py
```diff
@@ -573,6 +573,19 @@ def test_boolean_cut_cross_section(self, boolean_cut_princeton):
...
+        wire2 = make_polygon([[-1, 0, 1], [1, 0, -1]])
+
+        result = self.cadapi.boolean_cut(wire1.shape, wire2.shape, split=True)
+        assert len(result) == 2
+        assert all(r.Length() == pytest.approx(np.sqrt(2)) for r in result)
+
+        result = self.cadapi.boolean_cut(wire1.shape, wire2.shape, split=False)
+        assert len(result) == 1
+        assert result[0].Length() == pytest.approx(2 * np.sqrt(2))
+
```
josieapeters, line 588: all looks good. just for my understanding- 2 perpendicular wires, in the first case, split=True means wire2 splits wire1, and checking we now have two lines rather than one, and each length approx the right length. and in the second case... you apply boolean cut but you don't cut the wire? (is that where you'd want to apply a mask ? but not alter the original wire)

josieapeters (approved):

tests/codes/_shared/backend_api_tests.py
je-cook: so by default in the user facing interface in geometry split is not exposed. Boolean operations are not really defined on a wire they more design for solids (and there are usually better ways to split wires) but a user might expect to be able to split the wire. 

When at a lower level its nice to be able to disable this especially when debugging things

josieapeters: got it, thank you! :)

PR #4448, opened 2026-07-27
🧑‍💻 Remove erronious process warnings

je-cook:
## Description

Removes some obsolete variable names because process obs_var doesnt have a mechanism for 'this variable is now only output' therefore because we check on in and out for obsolete variable names we get some false positives. This excludes known false positives.

## Checklist

### DIFF
--- a/bluemira/codes/process/_teardown.py
+++ b/bluemira/codes/process/_teardown.py
@@ -219,6 +219,14 @@ class Teardown(CodesTeardown):
         return output_value
 
 
+ONLY_READ_FROM_FILE = (
+    "dz_blkt_upper",
+    "a_tf_wp_no_insulation",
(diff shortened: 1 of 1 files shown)


josieapeters (approved):
LGTM! :) quick curiosity question - is there a list on PROCESS for which variables are input, output etc? Had a quick scan but couldn't find a list or attritibute for the variables

je-cook:
> LGTM! :) quick curiosity question - is there a list on PROCESS for which variables are input, output etc? Had a quick scan but couldn't find a list or attritibute for the variables

I think they might have inputs somewhere but not a list of outputs as far as I'm aware. In future PROCESS will be dumping out its whole internal data structure so there wont so much of a distinction other than flagged allowed as input.

PR #4454, opened 2026-07-27
Standardise Green's functions

### LINKED ISSUE #4259: Standardise the mathematical expressions and notations in the Greens functions
### Describe the Task

With the incoming RZIp implementation, #3606, we will now have quite a variety of different ways of calculating and writing the equations of the various Greens functions in the magnetostatics module.

There is the potential for confusion at present, and the equations are now more complex to read. There is also the potential for some speed-up were all of these functions to share the same notation (e.g. re-mapping all greens functions onto a grid could be sped up through some common calculations)

### Current and Future State

_No response_

CoronelBuendia:
## Linked Issues

Closes #4259 

Unified mathematical notation and use of helper functions.

Slightly cheaper arithmetic and better treatment of masking and dodging of ZeroDivisionErrors (less frequent now).

Slightly different behaviour due to improvements related to handling near k2 clipping

Also changes `greens_all` to share behaviour with  other calculations (was not the case before)

Also adds explicit analytical treatment of x == 0
Bz:
image

These are probably slightly slower, which could be mitigated by element-wise calculations and vectorisation of all greens_funcs. This would be even harder to review...

Sorry for the horrible diff... I've tried to make up for it with a proper set of regression tests and some plots of the comparisons.

Comparison of differences when xc, zc -> x, z:

image
image
image
image

## Checklist

### DIFF
--- a/bluemira/magnetostatics/greens.py
+++ b/bluemira/magnetostatics/greens.py
@@ -63,7 +63,7 @@ def clip_nb(
 
 @nb.njit(cache=True, inline="always")
 def _common_geometry(xc, zc, x, z):
-    r"""
+    """
     Calculate the geometry shared by the Green's functions.
 
     .. math::
@@ -96,7 +96,7 @@ def _elliptic_integrals(k2):
 
 @nb.njit(cache=True, inline="always")
 def _i1_i2(u, u2, k2, e, k):
-    r"""
+    """
     Calculate
 
     .. math::
@@ -117,7 +117,7 @@ def _i1_i2(u, u2, k2, e, k):
 
 @nb.njit(cache=True, inline="always")
 def _radial_term(xc, x, h, i1, i2):
-    r"""
+    """
     Calculate
 
     .. math::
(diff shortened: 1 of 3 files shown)


bluemira/magnetostatics/greens.py
```diff
@@ -611,34 +459,50 @@ def greens_Bx(
...
     """
-    return -1 / x * greens_dpsi_dz(xc, zc, x, z, d_xc, d_zc)
+    h, u, u2, k2 = _common_geometry(xc, zc, x, z)
+    e, k = _elliptic_integrals(k2)
+    i1, i2 = _i1_i2(u, u2, k2, e, k)
+
+    xc2 = xc * xc
+    w2 = xc2 - x * x - h * h
+
+    bz = MU_0_2PI * (i1 + w2 * i2)
```
kj5248: Duplication. You can just use _radial_term for this.

```suggestion
    bz = MU_0_2PI * _radial_term(...)
```

PR #4458, opened 2026-07-29
🎨 Allow any shaped geometry to be normalised in 2D

### LINKED ISSUE #188: Non convex hull regions for position optimisation of coils
In GitLab by @je-cook on May 13, 2021, 14:58

The implemented position optimiser only works for Convex Hulls. To implement this for any polygon it is reasonably straight forward to normalise real space coordinates however converting to real space coordinates from normalised coordinates is not so straight forward.

To normalise and unnormalise the coordinates to avoid the limitation the only extra consideration is a plane that intersects your region boundaries more that twice. This currently raises an error within the code.

If that is true for normalising the coordinates you can effectively collapse the gaps in your region and normalise as if the region is a convex hull by subtracting the empty regions. 

- Possible downside optimisers may struggle with the non-linear relationship at region 'edges' 

Unnormalising coordinates is not so simple in 2D because you will have to iterate over an attempted unnormalisation checking at each point whether the current coordinate is within the region and subtracting the empty region, while expanding all other 'collapsed' sections of the region.

- The probable issue stems from whether you remove the x or z coordinate first because each change could lead to a different real position. If both are removed at the same time it could reduce the chance of this being an issue.

The current optimser limit included in the position optimiser already works for any polygon as implemented in `BLUEPRINT.geometry.inscribedrect`, although is not used at the moment (have a look at the convex flag). The only missing part is the position normalisation scheme.

This issue is just to store my current thoughts on the way forward if non convexhull regions are ever required. Feel free to comment and suggest improvements.

je-cook:
## Linked Issues

Closes #188 

## Description

This allows for optimisation of coil positions with any shaped region

The limitations are still

1) 1 coil per region (although this is not enforced, so maybe we should test and see)
2) The optimisation is not currently linked to the extents of the coil (although inscribed rectangle does function)

We should battle test this a bit...

## Checklist

### DIFF
--- a/bluemira/utilities/positioning.py
+++ b/bluemira/utilities/positioning.py
@@ -14,9 +14,9 @@ import abc
 from typing import TYPE_CHECKING
 
 import numpy as np
-from scipy.spatial import ConvexHull
 
 from bluemira.geometry.constants import D_TOLERANCE, VERY_BIG
+from bluemira.geometry.face import BluemiraFace
 from bluemira.geometry.plane import BluemiraPlane
 from bluemira.geometry.tools import slice_shape
 from bluemira.utilities.error import PositionerError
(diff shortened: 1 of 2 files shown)


je-cook:
Weird that freecad `is_closed` evaluates to `False` on a face whereas cadquery evaluates to `True`

PR #4466, opened 2026-08-05
[Bugfix] `CircularArcCurrentSource` fixed for theta != 360.

### LINKED ISSUE #4465: `CircularArcCurrentSource` not physically representative when using an angle range < 360 degrees
### Describe the bug

I've just discovered that `CircularArcCurrentSource` produces quasi-garbage when an angle range is used. This was never tested against and only the 360 case was checked with other calculations. It's actually quite difficult to pick up on, because the source does behave normally in some areas.

It is probably the reason why I was never able to get #3114 to work, and looking back at that PR, the premonition of "something far more sinister" was sadly accurate.

Ring of current vs 2 half-rings:

Image

Ring of current vs 4 quarter-rings:

Image

Root cause unknown, obviously appears related to angles.

### Steps to reproduce

_No response_

### Tracebacks

```bash

```

### Version

2.14.1.dev

### OS

### OS details

_No response_

CoronelBuendia:
## Linked Issues

Closes #4465 

## Description

Adds the missing azimuthal field component, thus making fields from circle arcs correct. It was hard to spot...

## Checklist


bluemira/magnetostatics/circular_arc.py
```diff
@@ -80,9 +84,18 @@ def bzc_integrand_full_p1(psi: float, r_pc: float, r_j: float, z_k: float) -> fl
...
+        log_arg_1 = r_j - r_pc * cos_psi + sqrt_term
+        if log_arg_1 <= 0:
+            log_arg_1 = LOG_EPS
+
+        term_1 = -z_k * np.log(log_arg_1)
+
+    log_arg_2 = -z_k + sqrt_term
+    if log_arg_2 <= 0:
+        log_arg_2 = LOG_EPS
+    return term_1 - r_pc * cos_psi * np.log(log_arg_2)
```
kj5248, line 98: Is there a reason you didn't include the final term here like you did for BRC and BZC? I know it gets added on in the primitive_bzc function with the integration of BF3 but it just seemed inconsistent stylistically.

bluemira/magnetostatics/circular_arc.py
```diff
@@ -421,8 +640,86 @@ def primitive_brc(
     else:
         # cos(-psi)^2 == cos(psi)^2
         result += 2 * bf1_r_pccos2_zk0_0_pi(r_pc, r_j)
+        result -= integrate(bf1_r_pccos2_integrand, args, phi_pc - theta, TWO_PI - theta)
```
kj5248, line 645: Have a couple of questions on how this has been set up and what it corresponds to in the paper, as I am confused why it needs so many statements and not just 3 rather than 6. Probably easier to go through on a call when you're back.

kj5248, line 645: Think this should be integrated from - theta to 2*pi - theta according to paper?

```suggestion
        result -= integrate(bf1_r_pccos2_integrand, args, - theta, TWO_PI - theta)
```

bluemira/magnetostatics/circular_arc.py
```diff
@@ -421,8 +640,86 @@ def primitive_brc(
...
+
+    else:
+        # The full 0 to 2*pi contribution of this odd integrand is zero.
         result -= integrate(
-            bf1_r_pccos2_integrand, args, phi_pc - theta, 2 * np.pi - theta
+            bf1_r_pcsincos_integrand,
+            args,
+            phi_pc - theta,
+            TWO_PI - theta,
         )
```
kj5248, line 726: Again not following the need for all the statements when looking at the paper and also how some of them cancel out

bluemira/magnetostatics/circular_arc.py
```diff
@@ -421,8 +640,86 @@ def primitive_brc(
...
+                np.pi - theta,
+            )
+
+    else:
+        # The full 0 to 2*pi contribution of this odd integrand is zero.
         result -= integrate(
-            bf1_r_pccos2_integrand, args, phi_pc - theta, 2 * np.pi - theta
+            bf1_r_pcsincos_integrand,
+            args,
+            phi_pc - theta,
```
kj5248, line 724: Again think this should just be - theta

```suggestion
            - theta,
```

bluemira/magnetostatics/circular_arc.py
```diff
@@ -477,9 +774,7 @@ def primitive_bzc(
 
         else:
             # result += 0 (the first part of BF1 is 0)
-            result -= integrate(
-                bf1_zk_integrand, args, phi_pc - theta, 2 * np.pi - theta
-            )
+            result -= integrate(bf1_zk_integrand, args, phi_pc - theta, TWO_PI - theta)
```
kj5248, line 780:
```suggestion
            result -= integrate(bf1_zk_integrand, args, - theta, TWO_PI - theta)
```

bluemira/magnetostatics/circular_arc.py
```diff
@@ -497,7 +792,7 @@ def primitive_bzc(
 
         else:
             result += 2 * bf2_rj_rpc_0_pi(r_pc, z_k)
-            result -= integrate(bf2_integrand, args, phi_pc - theta, 2 * np.pi - theta)
+            result -= integrate(bf2_integrand, args, phi_pc - theta, TWO_PI - theta)
```
kj5248, line 798:
```suggestion
            result -= integrate(bf2_integrand, args, - theta, TWO_PI - theta)
```

bluemira/magnetostatics/circular_arc.py
```diff
@@ -421,8 +640,86 @@ def primitive_brc(
...
+            # Integral from -pi to 0 equals the integral from pi to 0 after
+            # using the odd symmetry treatment in Feng's notation.
+            result += -bf1_r_pcsincos_zk0_0_pi(r_pc, r_j)
+
+        else:
+            result += -bf1_r_pcsincos_zk0_0_pi(r_pc, r_j)
+            result -= integrate(
+                bf1_r_pcsincos_integrand,
+                args,
+                np.pi,
```
kj5248, line 715: from paper think you missed a "-"

```suggestion
                - np.pi,
```

kj5248 (commented):
Had a look through the paper and compared what I got with it to what you had and found some bits that confused me and possible mistakes on your end. We can chat when you're back. Also plan to review the tools and tests and check how it all runs, which will put in another review.

tests/magnetostatics/test_circular_arc.py
```diff
@@ -78,3 +83,471 @@ def test_singularities(self):
...
+        points = self.singularity_points(
+            ds0=ds0,
+            normal0=normal0,
+            t_vec=t_vec,
+            start_angle_deg=start_angle_deg,
+            dtheta_deg=dtheta_deg,
+            exact=True,
+        )
+
+        self.assert_source_finite_at_points(arc, points)
```
kj5248, line 465: I went through these to see which singularity cases were hit by this and it mostly seemed to miss them. Found that each singularity case was only hit by 1 or 2 of the options for start_angle and theta. Not sure if this is an accuracy thing when changing into and out of cylindrical coordinates or if the singularity_points function needs a rework or it was just never designed to hit all the singularities for a given option.

bluemira/magnetostatics/circular_arc.py
CoronelBuendia: Will look into this - thanks

bluemira/magnetostatics/circular_arc.py
```diff
@@ -404,7 +625,7 @@ def primitive_brc(
         if theta == np.pi:
             result += bf1_r_pccos2_zk0_0_pi(r_pc, r_j)
 
-        if theta == 2 * np.pi:
+        if theta == TWO_PI:
```
kj5248, line 628: Should this not be elif?

```suggestion
        elif theta == TWO_PI:
```

kj5248 (commented):
Just spotted this whilst going through the singularity points

kj5248:
Just something I wanted to run by you @CoronelBuendia. For the singularities it requires a z_k of 0 however when I have plotted that it seems weird that this is where the singularities are. I've attached a figure with the singularity points on where green are the singularities, pink are what I thought would be a singularity case and red are the rest of the points.

image

bluemira/magnetostatics/circular_arc.py
CoronelBuendia: OK, it's just that the equations are different. Bz =  BF1 + BF2 + BF3, and the BF3 integrand has a separate set of oddities to it.

bluemira/magnetostatics/circular_arc.py
CoronelBuendia: This is all moot now (tragically). See my overall comment later.

bluemira/magnetostatics/circular_arc.py
CoronelBuendia: also moot, see comment later.

CoronelBuendia:
Thanks again for looking at this @kj5248. After an inordinate amount of time spent going into the seriously nitty gritty singularity stuff, I wrote some new tests where:
* For two different axes (x-z, i.e. PF coil, and x-y, i.e. TF coil like)
* For two fully circular coils (1 single circular arc, 2 circle arcs forming a circle)
* For many points along the edges including corners
* For a variety of angles around the circular source
I compare the x, y, z fields against 2-D semi-analytical formulation for a PF coil of rectangular cross-section in axisymmetry.

Tragically, I have found that either my implementation of the treatments of singularities (seem likely) or the mathematical treatments themselves (seems less likely), or both, are wrong.

Depressingly, I have also found that eliminating the singularities and just brute-forcing the integrands through the singularities (with the fancy integration fallbacks I have added) gets to within (and usually well below) 1e-5 T / MA accuracy compared with the 2-D semi-analytical formulation.

I now only treat two singularity cases in BF3 explicitly (they are trivial), and this also removes any ZeroDivisionErrors I was getting.

I'm not sure I would sign in blood saying I now cover all singular cases, but I have hammered them pretty hard in these new tests and now compare the singular cases to trusted answers from another numerical method.

I just paste the debug plots from the new tests here:
image
image

EDIT: I found more singularities... but this time not (usually) on the coil surface, but at the mirror opposite of an edge of the source term in space, e.g.:

image

I have put in a fallback 8-point GL volume-averaged calculation in the event an integration warning is raised.

Here is a plot traversing through the singularity (with the failure and fallback shown with a red cross):

image

kj5248:
Changes seem promising, is a shame about the singularity treatment not working but this seems to work now.

I changed your "test_2D_vs_3D_circular" to be absolute from relative and the differences seem to have some structure. The values are pretty inconsequential near the coil but I thought it was odd that there was so much structure to the difference:

image

There also seem to be small bright spots (Bz diff plot ~ x=6, z=3) in the diff plots that seems to be a feature of the meshgrid since if I move the grid the points move (~ x=12):

image

Again the values are so small it likely doesn't matter, just thought I'd point it out. Will continue to go through the changes.

PR #4492, opened 2026-08-24
_make_flux_surfaces to use same split function as in LegFlux class

geograham:
## Description

_make_flux_surfaces needs to use 'split' and 'sort' in the same way as the LegFlux class for equilibrium with separatrices that loop around the coils. N.B. Flux surface appears 'closed' because at this stage the code has no knowledge of the first wall. 

Steps:
- Add artificial break(s) at the largest z-coordinate(s) of the 'closed' flux line that loops around the divertor(s) coils. 
- If the equilibrium is DN then sort the split flux surfaces by IB then OB and select appropriate

There was also a typo - missing '[0]' was found and added.

## Checklist

### DIFF
--- a/bluemira/radiation_transport/flux_surfaces_maker.py
+++ b/bluemira/radiation_transport/flux_surfaces_maker.py
@@ -17,7 +17,7 @@ from bluemira.base.constants import EPS
 from bluemira.base.look_and_feel import bluemira_warn
 from bluemira.equilibria.equilibrium import Equilibrium
 from bluemira.equilibria.find import find_flux_surface_through_point
-from bluemira.equilibria.find_legs import LegFlux, NumNull, SortSplit
+from bluemira.equilibria.find_legs import LegFlux, NumNull, SortSplit, split
 from bluemira.equilibria.flux_surfaces import OpenFluxSurface, PartialOpenFluxSurface
 from bluemira.geometry.coordinates import Coordinates, coords_plane_intersect
 from bluemira.geometry.plane import BluemiraPlane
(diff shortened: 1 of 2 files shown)


bluemira/radiation_transport/flux_surfaces_maker.py
```diff
@@ -331,7 +331,14 @@ def _get_sep_out_intersection(
...
 def _make_flux_surfaces(
-    x, z, equilibrium, o_point, yz_plane, closed_perimiter, dl: float | None = None
+    x,
+    z,
+    equilibrium,
+    o_point,
+    yz_plane,
+    closed_perimiter,
+    outboard: bool | None = None,  # noqa: FBT001
+    dl: float | None = None,
```
je-cook: Outboard is not optional here and cant be None from the calling function 
```suggestion
    dl: float | None = None,
    *, 
    outboard: bool,
```

bluemira/radiation_transport/flux_surfaces_maker.py
```diff
@@ -401,7 +412,7 @@ def _make_flux_surfaces_ibob(
         )
         if (
             fs := _make_flux_surfaces(
-                x, o_point.z, equilibrium, o_point, yz_plane, lcfs_perimter, dl
+                x, o_point.z, equilibrium, o_point, yz_plane, lcfs_perimter, outboard, dl
```
je-cook:
```suggestion
                x, o_point.z, equilibrium, o_point, yz_plane, lcfs_perimter, dl, outboard=outboard
```

je-cook (changes requested):
2 related small ones, maybe we talk about this PR later too

bluemira/radiation_transport/flux_surfaces_maker.py
```diff
@@ -360,8 +367,12 @@ def _make_flux_surfaces(
         coords = coords.interpolate(dl=dl, preserve_points=True)
     if not coords.closed:
         return OpenFluxSurface(coords).split(o_point, plane=yz_plane)
-    coords.open()
     if coords.length > 1.05 * closed_perimiter:
+        coords = split(coords, NumNull.DN if equilibrium.is_double_null else NumNull.SN)
+        # Sort IB then OB
+        if isinstance(coords, list):
+            coords.sort(key=lambda coords: coords.x[0])
```
CoronelBuendia: I'd prefer a more robust sort here; I can picture situations where the first x value does not behave as one might expect.


```suggestion
            coords.sort(key=lambda coords: coords.x.min())
```

CoronelBuendia (dismissed):
OK, so I'm not entirely what this is fixing so I just have one minor

je-cook (approved):

PR #4500, opened 2026-08-25
Use absolute value of `target_value` when calculating tolerance in MagneticConstraint

clmould:
## Linked Issues

We found that if you use a negative value for the `target_value` without specifying the tolerance, you get a failure when using the PicardIterator in an optimisation. Here is the failure when I use a `PsiBoundaryConstraint` with a negative target value and no tolerance specified :
image

image

With no tolerance specified, the tolerance is calculated as follows in MagneticConstraint:

image

This default tolerance calculation should use the absolute value of the target instead, since tolerance cannot be negative and currently fails in this case.

Closes #{ID}

## Checklist


je-cook (approved):
Good catch, thanks

